PR #4, opened 2016-02-22
Fix link to overview page

a-tarasyuk:
https://facebook.github.io/draft-js/docs/quickstart-getting-started.html is broken link


zpao:
Thanks, however I accidentally fixed this locally and pushed before checking GitHub. We'll check here first next time :)

PR #8, opened 2016-02-22
Update broken example links in Inline Styles doc

mihalik:
This looks like the file that generates this page:  http://facebook.github.io/draft-js/docs/advanced-topics-inline-styles.html

Updating two links to point to the correct path.


zpao:
Thanks!

PR #9, opened 2016-02-22
Fix tex example (add missing react-dom dependency)

jerolimov:


zpao:
Thanks!

jerolimov:
Thanks for draft.js. Will think now whats missing for a "simple google docs" (with many many quotes)  :smirk:

PR #11, opened 2016-02-22
Fix logState calls in entity and link examples

razh:
`convertToRaw` expects a `ContentState` object.

Note that the `logState` functions in `plaintext.html` and `tweet.html` just call `.toJS()` without a `convertToRaw`.


hellendag:
Thanks!

PR #16, opened 2016-02-22
[docs] Change examples to use ES6 Classes

### LINKED ISSUE #7: Why are the code examples using React.createClass()?
I'm just wondering, why is `React.createClass()` being used in the code examples instead of using classes? It seems pretty inconsistent to me when ES6 is already being used and it's mixed up with ES5 stuff.

vitorbal:
Change examples to use ES6 Classes. Closes #7


hellendag:
This was mostly to avoid having to write out constructors. I'm cool with this, though. Thanks!

timche:
Thanks! :+1:

PR #29, opened 2016-02-23
Fix missing src/draft-js/lib/ dir

ZuTa:
- git cloned the repo
- npm install && npm build
- went to /website
- npm start
- went to http://localhost:8080/draft-js/index.html :
before

This PR creates src/draft-js/lib/ dir as part of postintall script.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "start": "node server/server.js",
-    "postinstall": "npm install .. && mkdir -p src/draft-js/lib"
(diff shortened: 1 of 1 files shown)


spicyj:
Can we instead add an empty `website/src/draft-js/lib/.gitkeep` file? (git add -f because it's ignored.) That's the ~standard way to handle this.

ZuTa:
@spicyj yeah, makes sense. TBH I haven't used `.gitkeep` files.

Let me update it

PR #37, opened 2016-02-23
Check devEngines on pretest

rgbkrk:
To support users on 5.x and `npm 3`. All the examples worked for me, including the TeX one.

Follow up to #31.

Thank you so much for this library, I've been wondering about handling `contentEditable` cleanly with React for editor components for some time. :heart:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "repository": "facebook/draft-js",
   "license": "BSD-3-Clause",
   "scripts": {
-    "prebuild": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
+    "pretest": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
     "build": "gulp",
     "lint": "eslint .",
     "test": "NODE_ENV=test jest"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- 5
+- 4
 sudo: false
 cache:
   directories:
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -2,6 +2,7 @@
 .*/__tests__.*
 .*/react/node_modules/.*
 .*/fbjs/node_modules/.*
+.*/node_modules/invariant/.*
 
 [include]
 ../node_modules/fbjs/lib/


millette:
This patch let's you build with npm 3.x too and I can confirm that it works for me (using node 4.3.1 and npm 3.7.3) on Debian.

rgbkrk:
Thanks @millette, updated the PR title to reflect that.

zpao:
The problem is that with the current set of transient dependencies, flow doesn't work. `invariant` will resolve to something else due to flat dependencies.

We can take this and allow npm 3 but we do need to make `flow check src` pass. You should be able to fix this specific instance by ignoring node_modules/invariant with something like this:

```diff
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -2,6 +2,7 @@
 .*/__tests__.*
 .*/react/node_modules/.*
 .*/fbjs/node_modules/.*
+.*/node_modules/invariant/.*

 [include]
 ../node_modules/fbjs/lib/
```

That doesn't guarantee that it won't break suddenly in the future but it'll work for now. Can you ensure it works and update this PR?

spicyj:
npm test also fails under npm 3 due to the mocking rules, which is why I used npm 2 on Travis.

zpao:
Oh right. And that's the exact reason we didn't do this for Relay. We can keep it as is then and move it to `"pretest"`?

rgbkrk:
> That doesn't guarantee that it won't break suddenly in the future but it'll work for now. Can you ensure it works and update this PR?

Gah, missed this bit of feedback. Addressing now.

rgbkrk:
Rebased to master to make sure that I'm testing the current merge.

```
~/code/src/github.com/facebook/draft-js$ flow check src

Found 0 errors
```

zpao:
@rgbkrk Sorry I had 2 comments in a row... Can you actually just leave this as is for the devEngines values and just change the script from prebuild to pretest? npm2 nested modules are currently required for tests to pass.

rgbkrk:
@zpao will do!

rgbkrk:
All set there.

package.json
```diff
@@ -59,8 +59,8 @@
     "webpack-stream": "^3.0.0"
   },
   "devEngines": {
-    "node": "4.x",
-    "npm": "2.x"
+    "node": ">=4.0.0",
+    "npm": ">=2.0.0"
```
zpao: Please revert this, and then can you squash to a single commit.

rgbkrk: Cool, rebased. Did you want to keep the changes to the flowconfig too?

rgbkrk:
Updated the title too while I was at it.

rgbkrk:
Clearly I had a failure at reading your review comments. Thanks for the patience.

rgbkrk:
Since the `.travis.yml` uses node 5, this now fails the tests since it's a `pretest`.

rgbkrk:
I'm going to go ahead and change the travis yml to be node 4 since that's the expectation there as well. Feel free to tell me another direction.

rgbkrk:
Also rebased down to one commit since I assume that's what you'll want here.

zpao:
Looks good, thanks!

@facebook-github-bot shipit

PR #38, opened 2016-02-23
removed two instances of unused content constants.

pooyaj:
The `content` constant is defined but never used inside `_removeLink` and `_addLink`


hellendag:
Thanks!

PR #44, opened 2016-02-23
fix link ContentBlock

vanegomez:


hellendag:
Why link directly to the markdown file? The goal here was to jump to the relevant section of the `ContentBlock` API Reference doc. The jump might have been unclear, though -- perhaps it should just link to the top of the `ContentBlock` doc?

hellendag:
Sorry, I don't think we want to link to the markdown file from the docs.

PR #56, opened 2016-02-24
[WIP] Only delete current block if blockBefore is media

### LINKED ISSUE #35: Unable to delete blockquote if its the top element
screen shot 2016-02-22 at 9 54 16 pm

[Video reproducing the issue](http://recordit.co/QS9U08PkmI)
# Steps to reproduce

I am using node `v4.3.1`
1. Build the project and open `examples/rich.html`
2. Click `Blockquote`
3. Hit Enter to create another blockquote
4. Use the arrow keys (or mouse) to go up one line to the top blockquote
5. Hit Delete

---
# Results
#### Expected:

The blockquote would delete allowing you to add new text type(s) to the line
##### Actual:

The blockquote does not delete if there are other blockquotes below it, making it difficult to
go back and add lines above it.

---

The issue only occurs if the topmost line is a blockquote.

Aweary:
- [x] Move the `getBlockForKey` check within the `if` block checking if the `blockBefore` is of the type `'media'`.
- [ ] Check that the block is unstyled so that when backspacing a blockquote/heading following a media block the style is stripped when backspacing once (and the block is completely deleted the second time).

Fixes #35

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -111,15 +111,13 @@ var RichTextEditorUtil = {
     var content = editorState.getCurrentContent();
     var startKey = selection.getStartKey();
     var blockAfter = content.getBlockAfter(startKey);
-
-    // If the current block is empty, just delete it.
-    if (blockAfter && content.getBlockForKey(startKey).getLength() === 0) {
-      return null;
-    }
-
     var blockBefore = content.getBlockBefore(startKey);
 
     if (blockBefore && blockBefore.getType() === 'media') {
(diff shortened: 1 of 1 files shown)


spicyj:
Actually, I just realized that moving the tryToRemoveBlockStyle logic to before the media handling would might seamlessly handle this logic and we can drop this extra check completely.

Aweary:
@spicyj I can implement that instead. I just need to get more familiar with the codebase.

spicyj:
Of course. :)

Aweary:
Closing this since I never got around to actually working on it 😞

Issue #54, opened 2016-02-23
Removing block style for code-block


tasti:
What's the reasoning behind preventing a `code-block` entity from becoming unstyled by backspacing when it's empty. Right now it behaves differently from the other entities (all the other block types can be unstyled).

Related code: https://github.com/facebook/draft-js/blob/master/src/model/modifier/RichTextEditorUtil.js#L418

hellendag:
IIRC, it caused confusion in cases like this:

```
function helloWorld() {
^
}
```

Where the caret is at the start of the second line. In this case, when users backspaced, they didn't want to strip the block type -- they wanted to delete the line and move the cursor back to the previous line.

One way to resolve this would be to check whether the line above is `code-block`, and if not, make the current block `unstyled`.

PR #59, opened 2016-02-24
Only make code-block unstyled if blockBefore isn't a code-block

tasti:
Fixes https://github.com/facebook/draft-js/issues/54 using suggestion.


spicyj:
@facebook-github-bot import

PR #64, opened 2016-02-24
Fix some CSS classnames

jerrysu:
I was just looking through the codez.

### DIFF
--- a/src/component/base/DraftEditor.css
+++ b/src/component/base/DraftEditor.css
@@ -42,7 +42,7 @@
   text-align: left;
 }
 
-.DraftEditor-alignLeft .public-DraftEditorPlaceholder/root {
+.DraftEditor-alignLeft .public-DraftEditorPlaceholder-root {
   left: 0;
   text-align: left;
 }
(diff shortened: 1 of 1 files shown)


tuilamap17:
1

hellendag:
@jerrysu in the house!

timmfin:
I know this is already merged, but if you really wanted to keep slashes in CSS you could escape them:

```scss
// This a an error in LibSass, "invalid top-level expression on line 2 at column 1" (tested via www.sassmeister.com)
.DraftEditor-alignLeft .public-DraftEditorPlaceholder/root {
  left: 0;
  text-align: left;
}

// But escaping the slash works (via http://stackoverflow.com/questions/18485059/forward-slash-in-class-name)
.DraftEditor-alignLeft .public-DraftEditorPlaceholder\/root {
  left: 0;
  text-align: left;
}
```

zpao:
The slashes have meaning internally and get transformed away so we won't do the escaping, but that's good to know :)

PR #66, opened 2016-02-24
Transform cx modules in gulp

spicyj:
This lets us reuse the original files directly.

### DIFF
--- a/src/component/base/DraftEditorPlaceholder.css
+++ b/src/component/base/DraftEditorPlaceholder.css
@@ -2,16 +2,16 @@
  * @providesModule DraftEditorPlaceholder
  */
 
-.public-DraftEditorPlaceholder-root {
-  color: #9197a3;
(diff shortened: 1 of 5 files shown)


src/component/base/DraftEditorPlaceholder.css
```diff
@@ -2,16 +2,16 @@
  * @providesModule DraftEditorPlaceholder
  */
 
-.public-DraftEditorPlaceholder-root {
-  color: #9197a3;
+.public/DraftEditorPlaceholder/root {
+  color: var(fbui-desktop-text-placeholder);
```
spicyj: wait damn it.

spicyj:
(cc @zpao – need to figure out what to do about those vars though…)

zpao:
Maybe we just use postcss? I wrote a quick barely tested plugin: http://astexplorer.net/#/np0DfVT78g/71

Issue #46, opened 2016-02-23
More examples


bkniffler:
I'd love to see a few more examples of usage:
- Custom blocks containing text (maybe tables? related to https://github.com/facebook/draft-js/issues/41)
- Media (Images, videos or simply colored divs, I struggle with having those behave like a part of the document, like deleting them or adding a new line after them; drag&drop (https://github.com/facebook/draft-js/issues/40) and cut/copy/paste would be even better)
- Selection tracking (something like inline toolbars, related to https://github.com/facebook/draft-js/issues/45)

mmahalwy:
+1

oyeanuj:
+1. 

For custom blocks, having rich link preview could be another option. For selection tracking, fixed static or even air toolbar examples would be swell.

bkniffler:
Wouldn't the markdown stuff be pretty simple using something along the tweet example? I've created a little gist that would use marked.js: https://gist.github.com/bkniffler/ba6a95d82f2c5c06f77b

hellendag:
> Dealing with Markdown in the text would be another item on the wishlist.

Do you mean generating markdown output? I mentioned in http://facebook.github.io/draft-js/docs/api-reference-data-conversion.html that we aren't providing a ContentState-to-markdown conversion tool with the framework since we don't have a production use for markdown at Facebook, but I expect that someone will create such a utility very soon.

Certainly using decorators could be one way to style the contents immediately when markdown is applied, if that's your objective.

WG-Com:
+1 headache for the document 

  no effect at _onBoldClick function , nothing changed,  and this is es5 

```
const MyEditor = React.createClass({
  ...

  _onBoldClick() {
   
    this.onChange(RichUtils.toggleInlineStyle(editorState, 'BOLD'));
  }

  render() {
    const {editorState} = this.state;
    return (
      <div>
        <button onClick={this._onBoldClick}>Bold</button>
        <Editor
          editorState={editorState}
          handleKeyCommand={this.handleKeyCommand}
          onChange={this.onChange}
        />
      </div>
    );
  }
});
```

hellendag:
@WG-Com: I believe that was resolved by #23.

mitermayer:
+1 this is great

AlastairTaft:
I'd love to see an example on how the toolbar works as seen in facebook notes. e.g. when you have a block selected, to the far left there's a couple of icons to alter your formatting or insert an image. Is this done by rendering the content block which has focus differently?

[...]

PR #67, opened 2016-02-24
Add a media example

tasti:
Looks like people were interested in seeing an example using media (https://github.com/facebook/draft-js/issues/46).

I created an example that shows how to use Custom Block Components to attach audio, image, and video into the editor.

Side note:

The docs (http://facebook.github.io/draft-js/docs/advanced-topics-block-components.html) led me to believe it's much simpler to render custom blocks. However, more steps (see `insertMedia(...)`) were needed to wrap around the media for it to work well. The `tex` example (https://github.com/facebook/draft-js/blob/master/examples/tex/js/modifiers/insertTeXBlock.js) has the same code to handle this. Can we provide this as a utility function?

### DIFF
--- a/examples/media/media.html
+++ b/examples/media/media.html
@@ -29,15 +29,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
       'use strict';
 
       const {
-        BlockMapBuilder,
-        CharacterMetadata,
-        ContentBlock,
         Editor,
         EditorState,
         Entity,
-        Modifier,
+        MediaUtils,
+        RichUtils,
         convertToRaw,
-        genKey,
       } = Draft;
 
       const {List, Repeat} = Immutable;
@@ -57,21 +54,44 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
             console.log(convertToRaw(content));
           };
 
+          this.handleKeyCommand = this._handleKeyCommand.bind(this);
+          this.addMedia = this._addMedia.bind(this);
           this.addAudio = this._addAudio.bind(this);
           this.addImage = this._addImage.bind(this);
           this.addVideo = this._addVideo.bind(this);
         }
 
-        _addAudio(e) {
-          this.onChange(insertMedia(this.state.editorState, 'audio'));
+        _handleKeyCommand(command) {
+          const {editorState} = this.state;
+          const newState = RichUtils.handleKeyCommand(editorState, command);
+          if (newState) {
+            this.onChange(newState);
+            return true;
+          }
+          return false;
         }
 
-        _addImage(e) {
-          this.onChange(insertMedia(this.state.editorState, 'image'));
+        _addMedia(type) {
+          const src = window.prompt('Enter a URL');
+          if (!src) {
+            return;
+          }
+
+          const entityKey = Entity.create(type, 'IMMUTABLE', {src});
+
+          return MediaUtils.insertMedia(this.state.editorState, entityKey, ' ');
         }
 
-        _addVideo(e) {
-          this.onChange(insertMedia(this.state.editorState, 'video'));
+        _addAudio() {
+          this.onChange(this._addMedia('audio'));
+        }
+
+        _addImage() {
+          this.onChange(this._addMedia('image'));
+        }
+
+        _addVideo() {
+          this.onChange(this._addMedia('video'));
         }
 
         render() {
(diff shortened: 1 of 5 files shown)


hellendag:
Want to take a look at what we're doing in Notes and see if it's effectively the same? If it looks generic enough, it seems like a good move to put it into `RichUtils` or maybe a separate `MediaUtils`.

tasti:
Yup, it's identical to what we do internally. I'll add it into a `MediaUtils` here.

examples/media/media.html
```diff
@@ -0,0 +1,211 @@
...
+
+      const Image = (props) => {
+        return <img src={props.src} style={styles.media} />;
+      };
+
+      const Video = (props) => {
+        return <video controls src={props.src} style={styles.media} />;
+      };
+
+      const Media = (props) => {
```
hellendag: Thinking it might be good to wrap the output of this function in a `div`. I'm not sure how I feel about an example that potentially returns `display: inline` elements for block renderers.

hellendag:
@facebook-github-bot import

tasti:
@facebook-github-bot import

(Doesn't look like it worked the first time)

spicyj:
@tasti At https://fburl.com/210824701 (fb internal link) you can see that it failed because the import script doesn't support binary files. Might need to import this one manually…

hellendag:
Yeah, just arc diff it and we'll go from there. :)

tasti:
@facebook-github-bot import

I popped off the commit with the binaries (makes it easier than manually importing). I'll add them back internally.

nikgraf:
This implementation is suuuuper helpful as a base example, but leaves a couple of UX issues open.

The div surrounding the media should be content editable fa. Otherwise if the content is not the full with (like with the image). You then can click in front of the image and start typing and will actually type in the next line. In case you have contenteditable deactivated you need to set the block id otherwise it will get to a null error in case you do a selection (see [here](https://github.com/nikgraf/react-unicorn-editor/blob/ffcaf9ccf9bae796f7151a510b0c5d580ca9e412/src/components/Editor/Sticker/index.js#L22)). This leads to issue when deleting which you can solve with a cleanup (see an example [here](https://github.com/nikgraf/react-unicorn-editor/blob/ffcaf9ccf9bae796f7151a510b0c5d580ca9e412/src/components/Editor/modifiers/cleanupEmptyStickers.js)).

The point I'm trying to make: this add media example has a few UX issues. Not sure if this should be fixed in the draftjs core or in the example …

tasti:
Thanks for catching that, @nikgraf!

I don't think we have a good solution for this yet. Even in Facebook Notes, we set `contentEditable={false}` and `data-offset-key={...}`. I'll look into how we can improve this :)

nikgraf:
@tasti sounds good - I'm happy to pair program on this or discuss API changes … :smile:

J22Melody:
When will this be merged?I'm looking forward to use `MediaUtils`.

tasti:
@J22Melody - This should be ready soon, just waiting on internal changes for https://github.com/facebook/draft-js/pull/200

voronianski:
@tasti also waiting for this! thanks a lot guys!

ianstormtaylor:
@tasti any update on this? Would love to be able to use `MediaUtils` for images!

tasti:
@ianstormtaylor Sorry it's taking long, I'm asking to speed up the review right now :)

src/model/modifier/MediaUtils.js
```diff
@@ -0,0 +1,84 @@
...
+    const selectionState = editorState.getSelection();
+
+    const afterRemoval = DraftModifier.removeRange(
+      contentState,
+      selectionState,
+      'backward'
+    );
+
+    const targetSelection = afterRemoval.getSelectionAfter();
+    const afterSplit = DraftModifier.splitBlock(afterRemoval, targetSelection);
```
ianstormtaylor: Is the need to split the block here in case the current block has text in it still? In the case that it's completely empty, might be better to not split the block, so that it doesn't end up as an extra empty block above the media block. Just a thought!

ianstormtaylor:
@tasti No worries!

src/model/modifier/MediaUtils.js
tasti: Yes and yes! I didn't get a chance to see this before the push, but feel free to make a PR for this.

max-mykhailenko:
When you plan release this to npm?

tasti:
@max-mykhailenko We're aiming for a release on Tuesday.

max-mykhailenko:
Do you release this feature?

guitardave24:
This is in

hellendag:
This was delayed a bit to make sure that the rename to `atomic` would be included in the release. 0.4 did not include this, but 0.5 will. That should be released next Tuesday. Apologies for the delay!

PR #71, opened 2016-02-24
Suppress contentEditable warning

mxstbr:
This will come into effect with the 15.0 release of React, see https://github.com/facebook/react/pull/6112

> Note: I wasn't entirely sure if you want to specify the `={true}`, since it can be omitted, and I couldn't find an example with/without it. Let me know if you want me to remove it

/cc @spicyj


spicyj:
With or without the true is fine, thank you. This seems good to me.

spicyj:
I actually ended up getting this as part of #179. Thanks for sending this in though!

PR #72, opened 2016-02-24
Replace all uses of var with const or let

garbles:
Variable declaration is inconsistent across the project. Most places using `var` but some using `const`/`let`.

PR removes all uses of `var` in favor of `const` or `let` (depending on whether a variable changes reference).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,5 +11,6 @@ module.exports = {
   rules: {
     'react/jsx-uses-react': 1,
     'react/react-in-jsx-scope': 1,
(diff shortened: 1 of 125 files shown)


hellendag:
Thanks for this! We're going to run a codeshift internally to do this, then sync to GitHub.

garbles:
Aye. No worries. Thanks for the fyi :)

hellendag:
Going to go ahead and close this one since we're going to do it internally and sync out. We've had a bit of a delay on the codemod, but I promise it will happen. :)

PR #76, opened 2016-02-24
Correct and clarify docs for handleBeforeInput

ohanar:
Currently the docs state that a `beforeInput` event is passed as an argument to the `handleBeforeInput` prop. This is incorrect, only the characters to be inserted are passed.

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -152,9 +152,11 @@ for details on usage.
 
 #### handleBeforeInput
 ```
-handleBeforeInput?: (e: SyntheticInputEvent) => boolean
+handleBeforeInput?: (chars: string) => boolean
 ```
-Handle a `beforeInput` event before character insertion occurs within the editor.
+Handle the characters to be inserted from a `beforeInput` event. Returning `true`
+causes the default behavior of the `beforeInput` event to be prevented (i.e. it is
+the same as calling the `preventDefault` method on the event).
 Example usage: After a user has typed `- ` at the start of a new block, you might
 convert that `ContentBlock` into an `unordered-list-item`.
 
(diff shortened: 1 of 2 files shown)


hellendag:
Which reference manual? This should continue to pass only the chars, so it sounds like an incorrect doc or typedef needs fixing rather than this code.

ohanar:
It is mentioned both in the reference manual [here](https://github.com/facebook/draft-js/blob/master/docs/APIReference-Editor.md#L153) and in the prop types [here](https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorProps.js#L97).

hellendag:
Got it, thanks. The correct fix here is to change both places to document that the argument is a `string`, not the `SyntheticInputEvent`, if you wouldn't mind updating the PR.

I would have hoped that Flow would have caught that, but I guess not. Sorry for the confusion!

ohanar:
Sorry, why is that the correct fix? I could imagine in some use cases having access to the full event might be helpful. I don't mind updating the pull request, but I would like to understand why you think that is the correct fix.

hellendag:
The main reason is the complexity of the event handling within the `editOnBeforeInput` function. As often as possible, we try to let typed characters fall through to be rendered natively (i.e. to preserve spellcheck highlighting, limit re-renders, etc.)

Exposing the event could, for instance, lead to a `preventDefault` call that affects native behavior that should follow the `handleBeforeInput` call.

ohanar:
Thank you for the explanation. That makes a lot of sense. I updated the pull request to update the docs and prop type.

hellendag:
The change is good, thanks! I haven't merged this PR yet because we're still working out our internal-to-GH syncing. But I will soon. :)

hellendag:
@facebook-github-bot import

PR #78, opened 2016-02-25
Fix linter warnings

garbles:
Fixes linter warnings.

Also adds linting to the CI build, but warnings still exit with `0` which doesn't cause the suite to fail. I'm not even sure whether you guys want the suite to fail on this condition; though the contributing docs asks users to make sure that the linter passes before submitting a PR.


hellendag:
@facebook-github-bot import

PR #94, opened 2016-02-25
Change button + nav color to match theme

tasti:
Some colors were left over from the flux template. So, I got rid of the greens and threw in some reds.


hellendag:
@facebook-github-bot shipit

Issue #95, opened 2016-02-26
Add support for h3, h4, and h5


mitermayer:
Hi there,

Was wondering if you guys have support for other heading types, If not are you open for PRS ? I am attempting to migrate a project to use draft js as a replacement for tinymce

thanks,
miter

hellendag:
Sure, send a PR. :) You'll want to add these values to `DraftBlockType` as well as `getElementForBlockType.js`.

mitermayer:
Just created a PR for it https://github.com/facebook/draft-js/pull/96

PR #96, opened 2016-02-26
extra-headings-support - Adding support for other blockmode headings

mitermayer:
- adding support for other blockmode headings

### DIFF
--- a/src/component/utils/getElementForBlockType.js
+++ b/src/component/utils/getElementForBlockType.js
@@ -17,9 +17,15 @@ import type {DraftBlockType} from 'DraftBlockType';
 function getElementForBlockType(blockType: DraftBlockType): string {
   switch (blockType) {
     case 'header-one':
-      return 'h2';
+      return 'h1';
     case 'header-two':
+      return 'h2';
+    case 'header-three':
       return 'h3';
+    case 'header-four':
+      return 'h4';
+    case 'header-five':
+      return 'h5';
     case 'unordered-list-item':
     case 'ordered-list-item':
       return 'li';
--- a/src/model/constants/DraftBlockType.js
+++ b/src/model/constants/DraftBlockType.js
@@ -20,6 +20,9 @@ export type DraftBlockType = (
   'paragraph' |
   'header-one' |
   'header-two' |
+  'header-three' |
+  'header-four' |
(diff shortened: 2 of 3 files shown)


mitermayer:
this solves https://github.com/facebook/draft-js/issues/95

website/src/draft-js/index.js
```diff
@@ -145,6 +145,9 @@ class StyleButton extends React.Component {
 const BLOCK_TYPES = [
   {label: 'H1', style: 'header-one'},
   {label: 'H2', style: 'header-two'},
+  {label: 'H3', style: 'header-three'},
```
mitermayer: Also added to the website demo

src/component/utils/getElementForBlockType.js
```diff
@@ -17,9 +17,15 @@ import type {DraftBlockType} from 'DraftBlockType';
 function getElementForBlockType(blockType: DraftBlockType): string {
   switch (blockType) {
     case 'header-one':
-      return 'h2';
```
mitermayer: I believe this was a typo ? so I updated the header-one to return h1. Is that a problem or was that the intendend behaviour ? If so I can amend the PR ?

hellendag: This was intended, but is a decision from long ago. I'll check to see if we can go ahead and change it. Your change is surely more sane for a public repo. :)

hellendag:
@facebook-github-bot import

hellendag:
I'm going to go ahead and add `header-six` to the imported diff to round it out. :)

src/model/constants/DraftBlockType.js
```diff
@@ -20,6 +20,9 @@ export type DraftBlockType = (
   'paragraph' |
   'header-one' |
   'header-two' |
+  'header-three' |
+  'header-four' |
+  'header-five' |
```
sstur: Why stop at five? It just seems like an arbitrary choice between 1 and 6. Maybe a better choice than 2, but interested to know why you chose 5.

hellendag: imgres

sstur: Hahaha, by that logic, we should include h7!

Issue #53, opened 2016-02-23
Warning: A component is `contentEditable` and contains `children` managed by React


iroy2000:
The following is an example editor component

`
import React from 'react';
import {Editor, EditorState} from 'draft-js';

export default class MyEditor extends React.Component {

    constructor(props) {
        super(props);

        this.state = {
            editorState: EditorState.createEmpty()
        };
    }

    onChange = (editorState) => {
        this.setState({editorState});
    }

    render() {
        const {editorState} = this.state;
        return <Editor editorState={editorState} onChange={this.onChange} />;
    }
}
`

And in my parent component

`import MyEditor from './MyEditor'`

....  // in the render() method
`
    render() {
        let error = this.getError();

        return (
            <form className="post__form">
                {error}
                <div className="form-group">
                    <label htmlFor="title">Title</label>
                    <input type="text" ref="title" className="form-control" id="title" placeholder="Title" />
                </div>
                <div className="form-group">
                    <label htmlFor="description">Description</label>

                    <MyEditor />

                </div>
                <button type="submit" onClick={this.onClickHandler.bind(this)} className="btn btn-default">Submit</button>
            </form>
        );
    }

`

**And I see this warning**

Warning: A component is `contentEditable` and contains `children` managed by React. It is now your responsibility to guarantee that none of those nodes are unexpectedly modified or duplicated. This is probably not intentional.

tasti:
You aren't doing anything wrong.

This a known issue (http://facebook.github.io/draft-js/docs/advanced-topics-issues-and-pitfalls.html#react-contenteditable-warning) and it's being tracked in the react repo: https://github.com/facebook/react/issues/5837

hellendag:
Yep, really looking forward to eliminating this warning. :)

davisml:
I ended up adding this code to my project as a temporary solution:

`console.error = (function() {
    var error = console.error

    return function(exception) {
        if ((exception + '').indexOf('Warning: A component is `contentEditable`') != 0) {
            error.apply(console, arguments)
        }
    }
})()`

jarsbe:
@davisml crazy, I was just looking for a way to do this.

davisml:
@jarsbe I'm glad I could help! :)

jarsbe:
@davisml That's some speedy tech support, cheers!

sdoomz:
As an alternative you can set your console filter to regexp mode with the value: `^((?!contentEditable).)*$`

[...]

PR #98, opened 2016-02-26
fix contentEditable warning

linkkingjay:
See https://github.com/facebook/draft-js/issues/53. The issue has been fixed in the React repo: https://github.com/facebook/react/pull/6112 .

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -262,6 +262,7 @@ class DraftEditor
             ref="editor"
             role={readOnly ? null : (this.props.role || 'textbox')}
             spellCheck={allowSpellCheck && this.props.spellCheck}
+            suppressContentEditableWarning={true}
             tabIndex={this.props.tabIndex}
             title={hasContent ? null : this.props.placeholder}>
             <DraftEditorContents


src/component/base/DraftEditor.react.js
```diff
@@ -262,6 +262,7 @@ class DraftEditor
             ref="editor"
             role={readOnly ? null : (this.props.role || 'textbox')}
             spellCheck={allowSpellCheck && this.props.spellCheck}
+            suppressContentEditableWarning={!readOnly}
```
hellendag: Couldn't this always just be `true`?

linkkingjay: @hellendag Of course it can. Fixed.

adcoolguy:
We have the same warning even after using this new attribute. Are we missing something?

                                <div className="control-group">
                                    <label htmlFor="desc" className="col-sm-2 control-label">Description:</label>
                                    <div className="col-sm-10">
                                        <Editor ref="desc" suppressContentEditableWarning editorState={editorState} onChange={this.onChange} />
                                    </div>
                                </div>

We use React 14.0. Thanks!

tasti:
@adcoolguy: The `suppressContentEditableWarning` prop will be added into React 15.0 (as mentioned https://github.com/facebook/react/issues/5837).

Also, you won't have to place the prop on your `Editor` component; this PR is placing it internally for you.

spicyj:
I actually ended up getting this as part of #179. Thanks for sending this in though!

PR #111, opened 2016-02-27
Exclude trailing block for type toggling if at zero offset

### LINKED ISSUE #86: Block-type changes can affect adjacent sibling block
**Repro:**
- Navigate to the [Draft.js homepage](https://facebook.github.io/draft-js/)
- Add a few blocks of text
- Triple-click a block of text
  - This will select the entire block
- Change the block type to **H1**

**Expected:**
- The block type for the selected line changes to a heading

**Actual:**
- The block type for the selected line *and the following line* changes to a heading

triple-click

As I typed this bug with Github Markdown editor, I noticed it also demonstrates this behavior (triple-clicking a line and applying **bold** will drop the trailing `**` at the beginning of the next line). Maybe this is out of the hands of Draft. However, I do believe that fixing the issue would lead to more reasonable UX.

hellendag:
Resolves #86.

In cases where the user triple-clicks on a block, then toggles the block type, it feels weird to have the trailing block also change block type.

The `SelectionState` resulting from the triple-click is correct, but to improve the interaction here, let's discard the trailing block for this modifier.

Test Plan:

View rich.html, add three lines of text.

- Triple-click first line, toggle H1. Verify that only the first line toggles.
- Repeat with second line, verify that only the second line toggles.
- Repeat with third line (no trailing block), verify that only the third line toggles.

Also select across multiple blocks by dragging, verify that toggling block types properly affects all visibly selected blocks.

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -18,6 +18,7 @@ var DraftModifier = require('DraftModifier');
 var EditorState = require('EditorState');
 
 var adjustBlockDepthForContentState = require('adjustBlockDepthForContentState');
+var nullthrows = require('nullthrows');
 
 import type ContentState from 'ContentState';
 import type {DraftBlockType} from 'DraftBlockType';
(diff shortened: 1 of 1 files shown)


hellendag:
@facebook-github-bot import

mzedeler:
This selection behavior is still relatively strange. Why hasn't it been fixed inside draft in stead of patching all the utilities that operate on `editorState` objects?

Issue #112, opened 2016-02-28
Link example not rendering link when rich text is pasted


brijeshb42:
In the [link example](https://github.com/facebook/draft-js/blob/master/examples/link/link.html), the links are only rendered when explicitly created through `Add Link` button but not when rich text containing links are pasted.
Also, when pasting rich text, the raw data of the block map contains entity data as
```
data: {
  url: "https://facebook.github.io/draft-js/docs/api-reference-editor-state.html#getblocktree",
}
```
Here, the key is `url` but the link example creates and uses key `href`. At first I thought this was the problem, but even after fixing the key usage, the links are not rendered on pasting.

hellendag:
Sorry, the link example wasn't really built with the paste case in mind. I should have thought of that. :)

> Here, the key is url but the link example creates and uses key href.

Good catch, I should have been consistent about that. I was also inconsistent about the entity type, using `link` instead of `LINK`.

Fixing these two items will repair the example.

brijeshb42:
This [PR](https://github.com/facebook/draft-js/pull/114) tries to fix the above issue that I created by adding the consistency you suggested. It can now render pasted link in rich text successfully.

PR #114, opened 2016-02-28
Fix the link example to render links when rich text containing links are pasted

brijeshb42:
Make the example for links consistent by using the same key (`url`) as defined by `draft-js` paste instead of `href` and use `LINK` in `Entity.create` instead of `link` to maintain consistency as suggested in [issue 112](https://github.com/facebook/draft-js/issues/112).

This PR enhances link example by:
* Rendering links in pasted rich text.
* Handling `ENTER` key press for link's url input box besides clicking on.

### DIFF
--- a/examples/link/link.html
+++ b/examples/link/link.html
@@ -128,7 +128,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
                   style={styles.urlInput}
                   type="text"
                   value={this.state.urlValue}
-                  onKeyUp={this.onLinkInputKeyDown}
+                  onKeyDown={this.onLinkInputKeyDown}
                 />
                 <button onMouseDown={this.confirmLink}>
                   Confirm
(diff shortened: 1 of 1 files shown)


examples/link/link.html
```diff
@@ -178,9 +186,9 @@
       }
 
       const Link = (props) => {
-        const {href} = Entity.get(props.entityKey).getData();
+        const {url} = Entity.get(props.entityKey).getData();
```
Tracyhernando79: Financial kill one of yourself heads

examples/link/link.html
```diff
@@ -121,6 +128,7 @@
                   style={styles.urlInput}
                   type="text"
                   value={this.state.urlValue}
+                  onKeyUp={this.onLinkInputKeyDown}
```
hellendag: Why not `onKeyDown`?

brijeshb42: No specific reason. Can change if needed.

hellendag: Let's stick with `onKeyDown` for consistency, since `onKeyUp` isn't used elsewhere in the framework at this point.

brijeshb42: Updated.

hellendag:
Thanks!

@facebook-github-bot shipit

PR #116, opened 2016-02-28
Fix react key warnings

amasad:
https://www.dropbox.com/s/788d5amion1dev5/Screenshot%202016-02-28%2013.24.30.png?dl=0


spicyj:
These are autogenerated from the files in `website/` on master. I think this already got fixed in the actual "rich" example so you should be able to copy the new component code over verbatim.

amasad:
cool, thanks

PR #123, opened 2016-02-29
Use es6-shim, mention polyfill requirement in docs

### LINKED ISSUE #83: Doesn't currently work in IE11
On IE11, Facebook.github.io/draft-js doesn't show the editor. Similarly for a basic editor running locally. 

if (!query.startsWith(name)) --> Object doesn't support property or method 'startsWith'

zpao:
Should fix #83


rgbkrk:
Cool, very glad you have this as a devDep and are not including it in the distributed package.

hellendag:
@facebook-github-bot shipit

PR #130, opened 2016-03-01
Added polyfills for IE9+

kim3er:
Draft.js website doesn't work in IE9+. This is because Draft.js relies on 3 methods that are not currently supported in IE; `String.startsWith`, `String.endsWith` and `Object.assign`. This pull request adds them.

I assumed the Babel Polyfill would have added all three, but that appears not to be the case.

Please see https://github.com/facebook/draft-js/issues/83 for more details.

### DIFF
--- a/website/src/draft-js/index.js
+++ b/website/src/draft-js/index.js
@@ -251,7 +251,52 @@ var index = React.createClass({
             </div>
           </section>
         </section>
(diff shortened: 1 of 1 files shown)


hellendag:
I believe #123 takes care of this.

kim3er:
Yep, my bad.

hellendag:
No worries, thanks for being proactive. :)

PR #131, opened 2016-03-01
Use Slackin link for Slack invites

hellendag:
Still not sure if we'll be moving over to Discord, but in the meantime this will streamline things.


hellendag:
@facebook-github-bot shipit

Issue #125, opened 2016-03-01
Copy/pasted custom block components may not be preserved


hellendag:
When copying a custom block within a `Draft` editor, pasting it back into the editor may not preserve it.

This can be reproduced within FB Notes by copying a section of text that includes an embedded photo, then pasting it back into the editor.

Source of issue: we check the DOM clipboard plaintext with the internal clipboard text to determine whether the pasted contents are the same as the internal clipboard. Since the custom block may render something other than the plain text of the `ContentBlock`, the check may fail, in which case the HTML contents are parsed and pasted instead. (https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnPaste.js#L113)

My idea to resolve this is to have an Editor-level random key to be rendered as an attribute within the DOM, perhaps at the `DraftEditorBlock` level. During a paste, we can check the pasted HTML for the random key. If it is present, we can assume that the paste is internal and should use the internal clipboard. If it is absent, we use the raw HTML instead.

This should preserve custom blocks during the paste, since the internal clipboard representation is just an `OrderedMap<string, ContentBlock>` being inserted into the `ContentState`. Following the paste, the custom block will be rendered correctly in the Editor.

spicyj:
Does it make more sense to generate a per-copy key rather than a per-editor one? Not sure how much control we have there.

hellendag:
It would have to be part of the copied HTML in order to have something to read from at paste time. So I think a per-copy key would mean having something rendered into the editor DOM prior to the copy command, then changed afterward. Not sure if that's necessary, if the per-editor key is effective enough.

remko:
Is this just for *custom* block components? Copy/pasting H1 doesn't preserve the H1 either for me.

hellendag:
@remko In the case you describe, it may be that the type of the target block for the paste is overriding the pasted input. The bug here is different, but there may be some worthwhile discussion around default block paste behavior.

hellendag:
The issue in question here is resolved by https://github.com/facebook/draft-js/commit/c0f241482b393d31937409e0104d1d67b8cc7952.

ivan-moroz:
I have a problem using draft.js on my web application. I need to change pasted from MS Word text. Using the **handlePastedText** method I try to change the pasted HTML and create new editorState:

`var fragment = convertFromHTML(this._clearHTML(html));

            var newContentState = Modifier.replaceWithFragment(
                contentState,
                selectionState,
                fragment
            );`

[...]

PR #135, opened 2016-03-02
generateBlockKey -> generateRandomKey

hellendag:
I'd like to generate an editor-level key, too. Might as well reuse the same module, since it's useful.


hellendag:
Editor-level key is to be used to solve #125.

spicyj:
:+1:

hellendag:
@facebook-github-bot import

PR #139, opened 2016-03-02
Consider passing in a style prop to the editor

AlastairTaft:
In my code I'd like to be able to pass in a `style` object like this

```html
<div style={styles.editorContainer} onClick={this._focus}>
  <SideControl />
  <PopoverControl />
  <Editor
    style={{paddingTop: 1}} // Fix for collapsing margins
    onChange={this._onChange}
    ref="editor"
    ...
  />
</div>
```

Currently I achieve the same thing with a style sheet but I don't like having to have the extra file dependency. 
```css
.DraftEditor-root {
  /* Stop collapsing margins breaking our side control positioning */
  padding-top: 1px;
}
```

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -220,7 +220,7 @@ class DraftEditor
     const hasContent = this.props.editorState.getCurrentContent().hasText();
 
     return (
-      <div className={rootClass}>
+      <div className={rootClass} style={this.props.style}>
         {this._renderPlaceholder()}
         <div
           className={cx('DraftEditor/editorContainer')}


hellendag:
I've been thinking about improving support for inline styles for elements throughout the editor.

```javascript
type EditorStyle = {
  root?: Map,
  placeholder?: Map,
  ...
};

// within DraftEditorProps
editorStyle: EditorStyle,
```

If a style for a certain element is defined, it is applied when rendering that element.

I stuck with CSS classes because it felt like for most engineers, that would probably be the most familiar way to style and override. But I'm definitely open to discussing that, especially if things are moving more toward inline styles in place of classes. (cc @spicyj)

What do you think about more extensive style support like the proposal above?

AlastairTaft:
I like being able to style the different elements. It's similar to the way material-ui works. A good example is [the dropdown component](http://www.material-ui.com/#/components/dropdown-menu), it allows you to pass in `iconStyle`, `labelStyle`, `menuStyle` and `underlineStyle`, while the `style` prop styles the root element.

However it's not as expressive as css, css allows you to style children of classes if you really want to. So having both a className and a way to hook in inline styles would be best in my opinion.

hellendag:
I think that sounds similar to the conclusion @spicyj and I came to when we chatted about it, too. We'd at least have to support both.

spicyj:
> css allows you to style children of classes if you really want to

I think we would recommend using a custom leaf renderer (maybe the only way to do this is with decorators?) so that you can use any styles you want – but in the interim it may well make sense to support both classnames and inline styles so that people can use whichever they prefer.

hellendag:
At the decorator/leaf level I don't *think* we need to change anything.

For decorators, you have total control over the rendering and can set CSS classes or style objects as desired.

For leaves, everything just uses `customStyleMap` and `DefaultDraftInlineStyle` and it's the lowest level of rendering, so style objects give you all the control you need. I can't really think of a scenario where one would need class names instead.

AlastairTaft:
I haven't heard of a leaf decorator before, by the sounds of it I guess it crawls the child elements? Handling styling in react is still a grey area for me, I can't decided what the best practice is. When you call into question media queries, pseudo elements and hover, seems like there isn't a nice way of handling that without css.

gscottolson:
@AlastairTaft There are some JavaScript approaches to styling that handle media queries, pseudo elements and hover. One that comes to mind is [Radium](http://stack.formidable.com/radium/).

My personal preference is to work with stylesheets, but I also prefer preprocessors to resolve browser vendor differences and enable conveniences like variables. I am currently using [stylus](http://stylus-lang.com/) with [CSS Modules](https://github.com/css-modules/css-modules), which works to solve the “everything is part of the global namespace” issue associated with vanilla CSS. [This article](http://glenmaddern.com/articles/css-modules) has a great introduction to the tradeoffs of working with CSS Modules.

AlastairTaft:
I don't know to much about Radium but understood it implements that stuff using JavaScript events, it may be fine but to me that just sound expensive and may result in a laggy user experience compared to native css, although I've never tried the library.

Thanks for the heads up on stylus and CSS Modules.

Nudelsieb:
It would be awesome to have an option to add a custom className to the Editor component to override the internal base class "DraftEditor-root" and do some custom styling alongside an additional style prop (as proposed here). I am using several instance at the same time and need to be able to style them independently from each other. Currently I don't know how to achieve that.

hellendag:
@Nudelsieb Can you target it by putting a class name on the container component? Or do you need to be able to target the editor node itself?

I'm going to go ahead and close this PR since it was last updated in March, but I think the discussion about top-level styling can continue. Feel free to put up a new PR if you'd like.

Nudelsieb:
I am not sure if it is worth having a new PR. In essence I would like to be able to manipulate the top level node like a normal html element with CSS, assign a class to it or throw inline styles at it. For example, if you decide to display only one element absolutely and change its dimensions, outline, etc, how would you do it, without influencing the remaining elements deriving from .DraftEditor-root. You could also change the default font top level. I had a lot of ideas. But my purpose it is necessary to treat each instance of the main container/element separately with CSS.

Thanks

Issue #68, opened 2016-02-24
content.findEntityRanges() fails to find Entities after calling Modifier.applyEntity().


ben-davis:
I'm using Draft to highlight hashtags in a readonly block of text (I understand Draft might not be the right tool for the job considering I don't require selection).

From what I understand, the correct approach should be:
- map over each `contentBlock`
- for each `contentBlock` find `(start, end)` ranges for each hashtag
- for each hashtag, create an `Entity` and a `SelectionState` for the tag
- apply that `Entity` using `Modifier.applyEntity` for the given `SelectionState`.

Here's what that looks like

```javascript
const contentState = ContentState.createFromText(content);
contentState.getBlockMap().map(contentBlock => {
  this._findWithRegex(HASHTAG_REGEX, contentBlock, (start, end) => {
    let selectionState = SelectionState.createEmpty(contentBlock.getKey());
    selectionState = selectionState.merge({
      anchorOffset: start,
      focusKey: contentBlock.getKey(),
      focusOffset: end,
    });

    const tag = contentBlock.getText().slice(start, end);
    const key = Entity.create('HASHTAG', 'MUTABLE', {tag: tag});

    Modifier.applyEntity(contentState, selectionState, key);
  });
});
```
where `this._findWithRegex` is the same function from the `tweet` example.

After doing this, I expect to be able to apply a decorator strategy using the following:
```javascript
_findHashtagEntitiesStrategy(contentBlock, callback) {
  contentBlock.findEntityRanges(
    (character) => {
      const entityKey = character.getEntity();
      return (
        entityKey !== null &&
        Entity.get(entityKey).getType() === 'HASHTAG'
      );
    },
    callback
  );
}
```
However it seems that `contentBlock.findEntityRanges` fails to find any `Entity`s for the any of the `ContentBlock`s.

hellendag:
For hashtags, you don't need to create entities at all, since there is no metadata -- there is just text. You can just use a decorator strategy/component pair, without needing to use `findEntityRanges`. This is what we do for hashtags in Facebook inputs.

ben-davis:
That's what I initially was doing, but the component in the pair just receives the text as a component in its children. I could read `React.Children.only(this.props.children).props.text`, but reading into the props of an undocumented API seems reckless. I should mention I need to read the text of the tag for routing and other functionality.

hellendag:
Ahh, interesting. I think it would be fine to make a change to pass the full decorated text as a prop to the decorator component.

[...]

PR #144, opened 2016-03-02
Add decoratedText prop to DecoratorComponent

### LINKED ISSUE #140: No easy way to get decorated text within decorator component
If you want the decorator styles to be different depending on the text that's decorated, it would be nice to get the text somehow. Best I came up with is:

```js
{
  strategy: ...,
  component: (props) => {
    // This isn't public API.
    const text = props.children.map((leaf) => leaf.props.text).join('');
    ...
  },
}
```

though it looks like `DraftEditorBlock.react` could easily pass down more information about the contained leaves. Maybe a reference to the block and the range would be enough?

`props.offsetKey` currently contains the block key, but `DraftOffsetKey` is not exposed as public API so you can't decode it, plus the block key is not useful without a reference to the editor state which I don't believe you have.

hellendag:
Resolves #140 and #68.

Tested with `console.log(this.props.decoratedText)` in render method of TokenSpan in entity.html.

### DIFF
--- a/src/component/contents/DraftEditorBlock.react.js
+++ b/src/component/contents/DraftEditorBlock.react.js
@@ -182,6 +182,7 @@ class DraftEditorBlock extends React.Component {
       return (
         <DecoratorComponent
           {...decoratorProps}
+          decoratedText={decoratedText}
           dir={dir}
           key={decoratorOffsetKey}
           entityKey={block.getEntityAt(leafSet.get('start'))}


spicyj:
Seems okay. :+1: I still have that gut feeling that we should expose the styles somehow but I don't know the best way. I guess we'll wait for a use case!

hellendag:
We could do that at some point, though decorators should really just use `this.props.children` to render the text contents, including styled ranges. Otherwise they'll risk breaking the expected `SelectionState`-to-DOM mapping.

hellendag:
@facebook-github-bot import

spicyj:
Yeah, I ran into that last night too when I accidentally rendered the raw text instead of the children.

PR #145, opened 2016-03-03
Allow to pass handleDrop, similar to handleDroppedFiles

bkniffler:
This will allow to handle drops and suppress the default behavior if
necessary.

### DIFF
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -14,8 +14,10 @@
 
 import type ContentBlock from 'ContentBlock';
 import type {DraftEditorCommand} from 'DraftEditorCommand';
+import type {DraftDragType} from 'DraftDragType';
 import type {DraftTextAlignment} from 'DraftTextAlignment';
 import type EditorState from 'EditorState';
+import type SelectionState from 'SelectionState';
 
 export type DraftEditorProps = {
   /**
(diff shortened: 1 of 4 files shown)


hellendag:
We should also include the prop in `DraftEditorProps` and in the component API doc.

I think `handleDroppedFiles` is also missing from those, so if you don't mind adding that one too, that would be cool.

src/component/base/DraftEditorProps.js
```diff
@@ -98,6 +98,11 @@ export type DraftEditorProps = {
 
   handlePastedFiles?: (files: Array<Blob>) => boolean,
 
+  // Handle dropped files
+  handleDroppedFiles?: (selection: SelectionState, files: Array<Blob>) => boolean,
+  // Handle other drops to prevent default text movement/insertion behaviour
+  handleDrop?: (selection: SelectionState, dataTransfer: DataTransfer, isInternal: boolean) => boolean,
```
hellendag: It might seem an odd request, but I really dislike boolean arguments. Can you create a module that defines a Flow type to denote this value?

```javascript
@providesModule DraftDragType

export type DraftDragType = 'internal' | 'external';
```

Then,

```javascript
const dragType = this._internalDrag ? 'internal' : 'external';
if (this.props.handleDrop && this.props.handleDrop(dropSelection, data, dragType) {
  ...
}
```

I do something similar for `DraftRemovalDirection`.

In truth I don't like boolean return values on these handlers either, so I should really clean up at some point too. :)

hellendag:
Thanks!

@facebook-github-bot import

Issue #146, opened 2016-03-03
getDirectionMap performance (large documents)


angelf:
I am investigating performance with large documents (thousands of blocks) which is a bit sluggish. The first thing I see is that every keystroke is triggering a call to `getDirectionMap`, which recalculates the direction for each node, and updates a map. The biggest overhead is on re-creating the OrderedMap (which most of the times is exactly the same as before)

<img width="424" alt="screen shot 2016-03-03 at 15 27 19" src="https://cloud.githubusercontent.com/assets/121845/13497412/72210df0-e155-11e5-8b85-c9f9f4287d59.png">

I am wondering if there is any way to avoid this memory allocation by only recalculating (and updating) the direction for blocks that have changed. Or to completely bypass this step if the editor has been initialized with textAligment. 

A similar situation is seen on `regenerateTreeForNewBlocks`. It ends with ends with a `prevTreeMap.merge(changed);`, but on most of the edits a single block has changed, so it'd be most efficient to just do a .set()

hellendag:
Curious -- how many thousands of blocks are you testing with?

This issue might be similar to what we're seeing in #63. Can you provide the profile file? I'd like to investigate it. I wouldn't expect it to be so expensive to perform these kinds of operations, so there may be a perf bottleneck within immutable-js.

`OrderedMap` is also known to be a bit expensive, so I'm sure `EditorBidiService.getDirectionMap` could be optimized, too. In most cases, of course, block direction won't change at all. We should probably only update direction for blocks that have changed, then continue looking until we find an unchanged direction. (Since `UnicodeBidiService` uses the direction of the previous block to determine direction of otherwise neutral blocks.) When we find an unchanged direction, we're done. The results can then be applied.

Regarding `regenerateTreeForNewBlocks`, we should actually only be generating new tree maps for changed `ContentBlock` objects, so I don't think it's quite the same. And I'm not sure whether using `merge` should be a problem there instead of `set`, as I don't expect them to have totally different perf profiles under the hood. That said, if you're seeing perf issues there, then there's something there to solve. :) Do you have a profile for this too?

angelf:
It is indeed a pretty monstrous document (7025 blocks). I've uploaded the CPU Profile to https://s3.amazonaws.com/vlex-interchange/getDirectionMap.cpuprofile

This is within a larger app so some of the stuff is really our fault, but I am pretty sure the getDirectionMap lag is not caused by us.  

[...]

PR #156, opened 2016-03-04
Only calculate directions for changed blocks && followers

angelf:
Fixes https://github.com/facebook/draft-js/issues/146, a significant lag when editing large documents, caused by every keystroke triggering a full re-calculation of the direction of every block.

It now only calculates direction for blocks that have changed or are right after a block whose direction changed.

### DIFF
--- a/src/model/immutable/EditorBidiService.js
+++ b/src/model/immutable/EditorBidiService.js
@@ -27,25 +27,52 @@ var bidiService;
 var EditorBidiService = {
   getDirectionMap: function(
     content: ContentState,
-    prevBidiMap: ?OrderedMap
+    prevBidiMap: ?OrderedMap,
+    prevContent: ?ContentState,
   ): OrderedMap {
+    if (content === prevContent) {
+      return prevBidiMap;
+    }
+
     if (!bidiService) {
       bidiService = new UnicodeBidiService();
     } else {
       bidiService.reset();
     }
 
-    var blockMap = content.getBlockMap();
-    var nextBidi = blockMap
-      .valueSeq()
-      .map(block => nullthrows(bidiService).getDirection(block.getText()));
-    var bidiMap = OrderedMap(blockMap.keySeq().zip(nextBidi));
+    var newBlockMap = content.getBlockMap();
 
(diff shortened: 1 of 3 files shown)


src/model/immutable/EditorBidiService.js
```diff
@@ -27,25 +27,52 @@ var bidiService;
...
+      // only update direction for blocks that have changed, then continue looking until we find an unchanged direction.
+
+      var prevLineChanged = false;
+      var prevBlockMap = prevContent.getBlockMap();
+
+      var hasChanged = false;
+
+      var newBidiMap = newBlockMap
+        .map((block, key) => {
+          if (prevLineChanged || block != prevBlockMap.get(key)) {
```
hellendag: I think there are some scenarios where this could fail. One possibility:

```
hello

                  שלום
```

At this point, the second block (which is empty) will be LTR, since it is neutral and depends on the block above. If that block were moved *entirely* below the second block but remained unchanged, it should become RTL by inheriting the direction of the block above it. All three blocks remain otherwise the same.

The conditional here would then never pass, so we would always preserve the direction value from `prevBidiMap`, and the empty block would incorrectly remain LTR.

How expensive is it to map `UnicodeBidiService` over the entire `blockMap`? If it's pretty cheap, then I wonder if we just always do that, keep its results in a regular object, iterate over the existing `bidiMap` and look into the bidi results to find differences, and selectively apply them with set/merge. If any blocks have changed order or been inserted, then we could just do the same `zip` as before. In most cases, it would be an improvement.

angelf: Yep, you're right. Let me try it and I'll get back.

hellendag:
Hey, just wanted to check whether you still plan to update this. :)

mitermayer:
Also interested on the outcome of this PR.

angelf:
Sorry @hellendag ! I've busy with other stuff and didn't see your comment. I do hope to get something back in about 2 weeks.

lacker:
It looks like this pull request has been abandoned so I am going to close it. Please feel free to reopen if you are still planning on working on this!

PR #157, opened 2016-03-04
Add support for block-level metadata

angelf:
It's now possible to associate entities to a whole block. These are
retrieved with block.getBlockEntity()

When spliting a block, the entity only remains linked to the first block
of the split

### DIFF
--- a/src/model/immutable/__tests__/ContentBlock-test.js
+++ b/src/model/immutable/__tests__/ContentBlock-test.js
@@ -84,7 +84,14 @@ describe('ContentBlock', () => {
     });
   });
 
-  describe('entity retrieval', () => {
+  describe('block entity retrieval', () => {
+    it('must retrieve block-level entities', () => {
+      var block = getSampleBlock().set('blockEntity', ENTITY_KEY);
+      expect(block.getBlockEntity()).toBe(ENTITY_KEY);
+    });
+  });
+
+  describe('caracter entity retrieval', () => {
     it('must properly retrieve entity at offset', () => {
       var block = getSampleBlock();
       expect(block.getEntityAt(0)).toBe(ENTITY_KEY);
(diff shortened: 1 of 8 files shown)


oyeanuj:
@angelf Thanks for this PR! Would it be possible to add to the documentation explaining this addition in a little bit more detail?

hellendag:
@facebook-github-bot import

src/model/immutable/__tests__/ContentBlock-test.js
```diff
@@ -84,7 +84,14 @@ describe('ContentBlock', () => {
...
 
-  describe('entity retrieval', () => {
+  describe('block entity retrieval', () => {
+    it('must retrieve block-level entities', () => {
+      var block = getSampleBlock().set('blockEntity', ENTITY_KEY);
+      expect(block.getBlockEntity()).toBe(ENTITY_KEY);
+    });
+  });
+
+  describe('caracter entity retrieval', () => {
```
gscottolson: `caracter` should be `character`

gscottolson:
This is awesome! Thanks, @angelf.

hellendag:
It's not currently clear how to apply or remove a block entity. Some `Modifier` methods would be helpful, and we might also need a new `EditorChangeType`.

SamyPesse:
`convertToRaw` is not working with blockEntity, the entity is not added to the map (https://github.com/facebook/draft-js/blob/3106fe50b845d3798575a9e34514172de2b0966e/src/model/encoding/convertFromDraftStateToRaw.js#L41).

I'll be happy to propose a PR to fix this, but I personally think that #216 is a better solution for "Block level metadatas".

angelf:
I'm happy with #216 approach. I actually find it more natural. @hellendag can you offer your thoughts on why you suggested to go with the Entity store?

hellendag:
@angelf Mostly just to follow the same pattern as per-character entity tracking. But I'm sold on the approach in #216. :)

angelf:
All right! I'm closing this one then.

Issue #153, opened 2016-03-04
covert to Raw stores only default draft styles


dotcor:
This looks like it is intentional. 
Inside encodeInlineStyleRanges:
```
function encodeInlineStyleRanges(block) {
  var styleList = block.getCharacterList().map(function(c) {
    return c.getStyle();
  }).toList();
  var styles = Object.keys(DefaultDraftInlineStyle);
  var ranges = styles.map(function(style) {
    return getEncodedInlinesForType(block, styleList, style);
  });
  return Array.prototype.concat.apply(EMPTY_ARRAY, ranges);
}
```

customStyleMap styles are simply not saved.

A simple solution would be to pass the customStyleMap as a second parameter to convertToRaw(), since the customStyleMap lives only inside the react component.

hellendag:
Nice catch, this is an oversight. Do you want to send a PR?

hellendag:
cc @ezequiel

PR #158, opened 2016-03-04
accept customStyleMap in convertToRaw()

dotcor:
Bug: only the default styles are saved when the content is 'exported'.


hellendag:
I think rather than passing in the `customStyleMap` as an argument, we can probably just use information within the `ContentBlock` itself to get the style ranges. Until recently, inline styles were not represented as `OrderedSet<string>`, so we needed to be able to get information about each available style from a single source.

Now, since styles are just strings, we can look at the `ContentBlock` and use the string values themselves.

hellendag:
Resolves #153.

hellendag:
Let me know if you need any support on this. :)

hellendag:
#217 addresses this.

Issue #74, opened 2016-02-24
The final space is `nbsp` (ascii 160), internal spaces are ascii 20


rcoh:
if you type `a b ` and call `editorState.getCurrentContent().getPlainText()` the internal space is a normal space but the final space is ascii 160. This causes problems if you want to match this text against other strings. I fixed it externally with the code below, but maybe this fix should be handled internally?

```
  cleanPlainText(editorState) {
    const text = editorState.getCurrentContent().getPlainText();
    const nbsp = String.fromCharCode(160);
    const regex = new RegExp(`${nbsp}`, 'g');
    return text.replace(regex, ' ');
  },
```

hellendag:
Hmm, this doesn't repro for me. You've typed `a`, ` `, `b`, correct? I'm not sure why there would be an `&nbsp;` there as well.

rcoh:
There is a space after the b and that's the one that comes out as the incorrect ascii code. I'll see if I can make a tester tonight.

hellendag:
Ah, so you typed `a`, ` `, `b`, ` `? And the two spaces are not the same character?

rcoh:
exactly

hellendag:
Which browser and OS?

rcoh:
chrome 48.0.2564.109, OS X 10.11.2, 

If it isn't reproducing elsewhere, perhaps it's a weird artifact of how I was using the control.

hellendag:
Are you able to determine whether the space bar character insertions are occurring via `editOnBeforeInput`?

hellendag:
@rcoh just wanted to follow up on this. I haven't seen this issue before, but I'd certainly like to fix it if it's widespread.

MindRave:
I think I'm seeing the same thing here. Chrome 48.0.2564.116 (64-bit), OSX 10.11.2. The last space is an &nbsp for me as well.

<img width="880" alt="screen shot 2016-03-02 at 23 51 17" src="https://cloud.githubusercontent.com/assets/4063406/13478640/41ae3298-e0d2-11e5-8116-ea4019dfa178.png">

hellendag:
You didn't type that with an <kbd>option</kbd>+<kbd>spacebar</kbd>, right? Just regular spaces?

MindRave:
Yes, just regular spaces.

hellendag:
If you add more spaces, does it add more `&nbsp;` characters? Or does that one continue to stick to the end of the text node?

[...]

Issue #151, opened 2016-03-04
Problem with Tweet example in own setup


LucidityDesign:
Hi,
first of all, the tweet example works fine when I download it and start it directly, so it must be a problem with my setup.

I've got the following Code:

```
import React from 'react';
import ReactDOM from 'react-dom';
import {CompositeDecorator, convertToRaw, Editor, EditorState, RichUtils} from 'draft-js';

'use strict';

class TweetEditorExample extends React.Component {
  constructor() {
    super();
    const compositeDecorator = new CompositeDecorator([
      {
        strategy: handleStrategy,
        component: HandleSpan,
      },
      {
        strategy: hashtagStrategy,
        component: HashtagSpan,
      },
    ]);

    this.state = {
      editorState: EditorState.createEmpty(compositeDecorator),
    };

    this.focus = () => this.refs.editor.focus();
    this.onChange = (editorState) => this.setState({editorState});
    this.logState = () => console.log(this.state.editorState.toJS());
  }

  render() {
    return (
      <div style={styles.root}>
        <div style={styles.editor} onClick={this.focus}>
          <Editor
            editorState={this.state.editorState}
            onChange={this.onChange}
            placeholder="Write a tweet..."
            ref="editor"
            spellCheck={true}
          />
        </div>
        <input
          onClick={this.logState}
          style={styles.button}
          type="button"
          value="Log State"
        />
      </div>
    );
  }
}

/**
 * Super simple decorators for handles and hashtags, for demonstration
 * purposes only. Don't reuse these regexes.
 */
const HANDLE_REGEX = /\@[\w]+/g;
const HASHTAG_REGEX = /\#[\w\u0590-\u05ff]+/g;

function handleStrategy(contentBlock, callback) {
  findWithRegex(HANDLE_REGEX, contentBlock, callback);
}

function hashtagStrategy(contentBlock, callback) {
  findWithRegex(HASHTAG_REGEX, contentBlock, callback);
}

function findWithRegex(regex, contentBlock, callback) {
  const text = contentBlock.getText();
  let matchArr, start;
  while ((matchArr = regex.exec(text)) !== null) {
    start = matchArr.index;
    callback(start, start + matchArr[0].length);
  }
}

const HandleSpan = (props) => {
  return <span {...props} style={styles.handle}>{props.children}</span>;
};

const HashtagSpan = (props) => {
  return <span {...props} style={styles.hashtag}>{props.children}</span>;
};

const styles = {
  root: {
[...]

PR #159, opened 2016-03-04
Move critical editor style into JS

hellendag:
There have been a number of bug reports caused by Draft.css not being included. The main one is that the editor can start flipping out if `white-space: pre-wrap` is not applied.

Users should generally be including Draft.css, and should be aware of behavior gaps if they choose not to. Documenting this accordingly.


hellendag:
Resolves #74 and #151.

spicyj:
Unclear to me which styles are super necessary and which are just nice, but this seems like an improvement.

hellendag:
The text alignment/direction ones are pretty important too, and should probably also be represented inline. I hacked around with that a couple months ago and dropped it, but I don't remember if it was because I ran into problems.

hellendag:
@facebook-github-bot import

PR #165, opened 2016-03-06
add start & end to decorated component

nikgraf:
In order to know which part to replace in case of an @-mention it is nice to have these parameters. You can get them from the selection as well, but it's bit tricky as the same work already done in the `strategy` like we did here: https://github.com/nikgraf/draft-js-plugin-editor/blob/master/src/mentionPlugin/MentionSearch/index.js#L39


hellendag:
I forgot, was this to obtain selection information from the editor context? Would my suggestion at the bottom of my recent comment in #169 help solve this?

nikgraf:
While in my case the start & end would be enough I can see that having the whole editorState exposed to the strategy being useful. :+1:

nikgraf:
Should we close this PR then?

hellendag:
Yeah, let's go ahead and close it. I think exposing the `editorState` seems like it's worth exploring.

PR #175, opened 2016-03-08
Android support

cgestes:
This fixes some issues for chrome support on Android. (including crosswalk)
- [X] use the provided selection in editOnInput
- [X] remove former text provided in compositionstart

remaining issues:
- [ ] moving the selection while composition is active discard the composition
- [ ] calling restoreEditorDOM() close the keyboard

I believe this can be merged after review without waiting for the other fixes. The experience is already much improved.

### DIFF
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -40,6 +40,7 @@ const RESOLVE_DELAY = 20;
 let resolved = false;
 let stillComposing = false;
 let textInputData = '';
+let formerTextInputData = '';
 
 var DraftEditorCompositionHandler = {
   onBeforeInput: function(e: SyntheticInputEvent): void {
@@ -50,7 +51,8 @@ var DraftEditorCompositionHandler = {
    * A `compositionstart` event has fired while we're still in composition
    * mode. Continue the current composition session to prevent a re-render.
    */
-  onCompositionStart: function(): void {
+  onCompositionStart: function(e): void {
+    formerTextInputData = e.data;
     stillComposing = true;
   },
 
@@ -125,6 +127,9 @@ var DraftEditorCompositionHandler = {
     const composedChars = textInputData;
     textInputData = '';
 
+    const formerComposedChars = formerTextInputData;
+    formerTextInputData = '';
+
     const editorState = EditorState.set(this.props.editorState, {
       inCompositionMode: false,
     });
@@ -149,12 +154,28 @@ var DraftEditorCompositionHandler = {
     this.exitCurrentMode();
     this.removeRenderGuard();
 
+    let contentState = editorState.getCurrentContent();
+    let selection = editorState.getSelection();
+    if (formerComposedChars && selection.isCollapsed()) {
+      var anchorOffset = selection.getAnchorOffset() - formerComposedChars.length;
+      if (anchorOffset < 0) {
+        anchorOffset = 0;
+      }
+      const toRemoveSel = selection.merge({anchorOffset});
+      contentState = DraftModifier.removeRange(
+        editorState.getCurrentContent(),
+        toRemoveSel,
+        'backward',
+      );
+      selection = contentState.getSelectionAfter();
+    }
+
     if (composedChars) {
       // If characters have been composed, re-rendering with the update
       // is sufficient to reset the editor.
-      const contentState = DraftModifier.replaceText(
-        editorState.getCurrentContent(),
-        editorState.getSelection(),
+      contentState = DraftModifier.replaceText(
+        contentState,
+        selection,
         composedChars,
         currentStyle,
         entityKey
--- a/src/component/handlers/edit/editOnCompositionStart.js
+++ b/src/component/handlers/edit/editOnCompositionStart.js
@@ -21,6 +21,7 @@ var EditorState = require('EditorState');
 function editOnCompositionStart(e): void {
   this.setRenderGuard();
   this.setMode('composite');
+  this._onCompositionStart(e);
   this.update(
     EditorState.set(this.props.editorState, {inCompositionMode: true})
   );
--- a/src/component/handlers/edit/editOnCompositionStart.js
+++ b/src/component/handlers/edit/editOnCompositionStart.js
@@ -18,7 +18,7 @@ var EditorState = require('EditorState');
  * The user has begun using an IME input system. Switching to `composite` mode
  * allows handling composition input and disables other edit behavior.
  */
-function editOnCompositionStart(): void {
+function editOnCompositionStart(e): void {
   this.setRenderGuard();
   this.setMode('composite');
   this.update(
--- a/src/component/handlers/edit/editOnInput.js
+++ b/src/component/handlers/edit/editOnInput.js
@@ -21,7 +21,7 @@ var findAncestorOffsetKey = require('findAncestorOffsetKey');
 var nullthrows = require('nullthrows');
 
 var isGecko = UserAgent.isEngine('Gecko');
-
+var isAndroid = UserAgent.isPlatform('Android');
 var DOUBLE_NEWLINE = '\n\n';
 
 /**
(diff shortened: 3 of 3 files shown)


src/component/handlers/edit/editOnInput.js
```diff
@@ -40,7 +40,7 @@ function editOnInput(): void {
   var domSelection = global.getSelection();
 
   var {anchorNode, isCollapsed} = domSelection;
-  if (anchorNode.nodeType !== Node.TEXT_NODE) {
+  if (anchorNode.nodeType !== Node.TEXT_NODE && anchorNode.nodeType !== Node.ELEMENT_NODE) {
```
cgestes: is this ok?

src/component/handlers/edit/editOnInput.js
```diff
@@ -88,7 +88,7 @@ function editOnInput(): void {
 
   var anchorOffset, focusOffset, startOffset, endOffset;
 
-  if (isGecko) {
+  if (true || isGecko) {
```
cgestes: I need to find how to check for chrome under Android

src/component/handlers/composition/DraftEditorCompositionHandler.js
```diff
@@ -149,12 +154,30 @@ var DraftEditorCompositionHandler = {
     this.exitCurrentMode();
     this.removeRenderGuard();
 
+    let contentState = editorState.getCurrentContent();
+    let selection = editorState.getSelection();
+    //let selecti
+    if (formerComposedChars) {
+      if (selection.isCollapsed()) {
```
hellendag: Nit: Combine conditionals.

src/component/handlers/composition/DraftEditorCompositionHandler.js
```diff
@@ -149,12 +154,30 @@ var DraftEditorCompositionHandler = {
...
+        var anchorOffset = selection.getAnchorOffset() - formerComposedChars.length;
+        if (anchorOffset < 0)
+          anchorOffset = 0;
```
hellendag: FB styling: Always use braces.

src/component/handlers/edit/editOnCompositionStart.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013-present, Facebook, Inc.
+ * Copyright (c) 2013, 2016-present, Facebook, Inc.
```
hellendag: Revert

src/component/handlers/edit/editOnCompositionStart.js
```diff
@@ -18,9 +18,10 @@ var EditorState = require('EditorState');
  * The user has begun using an IME input system. Switching to `composite` mode
  * allows handling composition input and disables other edit behavior.
  */
-function editOnCompositionStart(): void {
+function editOnCompositionStart(e): void {
   this.setRenderGuard();
   this.setMode('composite');
+  this._onCompositionStart(e);
```
hellendag: I think this private method might have problems if the class is munged. Also, if `compositionstart` is firing while we're in `EDIT` mode, then I don't think this function call has meaning.

src/component/handlers/edit/editOnInput.js
hellendag: What's the scenario here? For Android text replacement, the `span` can be the anchor node?

src/component/handlers/edit/editOnCompositionStart.js
cgestes: compositionstart is only fired while we are in editmode from what I saw. the composition one has never been called while I tested.
How is it supposed to be called?

hellendag: Right. `compositionstart` is used as a flag to indicate that the editor should switch to `composite` mode. This swaps in `DraftEditorCompositionHandler` as the editor's handler bundle, and all subsequent events are handled there until the editor switches back to `edit` mode. So you shouldn't need to add any code here.

cgestes: Compositionstart can have a non-empty data event which indicate the part of text that is going to be replaced by composition. (on most major web browser). 

One alternative solution would be to stop caring about those events and diff the DOM at strategical points, updating the model based on content and not on half broken events. Not sure if it's easily doable.

src/component/handlers/edit/editOnInput.js
cgestes: No idea, it just happens. Will check more carefully next time I debug draft-js

src/component/handlers/edit/editOnCompositionStart.js
hellendag: > Compositionstart can have a non-empty data event which indicate the part of text that is going to be replaced by composition. (on most major web browser).

We could check for `e.data` here, and if a value is present, use `Modifier.removeRange` to track the modified state and remove the range from the model. It would probably have to be flagged as natively rendered content to avoid re-rendering the editor during the composition. Would that help solve the issue here?

> One alternative solution would be to stop caring about those events and diff the DOM at strategical points, updating the model based on content and not on half broken events.

I'd prefer not to do DOM diffing.

Perhaps we could do something similar to the spellcheck handling in `editOnInput` -- grab the text content from the DOM and replace the equivalent range in the model. No diffing needed, just accommodate the DOM change.

cgestes: Modifier.removeRange is called with the e.data value of compositionstart later in the current patch. I wonder what is best, the text is not removed while in composition, it just get potentially updated at the end.

Replace DOM content within the model was what I had in mind. Will check editOnInput why I have time.

I'll be off next week.

src/component/handlers/edit/editOnInput.js
hellendag: Were you able to get more details on this case?

hellendag:
Ok, I'm going to import this and do some testing.

hellendag:
@facebook-github-bot import

sugarshin:
+1

jaaberg:
What's the status on this one, @hellendag? Would be nice to improve the Android support.

hellendag:
Patching and testing.

src/component/handlers/edit/editOnInput.js
```diff
@@ -21,7 +21,7 @@ var findAncestorOffsetKey = require('findAncestorOffsetKey');
 var nullthrows = require('nullthrows');
 
 var isGecko = UserAgent.isEngine('Gecko');
-
+var isAndroid = UserAgent.isPlatform('Android');
```
hellendag: Can we be more specific about targeting browser engines? Or does this need to apply across Android?

src/component/handlers/composition/DraftEditorCompositionHandler.js
```diff
@@ -149,12 +154,28 @@ var DraftEditorCompositionHandler = {
     this.exitCurrentMode();
     this.removeRenderGuard();
 
+    let contentState = editorState.getCurrentContent();
+    let selection = editorState.getSelection();
+    if (formerComposedChars && selection.isCollapsed()) {
+      var anchorOffset = selection.getAnchorOffset() - formerComposedChars.length;
+      if (anchorOffset < 0) {
```
hellendag: When can this happen? Can this conditional be removed? It feels sketchy.

butchmarshall:
- [ ]  calling restoreEditorDOM() close the keyboard

It looks like calling restoreEditorDOM setState changes the active element from the div -> body -> div rapidly.  This causes the keyboard to close.

dannyshisler:
Is there any update on this? I could really do with this Android keyboard issue being resolved. Thanks

andyesp:
What is the status on this? Do you need help testing?

walter0331:
Hi, I also have some text composition issues on iOS devices, are these issues related?https://github.com/facebook/draft-js/issues/626

jan4984:
for adding myself to participants

StevenLangbroek:
@hellendag @cgestes is there any word on when this might land? it's blocking us from implementing Draft in our codebase.

pklingem:
We're also waiting anxiously for this, commenting for notifications, apologies for noise.

KevinMartin:
@pklingem If you want notifications without giving everyone "noise", you could just press the Subscribe button on the right sidebar of any issue...

pklingem:
thanks @KevinMartin I just clicked subscribe.

jan4984:
please some one merge this PR?

kenjiru:
I've ran into this issue when porting my app to Cordova. I was a bit surprised to discover that draft doesn't support Android. Now this issue is the only thing blocking us from releasing our app.

mitermayer:
Hi @cgestes , it looks like this pull request has been abandoned, so I am going to close it. If someone is still interested in working on this please feel free to reopen it!

PR #179, opened 2016-03-09
upgrade to React 15 & suppress contentEditableWarning

nikgraf:
```
npm WARN EPEERINVALID draft-js@0.2.0 requires a peer of react@^0.14.0 but none was installed.
npm WARN EPEERINVALID draft-js@0.2.0 requires a peer of react-dom@^0.14.0 but none was installed.
```


package.json
```diff
@@ -31,8 +31,8 @@
     "immutable": "^3.7.4"
   },
   "peerDependencies": {
-    "react": "^0.14.0",
-    "react-dom": "^0.14.0"
+    "react": "^0.14.0 || 15.0.0-rc.1",
```
spicyj: Can we do `^0.14.0 || ^15.0.0-rc` which will catch the RCs and all subsequent 15.x stable versions?

nikgraf:
@spicyj good point, updated it

spicyj:
@facebook-github-bot shipit

PR #180, opened 2016-03-09
Updates Quick Start examples to run as expected

kennethormandy:
Hey! Thanks so much for the hard work on this project. I’m just making my way through the documentation now, and I couldn’t get the Quick Start examples to run until I made the following changes.

I have very little React experience, and even less with ES2015 stuff—it’s entirely possible I just misunderstood something in all these cases. Regardless, the examples didn’t work as-is for me, and to the best of my understand, these changes fix that.

The one change I made that was maybe more of an opinion was turning:

```js
const {editorState} = this.state;
return <Editor editorState={editorState} onChange={this.onChange} />
```

…into:

```js
return <Editor editorState={this.state.editorState} onChange={this.onChange} />
```

…which just reduced the complexity for me at the beginning, and had more in common with the initial Input example.

### DIFF
--- a/docs/QuickStart-API-Basics.md
+++ b/docs/QuickStart-API-Basics.md
@@ -30,7 +30,7 @@ to provide information about the text that the user has written.
 class MyInput extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { value: '' }
+    this.state = {value: ''};
     this.onChange = (evt) => this.setState({value: evt.target.value});
   }
   render() {
@@ -66,7 +66,7 @@ import {Editor, EditorState} from 'draft-js';
 class MyEditor extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { editorState: EditorState.createEmpty() }
+    this.state = {editorState: EditorState.createEmpty()};
     this.onChange = (editorState) => this.setState({editorState});
   }
   render() {
--- a/docs/QuickStart-API-Basics.md
+++ b/docs/QuickStart-API-Basics.md
@@ -30,6 +30,7 @@ to provide information about the text that the user has written.
 class MyInput extends React.Component {
   constructor(props) {
     super(props);
+    this.state = { value: '' }
     this.onChange = (evt) => this.setState({value: evt.target.value});
   }
   render() {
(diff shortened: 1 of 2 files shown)


docs/QuickStart-API-Basics.md
```diff
@@ -30,6 +30,7 @@ to provide information about the text that the user has written.
 class MyInput extends React.Component {
   constructor(props) {
     super(props);
+    this.state = { value: '' }
```
hellendag: this.state = {value: ''};

hellendag:
You're right about the missing state initialization, that was an oversight on my part. Nice catch.

I don't think it's so bad to have the destructuring assignment, especially since we're using ES6 classes and even `import` syntax here. But since the goal of these docs is to get people up to speed quickly, I don't want anyone stressing about syntax too much. So I'm cool with the change.

docs/QuickStart-Rich-Styling.md
```diff
@@ -44,15 +44,16 @@ We can observe and handle key commands via the `handleKeyCommand` prop, and
...
 class MyEditor extends React.Component {
   constructor(props) {
     super(props);
+    this.state = { editorState: EditorState.createEmpty() }
     this.onChange = (editorState) => this.setState({editorState});
   }
   handleKeyCommand(command) {
-    const {editorState} = this.state;
-    const newState = RichUtils.handleKeyCommand(editorState, command);
+    const newState = RichUtils.handleKeyCommand(this.editorState, command);
```
tasti: `this.editorState` should be `this.state.editorState`

kennethormandy:
Thanks for the review. I made the changes you suggested, and change one more instance of:

```js
const {editorState} = this.state;
this.onChange(RichUtils.toggleInlineStyle(editorState, 'BOLD'));
```

…to:

```js
this.onChange(RichUtils.toggleInlineStyle(this.state.editorState, 'BOLD'));
```

…for consistency with the other places I changed that in the Quick Start guide.

hellendag:
Looks good, thanks.

hellendag:
@facebook-github-bot import

PR #189, opened 2016-03-10
Make some dist build changes

zpao:
- Inline `process.env.NODE_ENV === 'production'` checks
- Compress CSS
- Add copyright headers to dist files

It's not a *huge* difference on the size front but definitely better. And the copyright thing is something we should do regardless of the rest.
screen shot 2016-03-09 at 4 26 17 pm


hellendag:
@facebook-github-bot shipit

PR #193, opened 2016-03-11
Ensure we're using latest fbjs alpha

zpao:
Fresh installs will be already be using this but upgrades might not be, so lets just guarantee it


hellendag:
@facebook-github-bot shipit

Issue #124, opened 2016-02-29
Media Wrapper


tasti:
Follow up on the discussion in https://github.com/facebook/draft-js/pull/67 (cc: @nikgraf).

I think as a baseline implementation, there should be some sort of `MediaWrapper` to take care of `contentEditable` and `data-offset-key`.

Along with the `MediaWrapper`, I discussed with @hellendag ways to support selection/highlighting media blocks. For example, clicking on a media block would have an overlay above it showing that it's selected. Or, if a media block is between text and you use the arrow key to move towards it, then it would gain selection.

In terms of how `media` is passed in, I can't really see any use-cases of `blockRendererFn` for a block type that isn't `media`. Maybe Custom Block Components should be handled at the `Entity` level rather than at the `Block` level.

hellendag:
Maybe we should be calling the `media` block type something like `atomic` instead.

oyeanuj:
> In terms of how media is passed in, I can't really see any use-cases of blockRendererFn for a block type that isn't media. Maybe Custom Block Components should be handled at the Entity level rather than at the Block level.

@tasti What about rich embeds or link previews? Maybe they would behave the same way as media..

tasti:
@oyeanuj Yeah, those seem like they would behave as `media`/`atomic`.

If you look at `DraftBlockType` (https://github.com/facebook/draft-js/blob/master/src/model/constants/DraftBlockType.js), `media` is pretty much a fallback for anything that isn't built-in. So always checking `if (block.getType() === 'media') { ... }` seems redundant in `blockRendererFn` if that's always the case.

nikgraf:
@tasti what if someone wants to have something else than media? I'm working on stickers implementation that would have type `'sticker'`. Could be a subtype of `atomic/media`, but should have custom component.

hellendag:
The component can be anything you choose, but the block type could be the same. You could use an entity type to differentiate.

nikgraf:
Ahh, right. Makes sense :+1:

tasti:
I mean should the `blockRendererFn` API take in an `Entity` instead and then check `if (entity.getType() === 'sticker') { ... }`?

hellendag:
> I mean should the blockRendererFn API take in an Entity instead and then check if (entity.getType() === 'sticker') { ... }?

[...]

PR #200, opened 2016-03-11
Custom block component wrapper

tasti:
cc: #124 

This creates a wrapper around Custom Block Components. The wrapper provides the necessary props (`contentEditable`, `data-block`, `data-edtior`, `data-offset-key`) required to correctly render custom block components, without the need for a developer to manually set these props within their `blockRendererFn`.

Details:
- This is a breaking API change for those creating wrappers manually.
- The custom block component is wrapped by an element based on its `DraftBlockType` (selected in `getElementForBlockType`).
- The `editable` property is optional (used for setting `contentEditable` on the custom block component). If `editable` isn't set, a `contentEditable` prop will not be set on the custom component (i.e. it would just inherit from above)

### DIFF
--- a/docs/Advanced-Topics-Block-Components.md
+++ b/docs/Advanced-Topics-Block-Components.md
@@ -40,6 +40,7 @@ function myBlockRenderer(contentBlock) {
   if (type === 'media') {
     return {
       component: MediaComponent,
+      editable: false,
       props: {
         foo: 'bar',
       },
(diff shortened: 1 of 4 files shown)


hellendag:
Awesome! I think this will help eliminate a bunch of the selection issues people have been running into so far in their custom blocks.

Will let others take a look and offer feedback before importing.

hellendag:
@facebook-github-bot import

tasti:
@facebook-github-bot-9 pushed a commit that _should have_ closed this pull request

https://github.com/facebook/draft-js/commit/67c5e69499e3b0c149ce83b004872afdf4180463

nikgraf:
super cool, thx @tasti

jan4984:
@tasti It's a cool feature that I can wrapper blocks in a box now. But I have a problem.

Saying I have some blocks wrapped in a box(colored border 'div' wrapperTemplate), now I can not append another box because the new added blocks will be in current box. as you commented in code "Group contiguous runs of blocks that have the same wrapperTemplate".

All I can do is making another name for my wrapperTemplate now. I'm not happy to add more and more names for a element.

Could we add some information in block meta data to control the blocks wrapper group?

PR #201, opened 2016-03-12
Improve `html` styles

gscottolson:
When the browser is narrower than `.container` or taller than the content, `<html>` `background-color` is visible:

<p align="center">
  
</p>

_You can also see this issue if you decrease the browser’s zoom level._

**Changes:**
- De-dupe the `html` selector
- Move `body` `background-color` to `html`
- Remove `body` rule
- :abc: `html` properties


hellendag:
@facebook-github-bot shipit

Issue #204, opened 2016-03-14
DraftBlockType has 'paragraph' as a type, however getElementForBlockType does not have paragraph reference


mitermayer:
getElementForBlockType is missing a paragraph definition

mitermayer:
added a PR for it https://github.com/facebook/draft-js/pull/205

Issue #52, opened 2016-02-23
Recommended way to create styles on-the-fly?


brandondurham:
In a previous project similar to Draft.js I used JSS to create a dynamic stylesheet that I could manage classes and styles in real-time. Primarily for things like changing text color to a user-selected color, or changing font-size to something the user selects. This allowed for a lot of flexibility, but also meant that the styles didn’t have to be created up front. They were created on-the-fly, as the user selected them.

Is there a recommended way to create classes and/or styles (mostly inline) on-the-fly and applying to the selection?

hellendag:
Take a look at the "Color" editor example. This example should give you an idea of how inline styles can be named and defined via the `customStyleMap` prop. In the "Color" example, the styles are treated as mutually exclusive (since it's font color) but inline styles do not have to be mutually exclusive in this way.

For instance, we could include something like this within the `customStyleMap` with color values in it:

```
small: {
  fontSize: '8px',
},
medium: {
  fontSize: '12px',
},
large: {
  fontSize: '16px',
},
```

Then, by using `RichUtil`, you could add these styles to ranges as needed. Characters could then have styles like `OrderedSet(['small', 'red'])` or `OrderedSet(['large', 'blue'])`.

At render time, these style values will be merged. In this way, you can define distinct inline styles and let React (more specifically, `DraftEditorLeaf`) worry about merging the values together.

Does this help?

brandondurham:
I think I’ve got that part figured out thus far, and it’s working wonderfully. My question is more about the CSS property values themselves, and whether or not they can be dynamic.

Here’s the thing: I’m working with many hundreds of font styles (~700) that each require their own CSS properties (`font-family`, `font-style` and `font-weight`). I _could_ create the styles for every possible font file that _may_ be loaded up front, but that would clearly be a lot more data for the user to load than is really necessary.

So, I’m wondering if there is a way to make the _value_ of the CSS property dynamic, and only set when requested by the user?

For instance, say I have a select menu with 100 different fonts in it that the user can choose from. To avoid preloading 100 fonts worth of CSS classes/properties, I’d like to only create and apply them when requested. So when selecting “Fancy Font Number 5” in the menu I need to:

- Create the CSS class and properties (`font-family`, `font-style` and `font-weight`);
- Load the font file with a Webpack require; and
- Apply the CSS class to the selection.

Can I add styles to the `customStyleMap` object _after_ the Editor is already mounted?

[...]

PR #205, opened 2016-03-14
adding-support-for-paragraph - making blockTypes configurable

mitermayer:
This fixes issue https://github.com/facebook/draft-js/issues/204

- Including paragraph on converter
- Updating website and demo
- Updating example


examples/rich/rich.html
```diff
@@ -157,8 +157,13 @@
       }
 
       const BLOCK_TYPES = [
+        {label: 'p', style: 'paragraph'},
         {label: 'H1', style: 'header-one'},
         {label: 'H2', style: 'header-two'},
+        {label: 'H3', style: 'header-three'},
+        {label: 'H4', style: 'header-four'},
```
mitermayer: added the extra missing headers from the rich example

hellendag:
cc @CLowbrow -- will it be a problem to have `p` map to `paragraph` in the paste processor?

mitermayer:
will let you guys decide, in the meantime i noticed some tests are broken fixing them now and amending the PR. Will move the missing headers from this PR to the other PR since it makes more sense in https://github.com/facebook/draft-js/pull/207 then here

CLowbrow:
@hellendag stuff broke with some renderers when we added the H4/5/6 stuff right? I'm worried the same stuff is going to happen here. Do our renderers know how to deal with this type? I'm half-remembering all of this, though. 

Right now 'unstyled' basically means paragraph. We can change it, but then what would unstyled be? Is it redundant?

mitermayer:
@CLowbrow could unstyled mean divs ? I would imagine Divs to be the most generic block container.

src/model/paste/__tests__/DraftPasteProcessor-test.js
```diff
@@ -130,8 +130,8 @@ describe('DraftPasteProcessor', function() {
   });
```
mitermayer: Updated 'p' to become 'div' since we were using 'p' with the assumption that would default to unstyled

spicyj:
`p` tags tend to be pretty annoying since browsers don't let you put any block-level tags (ex: `div`) inside of them. I've sworn off using `<p>` tags whenever I have a choice (opting instead for `<div class="paragraph">` or similar) and recommend others do too. Is there a reason that they're necessary here?

spicyj:
(I guess we already have the reverse mapping though?)

mitermayer:
@spicy, I agree to some extent to that. Paragraphs can definitely be an annoyance. However they provide semantics and also makes a lot easier on moving to draftjs from existing applications. Another option would be to make this behaviour configurable. We could default to paragraphs being unstyled but however allow the option to include and treat them as block types

CLowbrow:
The current behavior is a little complicated. We basically ignore divs in the html structure unless there are no semantic elements. HTML will often contain something like `<div><blockquote> hi </blockquote></div>` and since we are flattening the tree we want to use the most important element in that case.

We apply unstyled tags to the content of <p> tags and to blocks that don't have any of the tags we look for. For example `<h1> hello </h1> text <h2> goodbye</h2>` and `<h1> hello </h1> text <h2> goodbye</h2>` and `<h1> hello </h1><p> text <p><h2> goodbye</h2>' and `<div><h1> hello </h1><div> text <h2> goodbye</h2> will all generate the same block structure. 

Since, generally, we don't want to treat divs as meaningful tags, I don't think we need two types that mean text block. 

I might be missing something though. Can you provide an example of the case you are trying to solve for?

mitermayer:
One example is that if you currently copy and paste unsupported blocks it by default get's rendered as unstyled (which is currently used as paragraphs), which makes harder to export documents where we want to be able to correctly identify which elements where meant to be real paragraphs and which elements where just unsupported block elements.

CLowbrow:
That makes sense. 

@hellendag I think this is taking us back to the "pass a map of block tag names to types" thing. I think the best solution is for this to be configurable with minimal defaults. Thoughts?

mitermayer:
I like the idea of being configurable, happy to amend the PR if you guys reckon that's the way to go

hellendag:
> @hellendag stuff broke with some renderers when we added the H4/5/6 stuff right? I'm worried the same stuff is going to happen here. Do our renderers know how to deal with this type? I'm half-remembering all of this, though.

Yeah, they did. It was my fault though, I didn't sanitize it properly. `paragraph` is a bit more worrying to me, though, since we really just don't even use that type.

> @hellendag I think this is taking us back to the "pass a map of block tag names to types" thing. I think the best solution is for this to be configurable with minimal defaults. Thoughts?

Yep. I think this is the right thing to do. Configurable block mapping, inline style mapping, and link extraction.

mitermayer:
@hellendag @CLowbrow, been thinking through and I really see the benefits of having block mapping, inline styles and link extractions as a configurable configuration. 

1. Would allow consumers to create their custom block maps and would allow draft-js to not have to maintain new blocks and styles, would help on avoiding library bloating.
2. Would also make upgrades a lot less risky, since there would be no surprises, as each consumer would know exactly which styles and block elements they support based on their configuration setting.

In order for this to work, I would suggest that configurations should be explicit overwrites, this would allow us to either have a large default configuration on draft-js or just a very simple basic one.

My recommendation is keeping draft-js defaults for block mapping, inline style mapping very basic and instead move most of the advanced blocks and inline styles as configuration to the examples.

What are your thoughts on it ?

hellendag:
I definitely agree. Would you like to give it a try for this PR?

> My recommendation is keeping draft-js defaults for block mapping, inline style mapping very basic and instead move most of the advanced blocks and inline styles as configuration to the examples.

Yes. A sensible default is important and I think it makes sense to keep our existing mapping as that default.

mitermayer:
Sounds good, will amend this PR. The scope will be to make the current:

- Make it configurable settings
- Make the current settings the default settings
- Add tests and update documentation
- Update an example or create a new one to use configuration settings.

mitermayer:
Just commenting in here to inform that this PR is not dead yet, working on the above decision. Since it involves major internal changes is taking a little longer, But will update this ASAP

examples/rich/rich.html
```diff
@@ -89,6 +89,17 @@
             }
           }
 
+          var customBlockMap = Object.assign({}, DefaultDraftBlock, {
```
mitermayer: Should I update this to use Immutable ?

src/Draft.js
```diff
@@ -53,6 +55,9 @@ const DraftPublic = {
   Modifier: DraftModifier,
   RichUtils: RichTextEditorUtil,
 
+  DefaultDraftBlock,
```
mitermayer: I am also exposing now DefaultDraftInlineStyle for consistency and because we will need that once we make inline styles also configurable

src/component/base/DraftEditor.react.js
```diff
@@ -280,6 +281,9 @@ class DraftEditor
             <DraftEditorContents
               blockRendererFn={nullthrows(this.props.blockRendererFn)}
               blockStyleFn={nullthrows(this.props.blockStyleFn)}
+              customBlockMap={
+                {...this.props.customBlockMap}
```
mitermayer: As discussed, we treat customBlocks as a direct overwrite as opposed to extend, we should do the same for inlineStyles when we make them configurable

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,76 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+      const notUnstyledTag = tag !== draftBlockMap.unstyled.element;  // we do not want unstyled to be viewed as a valid match
+      const uniqueTag = matchedTags.indexOf(tag) === index;  // this will guarantee no duplicate tags added
+      return notUnstyledTag && uniqueTag;
+  });
+
+  return supportedTags;
+}
+
+// custom element conversions
+function getMultiMatchedType(
```
mitermayer: This will be useful in the future, and we should think about how could we expose as an API, this would allow applications to define which blockTypes should be returned if there are multiple blockTypes for the same html tag, example 'iframe with internal and external url sources' or anything like that. We could expose a way to add functions to a set of transformations array

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,76 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+  // if we have one matched type return it
+  // if we have multi matched types use the multi-match function to determine which type is it
+  switch(matchesFound) {
+    case 0:
+        return UNSTYLED;
+    case 1:
+        return matchedTypes[0];
     default:
-      return 'unstyled';
+        return getMultiMatchedType(tag, lastList, [getListBlockType]) || UNSTYLED;
```
mitermayer: One of the assumptions of allowing users to define a Custom block map, is that they have to always have a 'unstyled' block map.

src/model/immutable/DefaultDraftBlock.js
```diff
@@ -0,0 +1,72 @@
...
+
+const React = require('React');
+
+const cx = require('cx');
+
+const UL_WRAP = <ul className={cx('public/DraftStyleDefault/ul')} />;
+const OL_WRAP = <ol className={cx('public/DraftStyleDefault/ol')} />;
+const PRE_WRAP = <pre className={cx('public/DraftStyleDefault/pre')} />;
+
+module.exports = {
```
mitermayer: We have element as a 'tag element' and wrapper as a reactElement wrapper, any suggestions in here ?

gulpfile.js
```diff
@@ -187,6 +187,10 @@ gulp.task('watch', function() {
   gulp.watch(paths.src, ['modules']);
 });
 
+gulp.task('dev', function() {
```
mitermayer: Having a task that also do dist is very important for speed up development specially when doing tdd or exploratory testing using one of the example html files :)

src/model/immutable/DefaultDraftBlock.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule DefaultDraftBlock
```
mitermayer: Could not come up with a better name for it, DefaultDraftBlock ideally DraftBlockMap, but we already use that for other things

examples/rich/rich.html
hellendag: There is an issue open for this: https://github.com/facebook/draft-js/issues/52. @sstur is going to take a crack at it.

gulpfile.js
hellendag: If you don't mind, let's make this a separate PR. :)

src/model/immutable/DefaultDraftBlock.js
hellendag: How about `DraftBlockRenderMap`? Since this configuration is related to how types correspond to their rendered wrappers and elements.

src/model/immutable/DefaultDraftBlock.js
```diff
@@ -0,0 +1,72 @@
...
+
+const UL_WRAP = <ul className={cx('public/DraftStyleDefault/ul')} />;
+const OL_WRAP = <ol className={cx('public/DraftStyleDefault/ol')} />;
+const PRE_WRAP = <pre className={cx('public/DraftStyleDefault/pre')} />;
+
+module.exports = {
+  'header-one': {
+    wrapper: null,
+    element: 'h1'
+  },
```
hellendag: These objects could also have a Flow type definition, to enforce their usage elsewhere in code.

```javascript
export type DraftBlockRenderConfig = {
  wrapper: ?Object;
  element: string;
};
```

src/model/paste/__tests__/DraftPasteProcessor-test.js
```diff
@@ -50,25 +89,25 @@ describe('DraftPasteProcessor', function() {
...
-      [],
-      [],
+    ['ITALIC'],
+    ['ITALIC'],
+    ['ITALIC'],
+    ['ITALIC'],
+    ['ITALIC'],
+    [],
+    [],
+    [],
```
hellendag: Why remove the indentation here?

src/model/immutable/DraftBlockMap.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @providesModule DraftBlockMap
+ * @flow
+ */
+
+'use strict';
+
+import type Immutable from 'immutable';
+
+export type DraftBlockMap = Immutable.OrderedSet<string>;
```
hellendag: This type seems wrong. Isn't this more like a `Map<string, DraftBlockRenderConfig>`?

src/component/base/DraftEditor.react.js
hellendag: Yep. I think that's the right thing to do.

examples/rich/rich.html
mitermayer: Should I leave as it is then ?

gulpfile.js
mitermayer: Agreed, will remove this on the Tech review amend commit, in the meantime already created a PR for it https://github.com/facebook/draft-js/pull/242

src/model/immutable/DefaultDraftBlock.js
mitermayer: Spot on! I really like that naming! Will amend that on the Tech review commit

src/model/immutable/DraftBlockMap.js
mitermayer: AHh good catch! you are right. Forgot about this one

src/model/paste/__tests__/DraftPasteProcessor-test.js
mitermayer: this was failing on eslint with invalid indentation

mitermayer:
Update the PR based on tech review

gulpfile.js
```diff
@@ -187,10 +187,6 @@ gulp.task('watch', function() {
   gulp.watch(paths.src, ['modules']);
 });
 
-gulp.task('dev', function() {
```
mitermayer: This is now on a separate Pull request

src/model/constants/DraftBlockType.js
```diff
@@ -27,7 +26,6 @@ export type DraftBlockType = (
   'unordered-list-item' |
   'ordered-list-item' |
   'blockquote' |
-  'pullquote' |
```
mitermayer: Paragraph and pullquote were never used / referenced and were causing static type errors using flow. If any application requires it it can then use the configurable API to include it as demonstrated on the rich.html example file

src/model/paste/__tests__/DraftPasteProcessor-test.js
hellendag: Weird. I guess I disagree with the linter on this.

examples/rich/rich.html
```diff
@@ -89,6 +89,17 @@
             }
           }
 
+          var customBlockMap = Object.assign({}, DefaultDraftBlockRenderMap, {
+            'paragraph': {
+                wrapper: null,
+                element: 'p'
```
hellendag: Sorry to nitpick: FB standard is 2 spaces for indentation. (I guess the linter doesn't look at these.)

src/component/handlers/edit/editOnPaste.js
```diff
@@ -157,10 +156,6 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
   }
 }
 
-function filterOutNewlines(str: string): boolean {
-  return str.length > 0;
-}
-
```
hellendag: I thought this was fixed in a different PR?

src/component/utils/getElementForBlockType.js
```diff
@@ -12,32 +12,20 @@
...
+const DefaultDraftBlockRenderMap = require('DefaultDraftBlockRenderMap');
+
 import type {DraftBlockType} from 'DraftBlockType';
+import type {DraftBlockRenderMap} from 'DraftBlockRenderMap';
+
+function getElementForBlockType(
+  blockType: DraftBlockType,
+  customBlockMap: DraftBlockRenderMap
+): string {
+  const draftBlockRenderMap = customBlockMap !== undefined ? customBlockMap : DefaultDraftBlockRenderMap;
```
hellendag: Is it actually necessary to check for `undefined`, given that there is a default prop for `customBlockMap`?

src/component/utils/getWrapperTemplateForBlockType.js
```diff
@@ -12,32 +12,19 @@
...
-    case 'unordered-list-item':
-      return UL_WRAP;
-    case 'ordered-list-item':
-      return OL_WRAP;
-    case 'code-block':
-      return PRE_WRAP;
-    default:
-      return null;
-  }
+  const draftBlockRenderMap = customBlockMap !== undefined ? customBlockMap : DefaultDraftBlockRenderMap;
```
hellendag: Same question here.

src/model/constants/DraftBlockTag.js
```diff
@@ -0,0 +1,31 @@
...
+  'h1' |
+  'h2' |
+  'h3' |
+  'h4' |
+  'h5' |
+  'h6' |
+  'li' |
+  'blockquote' |
+  'pre' |
+  'media'
```
hellendag: `media` or `figure`?

src/model/constants/DraftBlockType.js
hellendag: :+1:

hellendag:
Not quite done reviewing this yet, but I need to stop and come back to it. (I'm really not a fan of how GH code comments are handled individually...)

johanneslumpe:
@mitermayer This PR is going to help when we want to get custom wrapping for `DraftEditorBlock`s right? Like you mentioned, when you actually want a `p` for example.

src/model/immutable/DraftBlockRenderMap.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+'use strict';
+
+import type Immutable from 'immutable';
+
+import type {DraftBlockRenderConfig} from 'DraftBlockRenderConfig';
+import type {DraftBlockType} from 'DraftBlockType';
+
+export type DraftBlockRenderMap = Immutable.Map<DraftBlockType, DraftBlockRenderConfig>
```
hellendag: I don't think that `DefaultDraftBlockRenderMap` matches this Flow type, since it's not an Immutable Map. Also the objects (`{wrapper: ..., element: ...}`) don't seem to match the `DraftBlockRenderConfig` type.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,80 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+  // if we have one matched type return it
+  // if we have multi matched types use the multi-match function to determine which type is it
+  switch (matchesFound) {
+    case 0:
+      return UNSTYLED;
+    case 1:
+      return matchedTypes[0];
     default:
-      return 'unstyled';
+      return getMultiMatchedType(tag, lastList, [getListBlockType]) || UNSTYLED;
```
hellendag: Isn't this assuming that the multi-match case is always going to be a list?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,80 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+  tag: string,
+  lastList: ?string,
+  draftBlockRenderMap: DraftBlockRenderMap
+): DraftBlockType {
+  const blockTypes = Object.keys(draftBlockRenderMap);
+  const matchedTypes = blockTypes.filter(function(blockType) {
+    return draftBlockRenderMap[blockType].element === tag || draftBlockRenderMap[blockType].wrapper === tag;
+  });
+  const matchesFound = matchedTypes.length;
+  const UNSTYLED = 'unstyled';
```
hellendag: I wouldn't bother const'ing this, you can just return the string directly. I know it's used twice below, but meh.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,80 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+    // we do not want unstyled to be viewed as a valid match
+    const notUnstyledTag = tag !== draftBlockRenderMap.unstyled.element;
+
+    // this will guarantee no duplicate tags added
+    const uniqueTag = matchedTags.indexOf(tag) === index;
+
+    return notUnstyledTag && uniqueTag;
+  });
+
+  return supportedTags;
```
hellendag: If `DraftBlockRenderMap` is an Immutable Map, then you can do:

```javascript
const unstyledElement = draftBlockRenderMap.unstyled.element;
return draftBlockRenderMap
  .map((config) => config.element)
  .valueSeq()
  .toSet()
  .filter((tag) => tag !== unstyledElement)
  .toArray()
  .sort();
```

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,23 +118,80 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+  }
+
+  return null;
+}
+
+function getBlockMapSupportedTags(
+  draftBlockRenderMap: DraftBlockRenderMap
+): Array<string> {
+  const blockTypes = Object.keys(draftBlockRenderMap);
+  const matchedTags = blockTypes.map(function(blockType) {
```
hellendag: Ok to use fat-arrow functions.

src/component/utils/getElementForBlockType.js
```diff
@@ -12,32 +12,20 @@
...
-    case 'blockquote':
-      return 'blockquote';
-    case 'media':
-      return 'figure';
-    default:
-      return 'div';
-  }
+  return draftBlockRenderMap[blockType] && draftBlockRenderMap[blockType].element ?
+    draftBlockRenderMap[blockType].element :
+    draftBlockRenderMap.unstyled.element;
```
hellendag: What if `unstyled` does not have a defined element? Thinking this should probably fall back to `div`.

src/component/base/DraftEditorProps.js
```diff
@@ -128,4 +129,7 @@ export type DraftEditorProps = {
   // Provide a map of inline style names corresponding to CSS style objects
   // that will be rendered for matching ranges.
   customStyleMap?: Object,
+
+  // custom block maps
```
hellendag: Could be helpful to have this documentation be a little more fleshed out. I think we'll definitely want a proper article on draftjs.org too, to really clarify what all this configuration is, but a little blurb here would be good too.

examples/rich/rich.html
```diff
@@ -33,7 +33,7 @@
     <script type="text/babel">
       'use strict';
 
-      const {Editor, EditorState, RichUtils} = Draft;
+      const {Editor, EditorState, RichUtils, DefaultDraftBlockRenderMap, DefaultDraftInlineStyle} = Draft;
```
hellendag: Another nit, to avoid the 80-char limit internally (sorry):

```javascript
const {
  Editor,
  EditorState,
  // etc.
} = Draft;

File: src/model/encoding/convertFromHTMLToContentBlocks.js (hunk 22 above)
Comment (mitermayer) [inline src/model/encoding/convertFromHTMLToContentBlocks.js]: In this iteration yes, I will include a new PR that allows to extend that. No reason to make this PR even larger. But currently thats the only core use case

File: src/model/immutable/DraftBlockRenderMap.js (hunk 21 above)
Comment (mitermayer) [inline src/model/immutable/DraftBlockRenderMap.js]: Now its converted to properly be using the correct Immutable objects :)

**Comment by mitermayer** [comment]
Updated the PR with all the tech review comments, please let me know what you guys think. Also will update the drafjs.org documentation for this in a separate PR once this is merged @hellendag

File: examples/rich/rich.html [hunk 29]
```diff
@@ -89,6 +95,20 @@
...
+          const customDraftBlockRenderMap = Immutable.Map({
+            'paragraph': Immutable.Map({
+              wrapper: null,
+              element: 'p'
+            }),
+            'unstyled': Immutable.Map({
+              wrapper: null,
+              element: 'p'
+            })
+          });
```
hellendag: I think this would be more appropriate as a `state` property, to indicate how it might be used (and imply how it may be updated) in practice. Also, then we don't have to create it in every render pass. :)

src/component/utils/getElementForBlockType.js
```diff
@@ -12,32 +12,23 @@
...
-      return 'blockquote';
-    case 'media':
-      return 'figure';
-    default:
-      return 'div';
-  }
+  return matchedBlockTypeTag;
 }
 
 module.exports = getElementForBlockType;
```
hellendag: Can this whole function be eliminated, with the logic above simply done inline within `DraftEditorContents`? Same for `getWrapperTemplateForBlockType`.

I don't recall whether there are other callsites for those functions, but I'm all in favor of cleaning up modules we don't need. :)

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,31 +118,78 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+    return 'unordered-list-item';
+  }
+
+  return null;
+}
+
+function getBlockMapSupportedTags(
+  draftBlockRenderMap: DraftBlockRenderMap
+): Array<string> {
+  const unstyledElement = draftBlockRenderMap.get('unstyled').get('element');
```
hellendag: Or:

```javascript
const unstyledElement = draftBlockRenderMap.getIn(['unstyled', 'element']);
```

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -116,31 +118,78 @@ function getBlockDividerChunk(block: DraftBlockType, depth: number): Chunk {
...
+    }
+    return 'unordered-list-item';
+  }
+
+  return null;
+}
+
+function getBlockMapSupportedTags(
+  draftBlockRenderMap: DraftBlockRenderMap
+): Array<string> {
```
hellendag: Can this be `Array<DraftBlockTag>`?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -271,8 +324,8 @@ function genFragment(
   // BR tags
   if (nodeName === 'br') {
     if (
-        lastLastBlock === 'br' &&
-        (!inBlock || getBlockTypeForTag(inBlock, lastList) === 'unstyled')
+      lastLastBlock === 'br' && (!inBlock ||
+      getBlockTypeForTag(inBlock, lastList, draftBlockRenderMap) === 'unstyled')
```
hellendag: Spacing nit, just a personal preference really:

```javascript
if (
  lastLastBlock === 'br' &&
  (
    !inBlock ||
    getBlockTypeForTag(...) === 'unstyled'
  )
) {

**Comment by hellendag** [comment]
Nice! I think we're really close. :)

File: src/component/utils/getElementForBlockType.js (hunk 30 above)
Comment (mitermayer) [inline src/component/utils/getElementForBlockType.js]: That's true, also in favour of deleting things! will do that!

File: src/model/encoding/convertFromHTMLToContentBlocks.js (hunk 31 above)
Comment (mitermayer) [inline src/model/encoding/convertFromHTMLToContentBlocks.js]: Will convert to use that!

File: src/model/encoding/convertFromHTMLToContentBlocks.js (hunk 32 above)
Comment (mitermayer) [inline src/model/encoding/convertFromHTMLToContentBlocks.js]: Agreed, will update to it

File: src/model/encoding/convertFromHTMLToContentBlocks.js (hunk 33 above)
Comment (mitermayer) [inline src/model/encoding/convertFromHTMLToContentBlocks.js]: I don't mind it, we should probably add this as a rule to eslintrc so that the formatting can be picked up on the linting

File: src/model/encoding/convertFromHTMLToContentBlocks.js (hunk 32 above)
Comment (mitermayer) [inline src/model/encoding/convertFromHTMLToContentBlocks.js]: as discussed on slack the problem is because 

```
271 function containsSemanticBlockMarkup(
272   html: string,
273   blockTags: Array<DraftBlockTag>
274 ): boolean {
275   return blockTags.some(tag => html.indexOf('<' + tag) !== -1);
276 }
```

the string comparison of indexOf makes flow thing that the type for blockTags needs to be a string

so it fails as:

```
src/model/encoding/convertFromHTMLToContentBlocks.js:292
292:   blockTags: Array<string>,
                       ^^^^^^ string. This type is incompatible with
138: :disappointed: Array<DraftBlockTag> {
             ^^^^^^^^^^^^^ string enum
```

We decided then to leave as string type for now

File: src/model/encoding/convertFromHTMLToContentBlocks.js [hunk 34]
```diff
@@ -382,12 +447,22 @@ function getChunkForHTML(html: string, DOMBuilder: Function): ?Chunk {
   // Sometimes we aren't dealing with content that contains nice semantic
   // tags. In this case, use divs to separate everything out into paragraphs
   // and hope for the best.
-  var workingBlocks = containsSemanticBlockMarkup(html) ? blockTags : ['div'];
+  var workingBlocks = containsSemanticBlockMarkup(html, supportedBlockTags) ?
+    supportedBlockTags :
+    ['div'];
```
mitermayer: I wonder if the intention here would be to default to unstyled if so will also update it here

mitermayer: will leave as it is for this PR to avoid any further change since it is already big enough

hellendag:
Ok, let's do this. :)

hellendag:
@facebook-github-bot import

mitermayer:
awesome!

src/Draft.js
```diff
@@ -11,6 +11,8 @@
 
 'use strict';
 
+const DefaultDraftBlockRenderMap = require('DefaultDraftBlockRenderMap');
```
bradleyayers: FYI the rest of the imports are alphabetical order.

hellendag:
Sorry for the long wait. :) This is in review internally, and we're being thorough to make sure it works cleanly.

rnicholus:
Will this resolve #52? I've paged through the PR, but I'm not certain. #52 seems to define a very common integration of a color picker. Currently, it is not possible to integrate a traditional color picker into Draft.js. Will this PR address this limitation? If so, can someone explain how?

hellendag:
Okay, I'm going to land and merge this today.

I've made quite a few changes, so let me know if you notice anything missing or incorrect.

jpuri:
Hello @mitermayer : I am working on creating an editor using DraftJS and we were looking for something similar in supporting paragraph blocks. Thanks a lot for good work 👍 

Can you please guild me where can I get documentation for passing configuration parameters to the editor. I could not find that in the commit.

mitermayer:
@jpuri Sorry for not including the documentation on this PR it will be bundled on a separate PR with example.

In the meantime you can follow the reverse of this commit, https://github.com/facebook/draft-js/commit/575e051369e11ecb9a362e843f557acc36d934b7 this had an example on how to integrate paragraph to the rich.html demo and also how to overwritte the default 'unstyled' to become a paragraph.

jpuri:
Thanks a lot @mitermayer, but editor gives me this error with the configurations, should I just ignore it ?

screen shot 2016-05-12 at 11 58 03 am

mitermayer:
@jpuri Will talk via slack in order to reduce feedback loop and help debug whats the problem

jpuri:
Thanks @mitermayer

mitermayer:
Attached the article documentation for it on the PR https://github.com/facebook/draft-js/pull/387

Issue #195, opened 2016-03-11
Handle pasted text to create entities


hingermayank:
Hello

While pasting the content, we can get the pasted content in onPasteRawText but then again editOnPaste handles the pasted data. My usecase is something like this:

I have a input where when a user types a comma separated values, I convert them into entities(immutable) and decorate them to look like this which I have done.

<img width="522" alt="screen shot 2016-03-11 at 7 17 08 pm" src="https://cloud.githubusercontent.com/assets/6342096/13703715/ebee3824-e7bd-11e5-8dbf-4ea9b3c0a9de.png">

But the same steps need to be done to the pasted data. I might be missing something on how to handle the pasted data.

Any help would be appreciated.

Thanks

hellendag:
Hmm. `onPasteRawText` could possibly be a canceling handler, i.e. `handlePastedText`. Then you would be able to `return true` if you handle the paste yourself.

hingermayank:
Yes. Something like handleBeforeInput is a cancelling event.

hellendag:
I would be fine with this. Would you like to submit a PR to replace `onPasteRawText` with `handlePastedText`?

hingermayank:
Sure. I will do the changes and submit the PR.

avk:
@hingermayank thanks for adding `handlePastedText`. Are there are any examples of how to use it? [The docs](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handlepastedtext) are a little sparse...

avk:
Found an example of `handlePastedText` for anyone searching in the future: https://github.com/facebook/draft-js/issues/416#issuecomment-221639163

PR #209, opened 2016-03-14
Enhancement: Adding handlePastedText handler to handle pasted text in…

hingermayank:
As discussed in the issue #195, adding handlePastedText handler to handle pasted text in Editor and removing nullThrows import (unused) to fix lint warning.

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -165,7 +165,7 @@ and to convert typed emoticons into images.
 
 #### handlePastedText
 ```
-handlePastedText?: (text: string, html: string) => boolean
+handlePastedText?: (text: string, html?: string) => boolean
 ```
 Handle text and html(for rich text) that has been pasted directly into the editor.
 
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -165,9 +165,9 @@ and to convert typed emoticons into images.
 
 #### handlePastedText
 ```
-handlePastedText?: (text: string) => boolean
+handlePastedText?: (text: string, html: string) => boolean
 ```
-Handle text that has been pasted directly into the editor.
+Handle text and html(for rich text) that has been pasted directly into the editor.
 
 #### handlePastedFiles
 ```
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -163,6 +163,12 @@ convert that `ContentBlock` into an `unordered-list-item`.
 At Facebook, we also use this to convert typed ASCII quotes into "smart" quotes,
 and to convert typed emoticons into images.
 
+#### handlePastedText
+```
+handlePastedText?: (text: string) => boolean
+```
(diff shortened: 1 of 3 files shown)


hellendag:
Thanks!

It occurs to me that we _may_ also want to make the raw HTML available to the paste handler. It's a risk, though, since that would mean that one wouldn't be able to benefit from internal clipboard detection. Plus, of course, parsing HTML into Draft data structures is not an easy task.

I think what you have here is definitely the right thing to do, but curious to get @CLowbrow's thoughts on HTML before importing the PR.

hingermayank:
Sure. This use case of passing the raw HTML did not occur to me. But yes parsing the HTML limits the use case.

CLowbrow:
How about:  we expose one `handlePaste` prop that takes a function that takes two nullable arguments: text, and html. User can examine both objects. If they return false, we don't do anything. If they return true, we do our regular song and dance to see if there is an internal clipboard/html to use.

CLowbrow:
If we want to be extra aggro we can prepend `` to pasted html, but that doesn't seem necessary.

dburrows:
+1 for access to the raw text before parsing - my current use case for this would be removing formatting but retaining new lines/converting <br>'s in content pasted into a <pre><code> area.

In the scenario where i'm pasting text within the same document, from styled text into a code block, does this complicate things? Is the inter-document paste already parsed?

dburrows:
> So if you intercept at this point, it will not be aware of the internal paste.

Cool

> we expose one handlePaste prop that takes a function that takes two nullable arguments: text and html

That sounds like a good solution, best of both worlds

hellendag:
@hingermayank Want to update the PR to provide both `html` and `text`?

docs/APIReference-Editor.md
```diff
@@ -163,6 +163,12 @@ convert that `ContentBlock` into an `unordered-list-item`.
 At Facebook, we also use this to convert typed ASCII quotes into "smart" quotes,
 and to convert typed emoticons into images.
 
+#### handlePastedText
+```
+handlePastedText?: (text: string, html: string) => boolean
```
hellendag: `html` should be an optional parameter. It may be a plaintext paste with no HTML available.

```javascript
handlePastedText?: (text: string, html?: string) => boolean
```

src/component/handlers/edit/editOnPaste.js
```diff
@@ -86,8 +85,13 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
   }
 
   let textBlocks: Array<string> = [];
-  var text = data.getText();
-  this.props.onPasteRawText && this.props.onPasteRawText(text);
+  const text = data.getText(),
+    html = data.getHTML();
```
hellendag: FB styling nit:

```javascript
const text = data.getText();
const html = data.getHTML();
```

hellendag:
@facebook-github-bot import

Issue #185, opened 2016-03-09
Move Entity information into State


thesunny:
Related to this thread:

https://github.com/facebook/draft-js/issues/129

@angelf noted the weakness in using Entity objects, in particular they are not properly handled with Undo.

Moving Entity data into the ContentState should solve this issue.

Furthermore, in the referenced discussion, entity data is now also associated with ContentBlock information to store block data. This is in the associated PR.

https://github.com/facebook/draft-js/pull/157

By moving Entity into the state, we could do things like store `depth` information for lists in the Entity data and have it work with undo.

Feels like there are a few ways to move Entity data into the ContentState. Just to start the discussion.
1. Entity information can be stored as part of the ContentState.
2. Entity information can be stored as part of the ContentBlock (inside the ContentState).

The benefit of 1. is that it should behave fairly similarly to the existing implementation except the Entity data would be stored as part of the ContentState. I'm supposing one main difference is that there would have to be cleanup done on the entity objects within ContentState such that when an entity is no longer used, it needs to get deleted. Haven't looked at the implementation in Entity but I'm guessing that when using Entity the references are just kept around and "cleaned up" during the convertToRaw (i.e. unused entities are not part of the export).

The benefit of 2. is that cleanup could operate on much smaller chunks (a block at a time). For example, if a ContentBlock is deleted, all of the entities disappear with it so no additional cleanup work needs to be done. There would still need to be some cleanup if, for example, a link is deleted within a block, then the Entity is removed from the block. However, we don't need to look through the entire ContentState to do the cleanup.

hellendag:
I'm not really worried about entity cleanup, especially since entities will be present in memory for `ContentState` objects in the undo stack anyway, whether they're on the `ContentState` or the `ContentBlock`.

I would be in favor of putting entities in `ContentState` rather than `ContentBlock`. A simple reason that comes to mind: if I want to split a block, it also means I'd have to split entities into their appropriate blocks as well. Or if I want to move a range of text from one block to another, I'd have to move any entities along with it. If the map lives higher up in the `ContentState`, then there is no added complexity -- we can continue referring to entities purely by key.

Overall, though, I am in favor of a change. The global map made sense for the original pre-immutable model, but is now mostly a hindrance.

thesunny:
That's a good point about splitting blocks that I didn't think about.

Also, I guess we would still need the convertToRaw methods anyways and the cleanup could be done during that time so that unused entities aren't saved.

hellendag:
[...]

PR #216, opened 2016-03-16
block level metadata

cgestes:
This commit add support for block level metadata.

It is the same feature as https://github.com/facebook/draft-js/pull/157

This PR does not rely on entity, but just add a data field to ContentBlock.

hense block metadata are store in contentState and are supported by undo/redo.

Why not using entities?
- Entity are not made for block. what does type and mutability would map to?
- Entity are planned to be integrated into contentState anyway (see #185)

### DIFF
--- a/src/model/encoding/convertFromDraftStateToRaw.js
+++ b/src/model/encoding/convertFromDraftStateToRaw.js
@@ -50,6 +50,7 @@ function convertFromDraftStateToRaw(
       depth: canHaveDepth(block) ? block.getDepth() : 0,
       inlineStyleRanges: encodeInlineStyleRanges(block),
       entityRanges: encodeEntityRanges(block, entityStorageMap),
+      data: block.getData().toJS(),
     });
   });
 
--- a/src/model/immutable/ContentBlock.js
+++ b/src/model/immutable/ContentBlock.js
@@ -24,6 +24,7 @@ import type {DraftInlineStyle} from 'DraftInlineStyle';
 var {
   List,
   OrderedSet,
+  Map,
   Record,
 } = Immutable;
 
@@ -35,12 +36,14 @@ var defaultRecord: {
   text: string;
   characterList: List<CharacterMetadata>;
   depth: number;
+  data: Map;
 } = {
   key: '',
   type: 'unstyled',
   text: '',
   characterList: List(),
   depth: 0,
+  data: Map(),
 };
 
 var ContentBlockRecord = Record(defaultRecord);
@@ -68,6 +71,10 @@ class ContentBlock extends ContentBlockRecord {
 
   getDepth(): number {
     return this.get('depth');
+  }
+
+  getData(): Map {
+    return this.get('data');
   }
 
   getInlineStyleAt(offset: number): DraftInlineStyle {
--- a/src/model/modifier/DraftModifier.js
+++ b/src/model/modifier/DraftModifier.js
@@ -15,7 +15,7 @@
 
 var CharacterMetadata = require('CharacterMetadata');
 var ContentStateInlineStyle = require('ContentStateInlineStyle');
-var {OrderedSet} = require('immutable');
+var {OrderedSet, Map} = require('immutable');
 
 var applyEntityToContentState = require('applyEntityToContentState');
 var getCharacterRemovalRange = require('getCharacterRemovalRange');
@@ -211,6 +211,33 @@ var DraftModifier = {
       operation
     );
   },
+
+  setBlockData: function(
+    contentState: ContentState,
+    selectionState: SelectionState,
+    blockData: Map,
+  ): ContentState {
+    const operation = (block) => block.merge({data: blockData});
+    return modifyBlockForContentState(
+      contentState,
+      selectionState,
+      operation
+    );
+  },
+
+  mergeBlockData: function(
+    contentState: ContentState,
+    selectionState: SelectionState,
+    blockData: Map,
+  ): ContentState {
+    const operation = (block) => block.merge({data: block.getData().merge(blockData)});
+    return modifyBlockForContentState(
+      contentState,
+      selectionState,
+      operation
+    );
+  },
+
 
   applyEntity: function(
     contentState: ContentState,
--- a/src/model/modifier/DraftModifier.js
+++ b/src/model/modifier/DraftModifier.js
@@ -25,7 +25,7 @@ var insertTextIntoContentState = require('insertTextIntoContentState');
 var invariant = require('invariant');
 var removeEntitiesAtEdges = require('removeEntitiesAtEdges');
 var removeRangeFromContentState = require('removeRangeFromContentState');
-var setBlockTypeForContentState = require('setBlockTypeForContentState');
+var modifyBlockForContentState = require('modifyBlockForContentState');
 var splitBlockInContentState = require('splitBlockInContentState');
 
 import type {BlockMap} from 'BlockMap';
@@ -204,10 +204,11 @@ var DraftModifier = {
     selectionState: SelectionState,
     blockType: DraftBlockType
   ): ContentState {
-    return setBlockTypeForContentState(
+    const operation = (block) => block.merge({type: blockType, depth: 0});
+    return modifyBlockForContentState(
       contentState,
       selectionState,
-      blockType
+      operation
     );
   },
 
--- a/src/model/transaction/modifyBlockForContentState.js
+++ b/src/model/transaction/modifyBlockForContentState.js
@@ -6,23 +6,23 @@
  * LICENSE file in the root directory of this source tree. An additional grant
  * of patent rights can be found in the PATENTS file in the same directory.
  *
- * @providesModule setBlockTypeForContentState
+ * @providesModule modifyBlockForContentState
  * @typechecks
  * @flow
  */
 
 'use strict';
 
-var Immutable = require('immutable');
+import Immutable from 'immutable';
 
+import type ContentBlock from 'ContentBlock';
 import type ContentState from 'ContentState';
-import type {DraftBlockType} from 'DraftBlockType';
 import type SelectionState from 'SelectionState';
 
-function setBlockTypeForContentState(
+function modifyBlockForContentState(
   contentState: ContentState,
   selectionState: SelectionState,
-  blockType: DraftBlockType,
+  operation: (block: ContentBlock) => ContentBlock,
 ): ContentState {
   var startKey = selectionState.getStartKey();
   var endKey = selectionState.getEndKey();
@@ -32,7 +32,7 @@ function setBlockTypeForContentState(
     .skipUntil((_, k) => k === startKey)
     .takeUntil((_, k) => k === endKey)
     .concat(Immutable.Map([[endKey, blockMap.get(endKey)]]))
-    .map(block => block.merge({type: blockType, depth: 0}));
+      .map(operation);
 
   return contentState.merge({
     blockMap: blockMap.merge(newBlocks),
@@ -41,4 +41,4 @@ function setBlockTypeForContentState(
   });
 }
 
-module.exports = setBlockTypeForContentState;
+module.exports = modifyBlockForContentState;
--- a/src/model/transaction/splitBlockInContentState.js
+++ b/src/model/transaction/splitBlockInContentState.js
@@ -18,6 +18,7 @@ var invariant = require('invariant');
 
 import type ContentState from 'ContentState';
 import type SelectionState from 'SelectionState';
+import {Map} from 'immutable';
 
 function splitBlockInContentState(
   contentState: ContentState,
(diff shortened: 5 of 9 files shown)


src/model/modifier/DraftModifier.js
```diff
@@ -211,6 +212,18 @@ var DraftModifier = {
     );
   },
 
+  setBlockData: function(
+    contentState: ContentState,
+    selectionState: SelectionState,
+    blockData: DraftBlockType
```
hellendag: Surely this should be a `Map` or another collection of some kind, rather than a `DraftBlockType`?

src/model/immutable/ContentBlock.js
```diff
@@ -35,12 +35,14 @@ var defaultRecord: {
...
   characterList: List<CharacterMetadata>;
   depth: number;
+  data: Object;
 } = {
   key: '',
   type: 'unstyled',
   text: '',
   characterList: List(),
   depth: 0,
+  data: Object(),
```
hellendag: An immutable `Map` would be better than `Object` here, I think.

src/model/encoding/convertFromDraftStateToRaw.js
```diff
@@ -50,6 +50,7 @@ function convertFromDraftStateToRaw(
       depth: canHaveDepth(block) ? block.getDepth() : 0,
       inlineStyleRanges: encodeInlineStyleRanges(block),
       entityRanges: encodeEntityRanges(block, entityStorageMap),
+      data: block.getData(),
```
hellendag: With an immutable `Map`, this would probably need to use `toJS()`.

src/model/transaction/setBlockDataForContentState.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+'use strict';
+
+var Immutable = require('immutable');
+
+import type ContentState from 'ContentState';
+import type SelectionState from 'SelectionState';
+
+function setBlockDataForContentState(
```
hellendag: In what scenario might we want blocks to share the same metadata? (Not arguing against it, just thinking about what this might entail.)

src/model/transaction/splitBlockInContentState.js
```diff
@@ -46,6 +46,7 @@ function splitBlockInContentState(
     key: keyBelow,
     text: text.slice(offset),
     characterList: chars.slice(offset),
+    data: Object(),
```
hellendag: Should the lower block never keep the same `data`? It might depend on how the `data` field is being used. Thoughts?

angelf: I can see it both ways. 

For instance, if block data was used to store 'depth' (instead of a dedicated property), I guess the reasonable thing to do would be to keep it. But if data was used to store an anchor location, that would not be ideal. 

Might it be useful to keep it but offer a callback (that pops-up to DraftModifier.splitBlock) to allow manipulating the new block? Or is this overkill?

hellendag: That seems a little overkill. The callback would also have to be threaded all the way to this function, and it wouldn't match any other transaction or `Modifier` functions.

My instinct here is that the data should be preserved for both blocks, then the client can manually remove the data if it's not desired for the specific use case. It means performing an extra step, but there's no magic or configuration involved.

remko:
@hellendag Silly question, but if it's just an extra step to clear the block data on split, where would this extra step go? It's not obvious to me where to hook this in so you are sure you control the block split behavior without having to copy/paste some logic from other places.

hellendag:
@remko I'm thinking it would be something like this:

```javascript
const afterSplit = splitBlockInContentState(...);
const withoutTrailingBlockData = Modifier.setBlockDataForContentState(
  afterSplit,
  afterSplit.getSelectionAfter().
  Map()
);
```

src/model/modifier/DraftModifier.js
cgestes: done :)

src/model/immutable/ContentBlock.js
cgestes: done.

src/model/transaction/setBlockDataForContentState.js
cgestes: both scenario are valid. In my case lines have icons at the beginning. Icons that should not not be propagated.

It would be nice to be able to tweak that kind of behaviors easily.

cgestes:
I have this issue with the type of the new block also.

maybe splitting, retaining the data, and having an optional callback would do the trick here.

remko:
@hellendag My concern is where you need to put this code to be sure you handle all 'split block' cases. A quick grep in the source code tells me I should catch Ctrl-O (with a custom key command handler) and Return (with a handleReturn), and make it call my custom behavior. Besides being suboptimal that I had to look into Draft internals, nothing guarantees that Draft won't add another point later on that calls splitContentBlock, and then my editor would break on an update. Or am I missing something?

hellendag:
Great point.

It occurs to me that https://github.com/facebook/draft-js/blob/06891490069d3ccfb89caf53f33a7545d39fc3ec/src/component/handlers/edit/editOnKeyDown.js#L88 should perhaps go through the regular command flow for `split-block`, rather than executing the newline insertion right away. That would give you the opportunity to handle the command yourself.

I also should do a better job of documenting the catchable default commands so that you don't have to go hunting in the source.

Then there's the problem of whether underlying implementation changes. I think one thing we might do here is simply have two methods exposed on `Modifier` -- one that preserves the block data, and one that does not. (Beneath that, the transaction function could accept a flag to indicate whether to preserve or not. But the higher-level API would clearly denote the behavior of the method.) Then as a caller, you can be certain that the method you're calling will not suddenly change under the hood.

remko:
@cgestes Are you still working on this PR? Is there anything I can do to help finish it?

src/model/transaction/mergeBlockDataForContentState.js
```diff
@@ -0,0 +1,45 @@
...
+    .map(block => block.merge({data: block.getData().merge(blockData)}));
+
+  return contentState.merge({
+    blockMap: blockMap.merge(newBlocks),
+    selectionBefore: selectionState,
+    selectionAfter: selectionState,
+  });
+}
+
+module.exports = setBlockDataForContentState;
```
hellendag: There's a lot of overlap between these two transaction modules...looks like some copy-pasting here. :)

Can they be combined into a single module, perhaps with a fourth parameter indicating the intended behavior? `Modifier` can continue to have two methods exposed, but they can call into the same function.

Also, would you mind adding a unit test for this transaction function?

src/model/modifier/DraftModifier.js
```diff
@@ -26,6 +26,8 @@ var invariant = require('invariant');
 var removeEntitiesAtEdges = require('removeEntitiesAtEdges');
 var removeRangeFromContentState = require('removeRangeFromContentState');
 var setBlockTypeForContentState = require('setBlockTypeForContentState');
+var setBlockDataForContentState = require('setBlockDataForContentState');
+var mergeBlockDataForContentState = require('mergeBlockDataForContentState');
 var splitBlockInContentState = require('splitBlockInContentState');
 
 import type {BlockMap} from 'BlockMap';
```
hellendag: Instead of requiring `Map` above, it can be imported as a Flow type.

```
import type {Map} from 'immutable';
```

cgestes: Ok I will change that. I'am off this week.

src/model/transaction/mergeBlockDataForContentState.js
cgestes: Sure.

cgestes:
refactored all operations on block to use a common modifyBlockForContentState

src/model/modifier/DraftModifier.js
```diff
@@ -204,13 +204,41 @@ var DraftModifier = {
     selectionState: SelectionState,
     blockType: DraftBlockType
   ): ContentState {
-    return setBlockTypeForContentState(
+    const op = block => block.merge({type: blockType, depth: 0});
```
hellendag: Nice! This is an elegant way to reuse the transaction function.

Nits:

- Internally we've been trying to use parens even on single-argument fat-arrow functions. The quick return syntax is fine, though.
- I'd like `operation` better than `op`. :)

src/model/transaction/modifyBlockForContentState.js
```diff
@@ -14,15 +14,14 @@
 'use strict';
 
 var Immutable = require('immutable');
-
 import type ContentState from 'ContentState';
-import type {DraftBlockType} from 'DraftBlockType';
 import type SelectionState from 'SelectionState';
+import type ContentBlock from 'ContentBlock';
```
hellendag: Nit: We like to alphabetize requires and imports, so `ContentBlock` should just be the first in the `import` list.

hellendag:
A couple of minor comments -- I think this looks good. :) Thanks!

cgestes:
PR fixed based on your comments.

hellendag:
Thanks for your patience. :)

hellendag:
@facebook-github-bot import

joshdover:
Any update on this? Looks like things are generally merged the same day they get moved to Phab, this one seems to be take a bit longer. Just want to keep the community in on the discussion if there is one.

hellendag:
@joshdover Mostly just depends on me having the time to get it through Phabricator. I import, sync internally as needed, perform testing on FB interfaces, and land once I have another pair of eyes sanity check it. I'm a bit slow on this one since it's a more complex change to a core object. :)

joshdover:
@hellendag Got it, thanks for the response and everything this project does :)

cgestes:
Any news?

hellendag:
I've got it passing lint, flow, and unit tests internally. Manual testing looks fine.

I'll see if I can get it landed today or tomorrow.

I think that in a followup, it would be good to add test coverage for `ContentBlock`, the conversion utilities, and the renamed transaction function. (I really should have noted this earlier -- sorry about that.) Since I'll be merging this without any existing use cases on Facebook, it's a little risky to pull this in and roll it into production without tests.

So once it's merged, if you wouldn't mind following up, that would be great! :)

cgestes:
I'll add test as needed, but it might not happen before end of may.

mitermayer:
Will be great to have this merged :)

spicyj:
I don't feel comfortable merging this until Isaac is back (#355) so this will sit until then, sorry.

dphoursa:
Any news on when this might be merged?

spicyj:
@dphoursa Nothing's changed since my last comment.

3miliano:
I think there are two pieces that are missing in this PR. 

The first piece is that while there are two methods in `Modifier`, we need to add equivalent `EditorChangeType` so that `EditorState.push` will understand the change (and redo/undo work correctly).

The second piece is that we need a way from a custom block to access this data and as far as I understand there is no way to access the block so you can issue a `getData`.

I have a very limited understanding from DraftJS architecture so I might be talking non-sense.

delijah:
+1

ianstormtaylor:
Any news on this?

ianstormtaylor:
Using this branch via a fork and building from it... noticed that @3miliano is right about needing a extra `EditorChangeType` for `push` to work properly.

(Not sure about what you mean about "from a custom block" @3miliano, since if you have access to the `ContentBlock` you can always call `.getData()`.)

hellendag:
Picking this back up. Sorry for the long wait.

jpuri:
Hi Team,
When is this one going to get merged, we are waiting 😄

hellendag:
Just one remaining blocking question on this internally, regarding how to handle serialization/deserialization of nested data in the `Map`.

Converting from raw JS to state entails wrapping metadata in a `Map`, without making nested values immutable. The reverse, from state to JS, performs a `.toJS()` with no regard for whether the inner data is immutable. This means it isn't really symmetric and would not necessarily be a clean round trip.

Hashing it out. :)

joshdover:
Awesome, thanks for everyone's hard work on this 🎉

joshmarinacci:
I'm so excited. When will this be released?

mjomble:
I noticed there's still an `EditorChangeType` missing for this. And possibly also a `RichUtils` method.
I could submit a PR for that unless someone's already working on it.

mjomble:
Ah, the EditorChangeType actually is already there:

https://github.com/facebook/draft-js/blob/4cb881139e34b9bac84ed5847e010e3a495bd56f/src/model/immutable/EditorChangeType.js#L21

but it is not yet documented:

https://github.com/facebook/draft-js/blob/4cb881139e34b9bac84ed5847e010e3a495bd56f/docs/APIReference-EditorChangeType.md

So I could submit a PR for updating the documentation and adding RichUtils.setBlockData

feliperohdee:
looks good @mjomble

petry:
Hi Guys,

Do you guys intend to launch a release in the coming days? We love to use this feature on https://github.com/globocom/megadraft 😄

krishnagopinath:
Is there some good example to demonstrate `Modifier.setBlockData`? I'm trying to create an `atomicBlock` which has an image. Thanks! :)

jpuri:
@krishnagopinath: this might help you: https://github.com/jpuri/react-draft-wysiwyg/blob/master/js/src/components/Controls/TextAlign/index.js#L62

PR #229, opened 2016-03-24
Hotfix/block type inline style override

### LINKED ISSUE #228: Current inline styles are lost when block type is changed
I noticed this when integrating Draft.js into my own application. In my case, I was setting per-block alignment based on user input by changing the block type using `RichUtils.toggleBlockType`. When I do this, if there are any existing currently inline styles, they are lost.

I can reproduce the same behavior on the [live Draft.js demo](https://facebook.github.io/draft-js/). Steps to reproduce:

1. Click bold and or italic.
2. Click the UL or H1 or BlockQuote button (or any other button that changes the block type). 

Notice the bold/italic style has been reverted. This does _not_ occur if you type some text in between steps 1 and 2. I attempted to trace the exact issue in the code, but due to my relative unfamiliarity with the codebase, I was unable to locate the specific bug or determine a suitable workaround. If you have any advice regarding a workaround, that would be most helpful.

Thanks much for a spectacular library!

rnicholus:
This fixes #228. I also fixed a few existing lint violations and added a simple but useful test to verify this behavior change.

### DIFF
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -384,11 +384,14 @@ class EditorState {
       redoStack: Stack(),
       lastChangeType: changeType,
       selection: contentState.getSelectionAfter(),
-      forceSelection
+      forceSelection,
     };
 
-    // don't override inline style on block type change
-    if (changeType !== 'change-block-type') {
+    // don't override inline style on block type or depth changes
+    if (
+      changeType !== 'adjust-depth' &&
+      changeType !== 'change-block-type'
+    ) {
       editorStateChanges.inlineStyleOverride = null;
     }
 
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -377,16 +377,22 @@ class EditorState {
       );
     }
 
-    return EditorState.set(editorState, {
+    var editorStateChanges = {
       currentContent: newContent,
       directionMap,
       undoStack,
       redoStack: Stack(),
       lastChangeType: changeType,
       selection: contentState.getSelectionAfter(),
-      forceSelection,
-      inlineStyleOverride: null,
-    });
+      forceSelection
+    };
+
+    // don't override inline style on block type change
+    if (changeType !== 'change-block-type') {
+      editorStateChanges.inlineStyleOverride = null;
+    }
+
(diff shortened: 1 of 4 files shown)


src/model/immutable/EditorState.js
```diff
@@ -377,16 +377,22 @@ class EditorState {
...
       currentContent: newContent,
       directionMap,
       undoStack,
       redoStack: Stack(),
       lastChangeType: changeType,
       selection: contentState.getSelectionAfter(),
-      forceSelection,
-      inlineStyleOverride: null,
-    });
+      forceSelection
```
hellendag: Trailing comma

src/model/immutable/EditorState.js
```diff
@@ -377,16 +377,22 @@ class EditorState {
...
+    };
+
+    // don't override inline style on block type change
+    if (changeType !== 'change-block-type') {
```
hellendag: Could probably include `'adjust-depth'` here too.

hellendag:
Thanks!

src/model/immutable/EditorState.js
rnicholus: This was pre-existing code, but I can certainly remove this. Note that trailing commas are legal in ES3+.

hellendag: Oh, no, I'm saying that `forceSelection` can keep its comma (as `inlineStyleOverride` did previously).

rnicholus: Ah, sorry. I'll re-add that then.

src/model/immutable/EditorState.js
rnicholus: I can certainly account for this case as well. It sounds like "adjust-depth" is used to represent tabs. Is this correct?

rnicholus:
@hellendag My latest commit seems to address your review comments.

hellendag:
> I can certainly account for this case as well. It sounds like "adjust-depth" is used to represent tabs. Is this correct?

Correct. Thanks for adding it.

hellendag:
@facebook-github-bot import

hellendag:
I'm going to merge this with a handful of changes to appease Flow. :)

PR #235, opened 2016-03-27
Editor now gracefully handles ctrl +m

### LINKED ISSUE #218: Typing ctrl-m in the editor causes an invariant violation
To reproduce, focus on an empty editor and type (eg) x, ctrl-m, x.

Upon typing ctrl-m, the cursor returns to the beginning of the line, and a new block is created, but the second x is also within the first block, and the second block is empty. Both blocks have the same data-reactid. The first block seems to take up 2 lines (with the second line empty), and the second block is on the third line of the editor.

The exception is raised when the user focuses the cursor on the second block.
(OSX/Chrome)

brookslyrette:
When `ctrl + m` is pressed the block is split.

Closes #218

### DIFF
--- a/src/component/utils/getDefaultKeyBinding.js
+++ b/src/component/utils/getDefaultKeyBinding.js
@@ -86,7 +86,7 @@ function getDefaultKeyBinding(
     case 75: // K
       return (!isWindows && isCtrlKeyCommand(e)) ? 'secondary-cut' : null;
     case 77: // M
-      return hasCommandModifier(e) ? 'split-block' : null;
+      return isCtrlKeyCommand(e) ? 'split-block' : null;
     case 79: // O
       return isCtrlKeyCommand(e) ? 'split-block' : null;
     case 84: // T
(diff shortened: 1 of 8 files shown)


src/component/utils/getDefaultKeyBinding.js
```diff
@@ -85,6 +85,8 @@ function getDefaultKeyBinding(
       return hasCommandModifier(e) ? 'code' : null;
     case 75: // K
       return (!isWindows && isCtrlKeyCommand(e)) ? 'secondary-cut' : null;
+    case 77: // M
+      return hasCommandModifier(e) ? 'split-block' : null;
```
hellendag: This

hellendag: Shouldn't this be `isCtrlKeyCommand(e)`?

brookslyrette: Yep thanks for catching that. I'll resubmit.

Is there any test code for key commands? I was thinking of adding a test around the button press but didn't see any good examples to work off of.

hellendag:
> Is there any test code for key commands? I was thinking of adding a test around the button press but didn't see any good examples to work off of.

There isn't currently a JS test for the default key bindings, since it's generally easy enough to test changes manually.

hellendag:
@facebook-github-bot import

PR #244, opened 2016-03-29
Opt-Space no longer scrolls (Chrome OSX)

### LINKED ISSUE #19: Pressing Opt-Space scrolls the page down
If you type Opt-Space to insert a nbsp (on http://facebook.github.io/draft-js/, for example) in Chrome Mac, the page scrolls down a screenful but it probably shouldn't move. Missing preventDefault() somewhere probably.

brookslyrette:
`Option + Space` no longer causes scrolling. This was only happening in Chrome/OSX. This was happening in the keyDown handler not the keyPress handler. 

I'm not sure if I like adding the ' ' to the editor in editOnKeyDown. Maybe I should implement this as a command?

Fixes #19

### DIFF
--- a/src/component/handlers/edit/editOnKeyDown.js
+++ b/src/component/handlers/edit/editOnKeyDown.js
@@ -14,7 +14,7 @@
 
 var DraftModifier = require('DraftModifier');
 var EditorState = require('EditorState');
-var KeyBindingUtil = require('KeyBindingUtil')
+var KeyBindingUtil = require('KeyBindingUtil');
 var Keys = require('Keys');
 var SecondaryClipboard = require('SecondaryClipboard');
 
--- a/src/component/handlers/edit/editOnKeyDown.js
+++ b/src/component/handlers/edit/editOnKeyDown.js
@@ -17,6 +17,7 @@ var EditorState = require('EditorState');
 var KeyBindingUtil = require('KeyBindingUtil');
 var Keys = require('Keys');
 var SecondaryClipboard = require('SecondaryClipboard');
+var UserAgent = require('UserAgent');
 
 var keyCommandBackspaceToStartOfLine = require('keyCommandBackspaceToStartOfLine');
 var keyCommandBackspaceWord = require('keyCommandBackspaceWord');
@@ -32,6 +33,7 @@ var keyCommandUndo = require('keyCommandUndo');
 import type {DraftEditorCommand} from 'DraftEditorCommand';
 
 var { isOptionKeyCommand } = KeyBindingUtil;
+var isChrome = UserAgent.isBrowser('Chrome');
 
 /**
  * Map a `DraftEditorCommand` command value to a corresponding function.
@@ -107,7 +109,7 @@ function editOnKeyDown(e: SyntheticKeyboardEvent): void {
       return;
     case Keys.SPACE:
       // handling for OSX where option + space scrolls
-      if (isOptionKeyCommand(e)) {
+      if (isChrome && isOptionKeyCommand(e)) {
         e.preventDefault();
         // insert a space into the editor
         const contentState = DraftModifier.replaceText(
--- a/src/component/handlers/edit/editOnKeyDown.js
+++ b/src/component/handlers/edit/editOnKeyDown.js
@@ -106,6 +106,7 @@ function editOnKeyDown(e: SyntheticKeyboardEvent): void {
       this.props.onDownArrow && this.props.onDownArrow(e);
       return;
     case Keys.SPACE:
+      // handling for OSX where option + space scrolls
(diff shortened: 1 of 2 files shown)


src/component/handlers/edit/editOnKeyDown.js
```diff
@@ -101,6 +107,19 @@ function editOnKeyDown(e: SyntheticKeyboardEvent): void {
...
+      // handling for OSX where option + space scrolls
+      if (isChrome && isOptionKeyCommand(e)) {
+        e.preventDefault();
+        // insert a space into the editor
+        const contentState = DraftModifier.replaceText(
+            editorState.getCurrentContent(),
+            editorState.getSelection(),
+            ' '
+        );
+        this.update(EditorState.push(editorState, contentState, 'space'));
```
hellendag: The correct `EditorChangeType` here is `'insert-characters'`. (I need to document this, but one should always stick to the `EditorChangeType` values for `push` calls.)

src/component/handlers/edit/editOnKeyDown.js
```diff
@@ -101,6 +107,19 @@ function editOnKeyDown(e: SyntheticKeyboardEvent): void {
...
       return;
+    case Keys.SPACE:
+      // handling for OSX where option + space scrolls
+      if (isChrome && isOptionKeyCommand(e)) {
+        e.preventDefault();
+        // insert a space into the editor
+        const contentState = DraftModifier.replaceText(
+            editorState.getCurrentContent(),
+            editorState.getSelection(),
+            ' '
```
hellendag: Spacing nit: FB uses two spaces for indents.

hellendag:
@brookslyrette on the import I ran for your last PR, our bot pulled in changes that had already been merged, I assume because the commits were still present in this stack. Can those older changes be rebased out of this PR?

Regarding the PR itself, I'm wondering if this should be handled within `getDefaultKeyBinding`. It would be functionally the same as handling it in `editOnKeyDown`, but at least it would be with other key bindings that check OS/browser combinations.

brookslyrette:
Thanks @hellendag. I'll make the corrections and rebase the PR. 

I had attempted to prevent the scrolling in `getDefaultKeyBinding` but since I was editing the editorState I placed it in `editOnKeyDown`. From what I can tell, I can't add the space to the editor state from `getDefaultKeyBinding` without creating a command. 

I'm happy to do so if you think it's a cleaner approach. What name would be appropriate for the command?

hellendag:
Oh of course, that's a great point. There would have to be a command string for it. Let's stick with what you have.

Were you able to determine what's actually causing this? I.e. why would Draft behave differently from a plain contenteditable?

brookslyrette:
As far as I can tell it is because `document.activeElement` is not the contentEditable area. It's that the div is focused: 

screen shot 2016-03-29 at 12 50 20 pm

That seemed a bit too fundamental for me to be messing around with.

hellendag:
Hmm, though that div is the `contenteditable`.

> That seemed a bit too fundamental for me to be messing around with.

No worries, I'll see if I can track it down. :)

brookslyrette:
Just a hunch. This only seems to happen when `Option + Space` is entered into an empty editor. 

Not sure if it's a redraw or the placeholder causing it but I'm sure it has to do with the initial state.

hellendag:
It happens whenever we have to `preventDefault` on the `textInput` event, i.e. when we need to render the character manually rather than allowing the native behavior. This will happen at the start of a paragraph, after an immutable entity, and other cases.

It seems that if the `textInput` event is prevented when the `keydown` is allowed, the scroll behavior occurs.

Since we need to prevent the `textInput` event in the case described above, I think we don't really have any choice but to prevent the `keydown` and perform the manual insertion as you've done here.

hellendag:
Repro: https://jsfiddle.net/hellendag/amLxv1w8/

src/component/handlers/edit/editOnKeyDown.js
```diff
@@ -101,6 +107,19 @@ function editOnKeyDown(e: SyntheticKeyboardEvent): void {
...
       return;
+    case Keys.SPACE:
+      // handling for OSX where option + space scrolls
+      if (isChrome && isOptionKeyCommand(e)) {
+        e.preventDefault();
+        // insert a space into the editor
+        const contentState = DraftModifier.replaceText(
+          editorState.getCurrentContent(),
+          editorState.getSelection(),
+          ' '
```
hellendag: Is this the correct character to insert here, if the goal of <kbd>Opt</kbd>+<kbd>Space</kbd> is to insert a non-breaking space? Or should this be `'\u00a0'`?

hellendag:
Thanks!

hellendag:
@facebook-github-bot import

PR #247, opened 2016-03-29
Add EditorState.setInlineStyleOverride

bryanrsmith:
As discussed on Slack, this change adds a convenience method to `EditorState` for setting the inline style override without resorting to the undocumented `EditorState.set` method.

### DIFF
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -210,6 +210,13 @@ class EditorState {
     return this.getImmutable().get('inlineStyleOverride');
   }
 
+  static setInlineStyleOverride(
+    editorState: EditorState,
+    inlineStyleOverride: DraftInlineStyle
+  ): EditorState {
+    return EditorState.set(editorState, { inlineStyleOverride });
+  }
+
   /**
    * Get the appropriate inline style for the editor state. If an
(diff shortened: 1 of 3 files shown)


src/model/immutable/EditorState.js
```diff
@@ -210,6 +210,10 @@ class EditorState {
     return this.getImmutable().get('inlineStyleOverride');
   }
 
+  static setInlineStyleOverride(editorState: EditorState, inlineStyleOverride: DraftInlineStyle): EditorState {
+    return EditorState.set(editorState, { inlineStyleOverride });
+  }
```
hellendag: Styling nit: we use an 80-char limit on lines, so it would be good to have this as:

```javascript
static setInlineStyleOverride(
  editorState: EditorState,
  inlineStyleOverride: DraftInlineStyle
): EditorState {
  return EditorState.set(editorState, { inlineStyleOverride });
}
```

Sorry for nitpicking on this, it's just that if it's not in the PR, I'll have to do it myself anyway. :)

bryanrsmith: Oops, my bad. I saw that in the guidelines & then forgot about it. Have you considered adding it to the eslint config?

Updated now.

hellendag:
@facebook-github-bot import

PR #255, opened 2016-04-01
Fix cursor behavior for soft new lines

johanneslumpe:
Forcing a selection will cause the browser to behave properly and show the cursor for multiple new lines.


hellendag:
Awesome!

hellendag:
@facebook-github-bot import

PR #258, opened 2016-04-02
Rename media to atomic

tasti:
cc: https://github.com/facebook/draft-js/issues/124

The idea is that `media` isn't generic enough (and can lead to confusion) for what this `DraftBlockType` serves as.


src/model/modifier/AtomicUtils.js
```diff
@@ -27,8 +27,8 @@ const {
   Repeat,
 } = Immutable;
 
-const MediaUtils = {
-  insertMedia: function(
+const AtomicUtils = {
```
hellendag: Maybe `AtomicBlockUtils`, `insertAtomicBlock`, etc?

I think the term "media" stands on its own a bit better than "atomic". The block type is fine as `'atomic'`, I think, but elsewhere it may be worthwhile to provide a bit more clarity in naming.

tasti:
I renamed to `AtomicBlockUtils`, `insertAtomicBlock`, etc. and left some parts as Media for clarity, such as the media example and parts of the docs.

hellendag:
LGTM. @tasti, are you able to import this?

tasti:
@facebook-github-bot import

tasti:
@hellendag, nvm I don't have the special powers

hellendag:
This was merged via https://github.com/facebook/draft-js/commit/0096d9873e5f7e98145c0a835a52a42f54003306.

PR #259, opened 2016-04-02
Small docs fix

Mistereo:


docs/QuickStart-Rich-Styling.md
```diff
@@ -64,7 +64,7 @@ class MyEditor extends React.Component {
     return (
       <Editor
         editorState={this.state.editorState}
-        handleKeyCommand={this.handleKeyCommand}
+        handleKeyCommand={this.handleKeyCommand.bind(this)}
```
hellendag: I'd prefer if this were instead bound in the constructor, in line with current ES6 React patterns.

hellendag:
Thanks! Just one comment.

Mistereo:
@hellendag that was what I initially did, but then noticed that we use "binds" in render with `_onBoldClick` below and in "Key bindings" section.
Not sure that we should update this too since we skip some of implementation ( `//...`) including constructor in this examples and just omitting `bind` may be confusing.

hellendag:
Binding should definitely not be done in `render`, so that example should probably just have the glossed-over implementation details fleshed out.

hellendag:
I think if you rebase, one or two of these changes is already ironed out.

PR #264, opened 2016-04-04
Make 'Return' key go through command flow

remko:
As suggested by @hellendag in https://github.com/facebook/draft-js/pull/216#issuecomment-197979832

This allows customization of custom block  split behavior when hitting 'Return'.


hellendag:
@facebook-github-bot import

PR #266, opened 2016-04-04
Always delete empty block on delete

### LINKED ISSUE #190: Pressing 'delete' on an empty block before a styled block removes style
Reproducable with the rich example: 

- Hit 'return' to create an empty block
- Type something and make it an H1 block
- Go up to the empty block
- Hit 'delete'
- The H1 block is now an unstyled block

This doesn't seem intended behavior?

remko:
When hitting 'Delete' on an empty block, deletes the current block (instead of merging in the following block). This behavior is consistent with most other word editors, and also consistent with the case where the empty block is followed by a media block (in which case it already deleted the block).

This resolves #190

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -176,7 +176,32 @@ const RichTextEditorUtil = {
 
     const blockAfter = content.getBlockAfter(startKey);
 
-    if (!blockAfter || blockAfter.getType() !== 'atomic') {
+    if (!blockAfter) {
+      return null;
+    }
+
+    // Always delete empty blocks followed by non-atomic blocks
+    if (blockAfter.getType() !== "atomic") {
+      if (length === 0) {
+        var target = selection.merge({
+          focusKey: blockAfter.getKey(),
+          focusOffset: 0,
+        });
+
+        var withoutEmptyBlock = DraftModifier.removeRange(
+          content,
+          target,
+          'forward'
+        );
+
+        var preserveType = DraftModifier.setBlockType(
+          withoutEmptyBlock,
+          withoutEmptyBlock.getSelectionAfter(),
+          blockAfter.getType()
+        );
+
+        return EditorState.push(editorState, preserveType, 'remove-range');
+      }
       return null;
     }
 
--- a/src/model/modifier/__tests__/RichTextEditorUtil-test.js
+++ b/src/model/modifier/__tests__/RichTextEditorUtil-test.js
@@ -148,5 +148,30 @@ describe('RichTextEditorUtil', () => {
 
       expect(blockMapAfterDelete.size).toBe(4);
     });
+
+    it('deletes an empty block', () => {
+      // Make first block empty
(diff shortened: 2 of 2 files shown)


src/model/modifier/RichTextEditorUtil.js
```diff
@@ -195,13 +195,17 @@ var RichTextEditorUtil = {
...
         withoutEmptyBlock.getSelectionAfter(),
-        'media'
+        blockAfter.getType()
       );
 
-      return EditorState.push(editorState, preserveMedia, 'remove-range');
+      return EditorState.push(editorState, preserveType, 'remove-range');
+    }
+
+    if (blockAfter.getType() !== 'media') {
```
hellendag: This can be `atomic` now.

remko:
Updated the patch to take into account the media->atomic changes.

remko:
I have updated the PR after the changes in the deletion behavior around atomic blocks.

I made the deletion only apply to empty blocks followed by non-atomic blocks. It feels like it would be more consistent if it would also apply to empty blocks followed by atomic blocks (there were no tests failing if I removed this condition from the code), but it seems it was decided in #332 to treat atomic blocks differently in this case, so leaving it like this.

src/model/modifier/RichTextEditorUtil.js
```diff
@@ -176,7 +176,32 @@ const RichTextEditorUtil = {
 
     const blockAfter = content.getBlockAfter(startKey);
 
-    if (!blockAfter || blockAfter.getType() !== 'atomic') {
+    if (!blockAfter) {
+      return null;
+    }
+
+    // Always delete empty blocks followed by non-atomic blocks
+    if (blockAfter.getType() !== "atomic") {
```
hellendag: FFR: Single quotes. :)

hellendag:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1525808661058014/int_phab) to review.

qrohlf:
@hellendag it seems like this was never merged - is the only outstanding issue here the double quotes?

spicyj:
My impression is that this should be done at a lower level, like in DraftModifier.removeRange. In particular, here is where we choose to keep the preceding block's attributes:

https://github.com/facebook/draft-js/blob/bb6175cb5038acee8abec3c0a696e4eef5b30e01/src/model/transaction/removeRangeFromContentState.js#L53

It seems like the best way to fix this is to thread through a `removalDirection` boolean down to that function from all of its callers and choose whether to use `startBlock` or `endBlock` appropriately given that direction.

Note that RichTextEditorUtil does not currently handle this logic (DraftModifier and/or removeRangeFromContentState do), so I don't think it makes sense to try to work around that low-level "broken" behavior with a change in a higher-level module.

lacker:
It seems like this pull request has been abandoned, so I am going to close it. Please feel free to reopen it if you are still interested in working on it!

PR #268, opened 2016-04-04
Fix performance issue with generateRandomKey

ukd1:
I was hacking with @nikgraf on some draft-js stuff and noticed a performance bug. With the old implementation, we randomly get a key from a space and store that we've seen it. To avoid duplicates, we check that the found key has not been seen before. As the chance of any one key occurring is equal, the chance of collisions goes up the more keys we use, which means more time spent in a loop finding a new key. Also, as we have to store a list of keys we've seen, we also use more memory.

New implementation is additive. I've made it also 5 characters long and base 16 just incase anyone really needs that, but if not it can just be an integer.

jsperf: http://jsperf.com/draft-js-random-key2

Much love, 

Russ
p.s. also added tests


hellendag:
Sure, seems fine to me. Was this causing real perf/memory problems for you?

ukd1:
@hellendag not at the moment, but I suspect it will with super large or highly edited documents :+1:

hellendag:
> @hellendag not at the moment, but I suspect it will with super large or highly edited documents :+1:

I think super-large documents may have perf problems that go beyond key generation. :)

ukd1:
@hellendag fyi I signed the cla, but after making the PR.

hellendag:
@ukd1 No problem. :)

hellendag:
@facebook-github-bot import

ukd1:
@hellendag do you need me to bring this back up to date with master?

hellendag:
Yeah, please do, thanks. Not sure how the import will handle it -- I may need to import it again.

ukd1:
@hellendag done 👍

hellendag:
I imported this, but after some discussion with @spicyj, I don't think it's good idea to continue with it.

The main issue at hand is the conversion step, which preserves keys. (This may have been why I initially used random keys, and I'd just forgotten.)

With random keys, it should generally be safe to convert back and forth without worrying about key collision. With pure incrementing on a global counter, collision is essentially guaranteed.

We could always remove keys from the conversion process, but I actually prefer to have them in there as they potentially enable clients to serialize entire `EditorState` objects.

Thanks for the proposal, though!

ukd1:
@hellendag @spicyj if you have a collision issue, it is guaranteed either way, though might be less likely with the older code - but you still have worse performance / memory usage.

Can you point me to where the collision happens? I'd be happy to check, write tests and prove if it does / doesn't happen. As javascript is not concurrent, it shouldn't happen in the code I wrote? If you're merging lists of keys, then yes, but it's fixable.

hellendag:
Let's say that I restore serialized state from an earlier edit session, with a handful of blocks that already have keys from the incremental approach:

```javascript
const blocks = [
  {key: '000001', ...},
  {key: '000002', ...},
];

const contentState = convertFromRaw(blocks);
```

The earlier keys will be preserved in the new `ContentBlock` objects. (https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromRawToDraftState.js#L43)

After this point, as soon as I create a new `ContentBlock` and call `generateKey()` for the first time, I will hit a collision: `generateKey()` will provide `'000001'`, which we have already allocated during rehydration.

ukd1:
@hellendag interesting and totally true; instant collision. Are you interested in me fixing that; should be a pretty simple fix tbh.

hellendag:
What do you have in mind for a fix?

ukd1:
@hellendag; rekey - the easiest, plus safest. Basically just changing ``key = key || generateRandomKey();`` to ``key = generateKey();``. Disadvantage is loosing the original keying; though, except hand debugging I don't see the advantage; this can be fixed by storing ``oldKey`` or similar though.

hellendag:
I think the main reason to preserve original keying is in case the developer has decided to store undo/redo data.

ukd1:
@hellendag so the re-key would make it hard / impossible to apply? Surely they wouldn't be using conver for that?

PR #276, opened 2016-04-08
Prevent loss of entity during spellchecking

johanneslumpe:
Right now entities are lost when using the spellchecking feature of a browser. This change fixes that.

- [x] Fix `IMMUTABLE` entity undo bug
- [x] Completely remove `IMMUTABLE` entity and insert characters if spell change adds characters
- [x] Completely remove `IMMUTABLE` entity and insert characters if spell change changes characters
- [x] Completely remove `IMMUTABLE` entity and insert all characters if spell change removes characters
- [x] Completely remove `SEGMENTED` entity and insert characters if if spell change removes characters
- [x] Completely remove `SEGMENTED` entity and insert characters if character within entity is changed
- [x] Completely remove `SEGMENTED` entity and insert characters if character is added within entity

### DIFF
--- a/src/component/handlers/edit/editOnInput.js
+++ b/src/component/handlers/edit/editOnInput.js
@@ -85,8 +85,6 @@ function editOnInput(): void {
   var applyEntity = true;
   var entityType = entity && entity.getMutability();
   var changeType = 'spellcheck-change';
-  var newContent;
-  var mergeSelection = true;
 
   if (entityType) {
     var isImmutableEntity = entityType === 'IMMUTABLE';
@@ -100,33 +98,16 @@ function editOnInput(): void {
       'apply-entity' :
       changeType;
 
-    if (isImmutableEntity || isSegmentedEntity) {
-      var characterRemoved = domText.length < modelText.length;
-
-      if (characterRemoved) {
-        newContent = DraftModifier.removeRange(
-          content,
-          selection,
-          'backward',
-        );
-        mergeSelection = false;
-      }
-
-      if (!characterRemoved) {
-        applyEntity = false;
-      }
-    }
+    applyEntity = !(isImmutableEntity || isSegmentedEntity);
   }
 
-  newContent = newContent ?
-    newContent :
-    DraftModifier.replaceText(
-      content,
-      targetRange,
-      domText,
-      block.getInlineStyleAt(start),
-      applyEntity ? block.getEntityAt(start) : null,
-    );
+  var newContent = DraftModifier.replaceText(
+    content,
+    targetRange,
+    domText,
+    block.getInlineStyleAt(start),
+    applyEntity ? block.getEntityAt(start) : null,
+  );
 
   var anchorOffset, focusOffset, startOffset, endOffset;
 
@@ -156,12 +137,10 @@ function editOnInput(): void {
   // Segmented entities are completely or partially removed when their
   // text content changes. For this case we do not want any text to be selected
   // after the change, so we are not merging the selection.
-  var contentWithAdjustedDOMSelection = mergeSelection ?
-    newContent.merge({
-      selectionBefore: content.getSelectionAfter(),
-      selectionAfter: selection.merge({anchorOffset, focusOffset})
-    }) :
-    newContent;
+  var contentWithAdjustedDOMSelection = newContent.merge({
+    selectionBefore: content.getSelectionAfter(),
+    selectionAfter: selection.merge({anchorOffset, focusOffset})
+  });
 
   this.update(
     EditorState.push(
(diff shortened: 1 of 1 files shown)


johanneslumpe:
When removing a segmented entity due to a character being added and then trying to undo that change, it has the same effect as when removing an immutable entity - the text gets appended to the end of the block. My guess is that this has something to do with the selection not matching up properly...

johanneslumpe:
The failing test seems unrelated to this change. Its some flow type failing in `src/component/contents/DraftEditorContents.react.js`

src/component/handlers/edit/editOnInput.js
```diff
@@ -79,12 +80,53 @@ function editOnInput(): void {
...
+    // Immutable or segmented entities cannot properly be handled by the
+    // default browser undo, so we have to use a different change type to
+    // force using our internal undo method instead of falling through to the
+    // native browser undo.
+    changeType = (isImmutableEntity || isSegmentedEntity) ?
+      'apply-entity' :
+      changeType;
+
+    if (isImmutableEntity || isSegmentedEntity) {
+      var characterRemoved = domText.length < modelText.length;
```
hellendag: What if spellcheck makes the string longer?

hellendag:
Thanks for the thoroughness! One question regarding spellcheck.

src/component/handlers/edit/editOnInput.js
johanneslumpe: @hellendag That works as expected. The case for changing a character and when it is longer is identical. Here are 3 gifs showing each behavior for each entity type:

`IMMUTABLE`
spellcheck_immutable

`MUTABLE`
spellcheck_mutable

`SEGMENTED`
spellcheck_segmented

hellendag:
It occurs to me that it's a little weird that the removal case removes the entire string after the user has selected a spellcheck option. I'm thinking that behavior probably makes sense for intentional deletion, but isn't quite right for a spellcheck.

The more correct thing may be to replace the string and just strip the entity for `IMMUTABLE` and `SEGMENTED`.

What do you think?

(Sorry for the thrash!)

LucidityDesign:
This also fixes the same problem on iOS

johanneslumpe:
@hellendag Now we always remove `IMMUTABLE ` and `SEGMENTED` entities, no matter what kind of change the spellcheck change will trigger. Only `MUTABLE` entities are preserved.

src/component/handlers/edit/editOnInput.js
```diff
@@ -79,11 +80,33 @@ function editOnInput(): void {
...
+
+    // Immutable or segmented entities cannot properly be handled by the
+    // default browser undo, so we have to use a different change type to
+    // force using our internal undo method instead of falling through to the
+    // native browser undo.
+    changeType = (isImmutableEntity || isSegmentedEntity) ?
+      'apply-entity' :
+      changeType;
+
+    applyEntity = !(isImmutableEntity || isSegmentedEntity);
```
hellendag:
```js
const preserveEntity = entityType === 'MUTABLE';
const changeType = preserveEntity ? 'spellcheck-change' : 'apply-entity';
```

then...

src/component/handlers/edit/editOnInput.js
```diff
@@ -79,11 +80,33 @@ function editOnInput(): void {
...
+  }
+
   var newContent = DraftModifier.replaceText(
     content,
     targetRange,
     domText,
-    block.getInlineStyleAt(start)
+    block.getInlineStyleAt(start),
+    applyEntity ? block.getEntityAt(start) : null,
```
hellendag:
```js
preserveEntity ? block.getEntityAt(start) : null,
```

src/component/handlers/edit/editOnInput.js
```diff
@@ -111,16 +134,19 @@ function editOnInput(): void {
     focusOffset = endOffset + charDelta;
   }
 
+  // Segmented entities are completely or partially removed when their
+  // text content changes. For this case we do not want any text to be selected
+  // after the change, so we are not merging the selection.
   var contentWithAdjustedDOMSelection = newContent.merge({
     selectionBefore: content.getSelectionAfter(),
-    selectionAfter: selection.merge({anchorOffset, focusOffset}),
+    selectionAfter: selection.merge({anchorOffset, focusOffset})
```
hellendag: Keep trailing comma please. :)

hellendag:
Sweet!

src/component/handlers/edit/editOnInput.js
johanneslumpe: Was using `var` btw because the codebase does only use `var` - will change!

hellendag:
@facebook-github-bot import

Issue #222, opened 2016-03-21
Toolbars do not wrap properly on small devices


genyded:
Look at the OL toolbar button (in blue to the right in the attached screen shot) on an iPhone 6 device and you can seen it is outside the container:

![image](https://cloud.githubusercontent.com/assets/3081676/13923456/aea71c68-ef4e-11e5-980f-c13db29266bb.png)

This can be worked around by rotating the device, but it does not make for a good first impression.

hellendag:
I kind of want to remove H3-H6 from this example. I don't think it's really useful to have them there, since this is just a lightweight demo -- H1 and H2 are enough to get the point across.

PR #279, opened 2016-04-08
Use inline-block for proper button wrapping

johanneslumpe:
This fixes https://github.com/facebook/draft-js/issues/222


hellendag:
@facebook-github-bot shipit

PR #285, opened 2016-04-09
Now uses internalClipboard when present and in Safari. 

### LINKED ISSUE #147: Copy/Paste not preserving style on Safari
When you copy paste a piece of text that has certain parts styled (e.g. bold), copy/paste does not preserve the style in Safari. I tried it with both the example on the Draft.js page and the FB Notes editor.

I was assuming this was a Safari limitation, but I've seen other rich-text editors (e.g. Google Docs, Medium) support copy/paste on Safari.

brookslyrette:
After looking at how CKEditor I realized I was overthinking this. No extra clipboard library was need. The clipboard was already being stored. 

The only drawback to this solution is that formatting is only preserved when cut and pasted within one editor instance. 

Fixes #147

### DIFF
--- a/src/component/handlers/edit/editOnPaste.js
+++ b/src/component/handlers/edit/editOnPaste.js
@@ -190,10 +190,9 @@ function areTextBlocksAndClipboardEqual(
   textBlocks: Array<string>,
   blockMap: BlockMap
 ): boolean {
-  let i = 0;
   return (
     textBlocks.length === blockMap.size &&
-    blockMap.toSeq().every((block) => block.getText() === textBlocks[i++])
+    blockMap.valueSeq().every((block, i) => block.getText() === textBlocks[i])
   );
 }
 
--- a/src/component/handlers/edit/editOnPaste.js
+++ b/src/component/handlers/edit/editOnPaste.js
@@ -190,17 +190,11 @@ function areTextBlocksAndClipboardEqual(
   textBlocks: Array<string>,
   blockMap: BlockMap
 ): boolean {
-  if (textBlocks.length != blockMap.size) {
-    return false;
-  }
   let i = 0;
-  for (const block of blockMap) {
-    if (block[1].getText() !== textBlocks[i]) {
-      return false;
-    }
-    i++;
-  }
-  return true;
+  return (
+    textBlocks.length === blockMap.size &&
+    blockMap.toSeq().every((block) => block.getText() === textBlocks[i++])
+  );
 }
 
 module.exports = editOnPaste;
--- a/src/component/handlers/edit/editOnPaste.js
+++ b/src/component/handlers/edit/editOnPaste.js
@@ -127,8 +127,9 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
         return;
       }
     // Safari does not properly store text/html. Use the internalClipboard
-    // if present.   
-    } else if (isSafari && internalClipboard) {
+    // if present and equal to what is on the clipboard.
+    } else if (isSafari && internalClipboard &&
+        areTextBlocksAndClipboardEqual(textBlocks, internalClipboard)) {
       this.update(
         insertFragment(this.props.editorState, internalClipboard)
       );
@@ -183,6 +184,23 @@ function insertFragment(
     newContent,
     'insert-fragment'
   );
+}
+
+function areTextBlocksAndClipboardEqual(
+  textBlocks: Array<string>,
+  blockMap: BlockMap
+): boolean {
+  if (textBlocks.length != blockMap.size) {
+    return false;
+  }
+  let i = 0;
+  for (const block of blockMap) {
+    if (block[1].getText() !== textBlocks[i]) {
+      return false;
+    }
+    i++;
+  }
+  return true;
 }
 
 module.exports = editOnPaste;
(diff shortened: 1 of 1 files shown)


src/component/handlers/edit/editOnPaste.js
```diff
@@ -124,6 +126,13 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
         );
         return;
       }
+    // Safari does not properly store text/html. Use the internalClipboard
+    // if present.   
+    } else if (isSafari && internalClipboard) {
```
hellendag: How do we know that the `internalClipboard` is what the user intended to paste? The code above checks to make sure that the pasted content is a match for our internal clipboard, but this doesn't do the same check.

So with this change, I might paste something from an external source after having earlier copy-pasted internally, and my new paste would be discarded in favor of the older internal clipboard.

hellendag:
Thanks for digging into this!

brookslyrette:
Great point. I'll dig back into the code and see if I can come up with some more robust handling.

src/component/handlers/edit/editOnPaste.js
```diff
@@ -176,4 +186,21 @@ function insertFragment(
...
+    return false;
+  }
+  let i = 0;
+  for (const block of blockMap) {
+    if (block[1].getText() !== textBlocks[i]) {
+      return false;
+    }
+    i++;
+  }
+  return true;
```
hellendag: How about this?

```js
return (
  textBlocks.length === blockMap.size &&
  blockMap.toSeq().every((block, ii) => block.getText() === textBlocks[ii])
);
```

brookslyrette: Thanks! I'm pretty new to `immutable-js` so I missed that. I'll make the change and do some testing before submitting.

src/component/handlers/edit/editOnPaste.js
```diff
@@ -176,4 +186,15 @@ function insertFragment(
...
 }
 
+function areTextBlocksAndClipboardEqual(
+  textBlocks: Array<string>,
+  blockMap: BlockMap
+): boolean {
+  let i = 0;
+  return (
+    textBlocks.length === blockMap.size &&
+    blockMap.toSeq().every((block) => block.getText() === textBlocks[i++])
```
brookslyrette: `blockMap.toSeq().every` does not include the index of the iteration: https://facebook.github.io/immutable-js/docs/#/Iterable/every (easily confused with: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) 

Because of this I made a small tweak to @hellendag's proposed code.

hellendag: Ah, sorry -- I think you could do `blockMap.valueSeq().every(...)` to fix that. `toSeq` will keep the OrderedMap's keys, but I believe `valueSeq` should be zero-indexed.

hellendag:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/501720186680049/int_phab) to review.

src/component/handlers/edit/editOnPaste.js
```diff
@@ -124,6 +126,14 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
         );
         return;
       }
+    // Safari does not properly store text/html. Use the internalClipboard
+    // if present and equal to what is on the clipboard.
+    } else if (isSafari && internalClipboard &&
```
mitermayer: I am not a big fan of browser target fixes like this. Whats the root cause of this problem ? can we solve with feature detection ?  I think we should avoid browser specific hacks in the library.

mitermayer: "Safari does not properly store text/html."  - what you mean by properly ? Also in case that it is a Safari bug would it mean that in the future this snippet would become a code smell ? 

Sorry for being a bit picky in here, but with personal experiences working with multiple text editors in the past as soon as we introduce browser hacks/fixes branching inside the code base the code quality will degenerate very quickly

brookslyrette: HTML not being stored because of a bug: https://bugs.webkit.org/show_bug.cgi?id=19893 (you can see what is stored in #147)

I'm happy to move the code wherever, since it was Paste specific I placed it here.

spicyj:
Would it make more sense to check for "has com.apple.webarchive and does not have text/html" instead of hardcoding Safari? I guess it'd be even better if we could parse the webarchive data but that's probably undocumented or hard. (I just have an allergic reaction to UA tests, especially ones that are unbounded by version.)

brookslyrette:
I think that is a great idea. I'll make the change.

brookslyrette:
Anything I can do to help get this merged?

hellendag:
I'm going to include your latest changes and try to merge it today. :)

hellendag:
@brookslyrette Landing this now.

Two notes, which can be resolved separately this is merged:

- I think that it would be good for Safari to use the `editorKey` check if possible. I notice that when copy-pasting an atomic block with a custom decorator, the paste falls back to the raw text.
- The logic here can be cleaned up, since the outcome of both branches here is the same. (Call `this.update` and return.)

Thanks for your patience!

PR #287, opened 2016-04-09
Remove h3-h6 from the lightweight editor demo

jaaberg:
Removed the H3-H6 from the demo as it is just a lightweight demo, and H1-H2 should be enough to get the point across. As suggested by @hellendag in [#222 (comment)](https://github.com/facebook/draft-js/issues/222#issuecomment-199341556).


hellendag:
Thanks!

hellendag:
@facebook-github-bot shipit

PR #292, opened 2016-04-10
Update Advanced-Topics-Block-Components.md

dannav:
When researching how to create a custom editor block, there was a lot of mention of DraftEditorBlock but no information on how it was exposed... Thought this would help others out in the future.

### DIFF
--- a/docs/Advanced-Topics-Block-Components.md
+++ b/docs/Advanced-Topics-Block-Components.md
@@ -76,7 +76,7 @@ If your component contains text as provided by your `ContentState`, your custom
 component should compose a `EditorBlock` component. This will allow the
 Draft framework to properly maintain cursor behavior within your contents.
 
-By default it must have props defined. You can set all properties on the `EditorBlock` to those passed in by a blockRendererFn as follows:
+By default it must have props defined. You can set all properties on the `EditorBlock` to those passed in by a `blockRendererFn` as follows:
 
 ```js
 import React from 'react';


docs/Advanced-Topics-Block-Components.md
```diff
@@ -76,6 +76,28 @@ If your component contains text as provided by your `ContentState`, your custom
...
+    // class methods
+  
+    render() {		
+  		return (
+  			<EditorBlock 
+  				{...this.props} 
+  			/>
+  		)
+  	}
+	}
```
hellendag: FB standard is to use two spaces per indent. :)

docs/Advanced-Topics-Block-Components.md
```diff
@@ -76,6 +76,28 @@ If your component contains text as provided by your `ContentState`, your custom
...
+  class YourCustomBlockComponent extends React.CreateClass {
+    
+    // class methods
+  
+    render() {		
+  		return (
+  			<EditorBlock 
+  				{...this.props} 
+  			/>
+  		)
```
hellendag: Nit: please use semicolons.

Also, I think this whole return statement could just be a single line.

```javascript
  return <EditorBlock {...this.props} ... />;
```

docs/Advanced-Topics-Block-Components.md
```diff
@@ -76,6 +76,28 @@ If your component contains text as provided by your `ContentState`, your custom
...
 
+`DraftEditorBlock` is exposed as `EditorBlock` in the core library. By default it must have props
+defined. You can set all properties on the DraftEditorBlock to those passed in by your blockRendererFn
+as follows:
+
+```js
+  import React from 'react';
+  import { EditorBlock } from 'draft-js'; // EditorBlock is "DraftEditorBlock"
+  
+  class YourCustomBlockComponent extends React.CreateClass {
```
hellendag: `extends React.Component {`

docs/Advanced-Topics-Block-Components.md
```diff
@@ -76,6 +76,28 @@ If your component contains text as provided by your `ContentState`, your custom
 component should compose a `DraftEditorBlock` component. This will allow the
 Draft framework to properly maintain cursor behavior within your contents.
 
+`DraftEditorBlock` is exposed as `EditorBlock` in the core library. By default it must have props
```
hellendag: Throughout this content, this should refer to `EditorBlock` instead of `DraftEditorBlock`, since that's how it is exposed. Then you don't need to qualify the renaming with this sentence.

docs/Advanced-Topics-Block-Components.md
```diff
@@ -76,6 +76,28 @@ If your component contains text as provided by your `ContentState`, your custom
...
+defined. You can set all properties on the DraftEditorBlock to those passed in by your blockRendererFn
```
hellendag: `blockRendererFn`

hellendag:
Thanks, adding this to the docs seems like a useful change. :)

dannav:
Cool updated based on your notes

docs/Advanced-Topics-Block-Components.md
```diff
@@ -73,9 +73,24 @@ It is strongly recommended that you use `editable: false` if your custom
 component will not contain text.
 
 If your component contains text as provided by your `ContentState`, your custom
-component should compose a `DraftEditorBlock` component. This will allow the
+component should compose a `EditorBlock` component. This will allow the
 Draft framework to properly maintain cursor behavior within your contents.
 
+By default it must have props defined. You can set all properties on the `EditorBlock` to those passed in by a `blockRendererFn` as follows:
```
hellendag: I think the sentence `By default it must have props defined` is a little confusing. How about:

"You should apply `this.props` to your `EditorBlock` to pass along props supplied by `blockRendererFn`."

dannav:
I think ***should*** should be replaced with ***must***. Reason being, it definitely requires props to be defined or it will throw an error. Haven't looked into why just yet, but I can in a bit - so that more clarity is provided. 

*Should* implies that it's not required and can cause confusion for people that decide not to pass any props.

hellendag:
The most important prop here is `this.props.children`. Unless I'm forgetting something, I don't think *all* props need to be propagated, but that one must be used if the developer intends to render text.

If some other children are rendered instead of the ones provided via props, selection handling will lead to errors.

It may be worth clarifying this in the documentation accordingly, by explicitly setting `{this.props.children}` for the `EditorBlock` child.

dannav:
Ok thanks, I'll verify this and update accordingly

joshmarinacci:
Please integrate this. I searched all over to learn how to create a custom block with text, before finding this PR.

hellendag:
@dannav If you update the PR as discussed above, I'm happy to merge. :)

flarnie:
Just pinging this - I'm happy to merge this if the changes requested are made. I would do it, but it doesn't look like I have permission to update this PR.

flarnie:
Sorry to see this go unmerged, but since it's been inactive for a while I'm going to close it. Feel free to reopen with the edits requested in the initial review.

PR #301, opened 2016-04-13
Let cmd+backspace delete new line character.

### LINKED ISSUE #295: cmd+backspace should delete the \n if there is no text in the selected line
The default behavior of most multi-line inputs allows you to spam cmd+backspace to delete multiple lines of text. I'm always surprised when I can't in draft.

rickyvetter:
This adds extra logic to allow cmd+backspace to check for empty selections
at the beginning of a line a delete the new line character instead of doing
nothing. This behavior matches Chrome textarea.

Closes #295

### DIFF
--- a/src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
+++ b/src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
@@ -16,6 +16,7 @@ var EditorState = require('EditorState');
 
 var expandRangeToStartOfLine = require('expandRangeToStartOfLine');
 var getDraftEditorSelectionWithNodes = require('getDraftEditorSelectionWithNodes');
+var moveSelectionBackward = require('moveSelectionBackward');
 var removeTextWithStrategy = require('removeTextWithStrategy');
 
 function keyCommandBackspaceToStartOfLine(
@@ -24,6 +25,11 @@ function keyCommandBackspaceToStartOfLine(
   var afterRemoval = removeTextWithStrategy(
     editorState,
     strategyState => {
+      const selectionState = strategyState.getSelection();
+      if (selectionState.isCollapsed() &&
+          selectionState.getStartOffset() === 0) {
+        return moveSelectionBackward(strategyState, 1);
+      }
       var domSelection = global.getSelection();
       var range = domSelection.getRangeAt(0);
       range = expandRangeToStartOfLine(range);


src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
```diff
@@ -24,18 +25,24 @@ function keyCommandBackspaceToStartOfLine(
...
-      var selection = getDraftEditorSelectionWithNodes(
-        strategyState,
-        null,
-        range.endContainer,
-        range.endOffset,
-        range.startContainer,
-        range.startOffset
-      ).selectionState;
+
+      if (range.collapsed && range.startOffset === 0) {
```
hellendag: Instead of obtaining this from the DOM selection, you could look it up from `strategyState`.

```js
const selectionState = strategyState.getSelection();
if (selectionState.isCollapsed() && selectionState.getStartOffset() === 0) {
  return moveSelectionBackward(strategyState, 1);
}

// do what it was doing before
const domSelection = ...;
const range = ...;
// etc.
```

This way, you can avoid even looking into the DOM selection for the case in question.

Also, you should operate with `strategyState` (which is an `EditorState`) rather than the outer `editorState`. It should be the same object, but in theory it could be different.

rickyvetter: Gotcha. Using the DOM selection seemed like a really dirty way to do this. Makes a lot more sense.

src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
```diff
@@ -24,19 +25,24 @@ function keyCommandBackspaceToStartOfLine(
...
-        null,
-        range.endContainer,
-        range.endOffset,
-        range.startContainer,
-        range.startOffset
-      ).selectionState;
-      return selection;
+      const selectionState = strategyState.getSelection();
+      if (selectionState.isCollapsed() &&
+          selectionState.getStartOffset() === 0) {
```
hellendag: Can this fit on one line?

src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
```diff
@@ -24,19 +25,24 @@ function keyCommandBackspaceToStartOfLine(
...
+        return moveSelectionBackward(strategyState, 1);
+      } else {
```
hellendag: `else` not needed, since the block above returns.

hellendag:
sry for nitpicking

src/component/handlers/edit/commands/keyCommandBackspaceToStartOfLine.js
rickyvetter: It's 82 cols.

hellendag:
@facebook-github-bot import

PR #306, opened 2016-04-14
Fix Block.getLength() to be Unicode aware

prenaudin:
As ```Block.getLength()``` is widely used, it should be Unicode aware.

It fixes not just ```Block.getLength()``` but also ```Modifier.applyEntity()``` which uses internally ```Block.getLength()``` to apply an ```Entity``` at a given range.

Let me know if you need more test coverage on this issue.

### DIFF
--- a/docs/APIReference-ContentBlock.md
+++ b/docs/APIReference-ContentBlock.md
@@ -173,9 +173,8 @@ getLength(): number
 ```
 Returns the length of the plaintext for the `ContentBlock`.
 
-This value uses the standard JavaScript `length` property for the string, and
-is therefore not Unicode-aware -- surrogate pairs will be counted as two
-characters.
+This operation is Unicode-aware -- surrogate pairs will be counted as one
+character.
 
 ### getDepth()
 
--- a/src/model/immutable/ContentBlock.js
+++ b/src/model/immutable/ContentBlock.js
@@ -13,6 +13,7 @@
 
 'use strict';
 
+var UnicodeUtils = require('UnicodeUtils');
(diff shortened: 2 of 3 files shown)


hellendag:
I'm not convinced that this is correct. This might be an annoying aspect of JS string handling, but it is consistent. Additionally, considering surrogate pairs as two characters matches the expectations of text node length in the DOM.

Are you encountering bugs related to this?

hellendag:
To clarify the point on DOM text node length, this is important for setting selection state at appropriate points within text nodes, handling autocorrect/spellcheck, and so on. We could instead translate back and forth from Unicode-aware to non-Unicode-aware for these purposes, but it seems simpler to me to deal with this kind of translation only when necessary, e.g. handling character deletion.

prenaudin:
I get your point and I agree with you 👌
Also, as the documentation is explicit, we could use `strlen(block.getText())` if we need to be Unicode aware.

But I'm still encountering several issues with surrogate pairs when using `ContentState.createFromText()` which uses `block.length` internally. See https://github.com/facebook/draft-js/blob/master/src%2Fmodel%2Fimmutable%2FContentState.js#L143

Should a surrogate pair be one or two characters in the 'Character List'?

hellendag:
Surrogate pairs are represented with two characters in `characterList`, to reflect the JS/DOM expectations of the string.

Can you describe the issues you're having with `ContentState.createFromText()`?

prenaudin:
Thank you for the answer, I'll close this issue and open an other one once I have a clear script to reproduce the bug.

PR #313, opened 2016-04-15
Upgrade flow to v0.23.0

gabelevi:
This PR should fix the travis build.

A custom Promise definition is needed since Flow removed non-standard APIs from its builtin libs

### DIFF
--- a/src/flow/Promise.js
+++ b/src/flow/Promise.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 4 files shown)


zpao:
Promise definition will need to go into fbjs because the failure is actually from Deferred. Putting it locally here is fine short term but I want to fix this for our other projects. I have some more info #305 because I actually think this type failure is wrong (at least here)

src/flow/Promise.js
```diff
@@ -0,0 +1,40 @@
+/**
```
zpao: If we do this here, please move this file to `<root>/flow/`, this isn't compatible with the build process.

hellendag:
I think @zpao's PR's take care of this.

PR #315, opened 2016-04-15
Document EditorChangeType

hellendag:
Also remove some unused values and alphabetize the enum.


hellendag:
@facebook-github-bot shipit

PR #322, opened 2016-04-18
Upgrade flow to 0.23

zpao:
Simplified version of #313. We don't actually need the Promise definition here (anymore?), though I did ship it in fbjs@0.8.1.

I also added the flow version we expect to the flowconfig so we'll hard fail there if we ever mismatch.


hellendag:
@facebook-github-bot shipit

PR #328, opened 2016-04-20
Upgrade fbjs to 0.8.1 to force that dependency for flow

zpao:
Mostly this should trigger Travis CI to upgrade I think, but also it's technically right and I should have done this in my Flow change.


hellendag:
@facebook-github-bot shipit

PR #332, opened 2016-04-22
Fix deletion behavior around atomic blocks

### LINKED ISSUE #326: Draft breaks under weird circumstances
Hi! My team and I were using draft to build a beautiful rich text editor, so first of all, thank you for open sourcing this, is great!

The issue that we found was this:

1. Focus on draft
2. Add an `atomic` block
3. Press `up arrow` an position in the first editable block (non-atomic)
4. Press `del` (on windows) or `fn + del` (on mac I think)
5. Try to write something
6. Error `Uncaught Invariant Violation: Unknown DraftEntity key.`

And draft is left broken

I think the solution will be focus in the next editable block and if there's none, add a new one.
I would like to hear what you guys think about this.

Thanks in advance, best regards, Gabriel.

hellendag:
There is currently a selection state bug for deletion behavior around atomic blocks. Namely, removing an atomic block from a neighboring block will remove the atomic block but leave the cursor in an invalid state.

Make it so that atomic blocks will be removed during:

- Forward deletion from the last offset of a preceding block
- Backward deletion from the zero offset of a following block

@tasti and I discussed whether there should be some special behavior for forward deletion from an empty preceding block, i.e. removing the current block and moving the cursor upward, but this seemed magical.

We can try this out and adjust the behavior as needed.

Test Plan:

View media.html. Add text and an image block. Put cursor in block above image, forward delete. Verify that atomic block is removed. Repeat with cursor in block below image and backspace, verify same.

Fixes #326.

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -16,6 +16,7 @@
 var DraftEntity = require('DraftEntity');
 var DraftModifier = require('DraftModifier');
 var EditorState = require('EditorState');
+var SelectionState = require('SelectionState');
 
 var adjustBlockDepthForContentState = require('adjustBlockDepthForContentState');
 var nullthrows = require('nullthrows');
@@ -23,7 +24,6 @@ var nullthrows = require('nullthrows');
 import type ContentState from 'ContentState';
 import type {DraftBlockType} from 'DraftBlockType';
 import type {DraftEditorCommand} from 'DraftEditorCommand';
-import type SelectionState from 'SelectionState';
 import type URI from 'URI';
 
 var RichTextEditorUtil = {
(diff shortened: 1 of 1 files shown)


hellendag:
This should be reconciled with #266 (cc @remko) to determine the most consistent behavior.

hellendag:
@remko, I'm going to go ahead with this PR to fix the existing deletion issues and the selection issues that go along with them -- it's a bad error and needs to go away. :(

I'm happy to continue discussion to define the best long-term approach.

remko:
@hellendag Sure. I'll revisit #266 afterwards.

hellendag:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/969524493102461/int_phab) to review.

Issue #43, opened 2016-02-23
convertFromRaw returns Array and not ContentState


davisml:
From the documentation

> convertFromRaw(rawState: RawDraftContentState): ContentState
> Given a raw state, convert it to ContentState object. This is useful when restoring contents to use within a Draft editor.

From my testing, convertFromRaw actually returns an Array of ContentBlock objects. To convert this back to a ContentState you must do the following:

> ContentState.createFromBlockArray(convertFromRaw(rawData))

hellendag:
#30 resolved this. Thanks!

Issue #28, opened 2016-02-23
convertFromRaw returns an Array of ContentBlocks not a ContentState


dustin-H:
Hi there,

As documented [here](https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#content) the function `convertFromRaw` should return a `ContentState` object. 

However it returns an array of `ContentBlock` objects.

**This works:**
```js
var data = '{"entityMap":{},"blocks":[{"key":"4oah6","text":"Dieser Test","type":"unstyled","depth":0,"inlineStyleRanges":[{"offset":7,"length":4,"style":"BOLD"},{"offset":0,"length":11,"style":"CODE"},{"offset":0,"length":6,"style":"ITALIC"}],"entityRanges":[]}]}'

var json = JSON.parse(data)

var blockArray = convertFromRaw(json);

var contentState = ContentState.createFromBlockArray(blockArray);

var editorState = EditorState.createWithContent(contentState);
```

**However this throws an error:**
```js
var contentState = convertFromRaw(json);

var editorState = EditorState.createWithContent(contentState);
```

**Error:**
<img alt="screen shot 2016-02-23 at 00 56 54" src="https://cloud.githubusercontent.com/assets/6003519/13236988/6c2ea194-d9c8-11e5-9dc9-35c41d527f25.png">

Is this a mistake in the docs or a bug?

hellendag:
Great catch. It's a mistake in the docs.

dustin-H:
Ok, thanks... PR is in progress ;)

maksimsemenov:
Hi there,

There is still mistake in the docs: https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#content

convertFromRaw(rawState: RawDraftContentState): ContentState

dustin-H:
Where? Can't find it.

maksimsemenov:
Here is a screenshot of this page: https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#content

I marked the mistake with the red rectangle
![group](https://cloud.githubusercontent.com/assets/3608139/14906894/d31150ba-0d8e-11e6-8b9f-c057a8a1cd3c.png)

dustin-H:
What do you think should be there?

maksimsemenov:
From conversation above instead of **ContentState** there should be array of **ContentBlocks**

dustin-H:
[...]

PR #333, opened 2016-04-22
Update entity example to provide contentState instead of blockArray

kyledetella:
As mentioned (and subsequently fixed in the documentation) by issues #28, #30, and #43 `createFromRaw` returns an `Array` of `ContentBlock` objects. This updates the [entity example](https://github.com/facebook/draft-js/blob/master/examples/entity/entity.html#L115-L118) to reflect those changes.

### DIFF
--- a/examples/entity/entity.html
+++ b/examples/entity/entity.html
@@ -112,10 +112,10 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
             },
           ]);
 
-          const blocks = convertFromRaw(rawContent);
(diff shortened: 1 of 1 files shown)


hellendag:
`convertFromRaw` returns a `ContentState` as of https://github.com/facebook/draft-js/commit/844b85e6576b0575a716bbfe0381336737482b53.

This example was fixed in a followup by https://github.com/facebook/draft-js/commit/ee082faaa2e7bd57ef755e1674511aa9aa11fb26.

Master is currently correct -- the documentation is incorrect and needs to be updated. I'll fix that up.

PR #334, opened 2016-04-24
Add index to block callbacks

carlesba:
Could be pretty handy to get the block index on the callbacks that run on every block.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -56,8 +56,8 @@ const handlerMap = {
 };
 
 type DefaultProps = {
-  blockRendererFn?: (block: ContentBlock) => ?Object;
-  blockStyleFn?: (type: number) => string,
+  blockRendererFn?: (block: ContentBlock, index: number) => ?Object;
+  blockStyleFn?: (type: number, index: number) => string,
   keyBindingFn?: (e: SyntheticKeyboardEvent) => ?string,
   readOnly?: boolean,
   spellCheck?: boolean,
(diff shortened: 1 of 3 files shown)


guitardave24:
What's a good use case?

Not sure index is a good thing to pass in. Why do you need index if you already have block.getKey()?

However, I do think it'd be nice for the style fn to be given a block or block key.

carlesba:
Right now, if you need to know the block index you have to:
```
const blockKey = block.getKey()
const index = this.state.editorState.getCurrentContent().getBlockAsArray().findIndex(block => block.getKey() === blockKey)
```
What could be pretty expensive when you need this on one of this callbacks. This way is almost free to get and useful.

Yes, could be weird to add it on the style fn but I did it for consistency.

hellendag:
FWIW, you don't need to convert to an array. You can do:

```js
const index = this.state.editorState
  .getCurrentContent()
  .getBlockMap()
  .takeWhile((block) => block !== targetBlock)
  .size;
```

It may be useful to add this as a method on `ContentState`, and I would be open to that.

Have you found that this is expensive in practice?

I can't really think of any other ways in which the numeric index of the block is being used in core code (though I may have forgotten something) so I'm a little hesitant to start exposing it here.

carlesba:
Well, this is a loop trying to find the index so if I need to do that for each block on every insertion I bet it's going to be expensive.

That's way I'd like to get block's index in those callbacks so I could move some calculations in there and get the index for free.

carlesba:
I guess I'm facing the problem in the wrong way.

carlesba:
Since I discovered that we can use `EditorBlock` inside the block Components I'm trying to rebuild a Gutter in an easier and cheaper way but, to get that, I need to get the block index on the `blockRendererFn` callback. Otherwise I have to find the block index manually for every block which is pretty expensive.

Does this make sense for you?

spicyj:
I don't love the implication that we'd have to rerender all the blocks when inserting one at the top. So when Isaac is back (#355) I'll let him make the call here. :)

PR #335, opened 2016-04-24
Fix TeX example

markusjwetzel:


hellendag:
See my comment on #333. Is your checkout current?

PR #336, opened 2016-04-24
[Docs] Fix return types in Data Conversion doc

hellendag:
As noted in #333.


hellendag:
@facebook-github-bot shipit

Issue #339, opened 2016-04-25
[todo] Change `handle` functions to return enum instead of boolean


hellendag:
Currently, [`handle` functions](http://facebook.github.io/draft-js/docs/advanced-topics-key-bindings.html#customization) return a boolean to indicate to the Draft handler core whether the event has been handled or not. I wrote this code, and even I still forget whether `true` means that it was handled or deferred. 🙊 

To remedy this, I propose that `handle` functions instead return values of a simple binary enum, much the same way that `DraftRemovalDirection` clarifies the direction of a deletion. This should allow developers to know immediately which value to return, eliminating confusion about the outcome of the handler.

For instance, the enum could look like:

``` js
/* @flow */
export type DraftHandleValue = 'handled' | 'not-handled';
```

And a handler function could look like:

``` js
handleReturn(e: SyntheticKeyboardEvent): DraftHandleValue {
  // do stuff
  return 'handled';
}
```

This would be a breaking change, but definitely worthwhile.

This should be a pretty easy PR for anyone looking to get started with the codebase. It would require a new module to export the Flow type, plus changes at all the handler callsites to accommodate the non-boolean values.

Referenced at https://github.com/facebook/draft-js/issues/141#issuecomment-214408348.

grrowl:
Surely it should mirror native DOM's behaviour, which is `return false` -> don't perform additional actions (or, `preventDefault`)?

guitardave24:
Agree with Tom. I feel it's far too prominent in handler type code to
return undefined (not intentionally and not) to allow propagation. Having
to go through all handler code to return an explicit valid to represent not
handled seems fairly gratuitous.

guitardave24:
To that end it might be worth allowing === true as well and not make it a
breaking change. (On phone haven't looked at PR so maybe you're already
doing this.)

hellendag:
As I noted, I find boolean returns confusing, whether it mirrors native DOM behavior or not. Apart from legacy reasons, why is it a more correct convention to return `false` than `true` when no further actions should be performed? :) (It's also worth noting that as of React 14, return values in event handlers no longer have meaning [[#](https://facebook.github.io/react/docs/events.html)].)

As for returning `undefined`, the Flow type definitions for these handler functions currently insist upon boolean returns (https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorProps.js#L88), though of course I can't guarantee that Draft users are using Flow. In theory, though, everyone should return explicit values.

[...]

PR #340, opened 2016-04-26
Convert handler signature to return a `DraftHandleValue` value, instead of a boolean - #339

johanneslumpe:
Changes for https://github.com/facebook/draft-js/issues/339

@hellendag I'm not personally happy with all the literal `'handled'` strings now. The same counts for every usage of string literals, like the values of `DraftRemovalDirection`, `DraftBlockType` or any other type value. What are your thoughts on extracting them into constant files and using those constants instead?

Something along these lines:
```js

import { HANDLED } from 'DraftHandleValueConstants';

/*...*/
handleReturn(e) {
  return HANDLED;
}
/*...*/
```
Since we already have those values to create the flow types, is that maybe a point where we can hook in to generate those constants, so we wouldn't have to duplicate them in two files?

### DIFF
--- a/src/component/handlers/drag/DraftEditorDragHandler.js
+++ b/src/component/handlers/drag/DraftEditorDragHandler.js
@@ -88,7 +88,7 @@ var DraftEditorDragHandler = {
     const files = data.getFiles();
     if (files.length > 0) {
       if (this.props.handleDroppedFiles &&
-          this.props.handleDroppedFiles(dropSelection, files)) {
+          this.props.handleDroppedFiles(dropSelection, files) === 'handled') {
         return;
       }
 
@@ -107,7 +107,7 @@ var DraftEditorDragHandler = {
     const dragType = this._internalDrag ? 'internal' : 'external';
     if (
       this.props.handleDrop &&
-      this.props.handleDrop(dropSelection, data, dragType)
+      this.props.handleDrop(dropSelection, data, dragType) === 'handled'
     ) {
       return;
     }
--- a/src/component/handlers/drag/DraftEditorDragHandler.js
+++ b/src/component/handlers/drag/DraftEditorDragHandler.js
@@ -23,6 +23,7 @@ const getUpdatedSelectionState = require('getUpdatedSelectionState');
 const nullthrows = require('nullthrows');
 
 import type SelectionState from 'SelectionState';
+const isEventHandled = require('isEventHandled');
 
 /**
  * Get a SelectionState for the supplied mouse event.
@@ -87,8 +88,10 @@ var DraftEditorDragHandler = {
 
     const files = data.getFiles();
     if (files.length > 0) {
-      if (this.props.handleDroppedFiles &&
-          this.props.handleDroppedFiles(dropSelection, files) === 'handled') {
+      if (
+        this.props.handleDroppedFiles &&
+        isEventHandled(this.props.handleDroppedFiles(dropSelection, files))
+      ) {
         return;
       }
 
@@ -107,7 +110,7 @@ var DraftEditorDragHandler = {
     const dragType = this._internalDrag ? 'internal' : 'external';
     if (
       this.props.handleDrop &&
-      this.props.handleDrop(dropSelection, data, dragType) === 'handled'
+      isEventHandled(this.props.handleDrop(dropSelection, data, dragType))
     ) {
       return;
     }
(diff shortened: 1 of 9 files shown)


hellendag:
Yeah, it's a fair question.

I had initially used constant maps instead throughout this codebase, but changed it a while back to the current setup. The use of Flow string unions was suggested to me by the Flow team, and I've been happy enough with it, apart from bugs with unions of unions. (Though that's a separate issue.)

I think the confusion around possible `EditorChangeType` values -- i.e., developers just using whatever random string they want in `EditorState.push()` -- has indicated that the string unions don't offer as much direction as constant maps would. The same has taken place with `DraftBlockType`. Every time I see a gist with arbitrary strings in calls to those functions, I say "aw dammit" and think about how that could have been avoided.

If non-FB developers aren't using Flow in their projects, then their type errors won't get caught. Using constant maps would at least indicate where the valid values can be found.

I'll chat with Flow folks and see if they have different opinions on this now. I'd like a solution that can be both statically typechecked and easily understandable and followable by humans.

johanneslumpe:
@hellendag Any updates on whether we should use string constants or not?

hellendag:
@johanneslumpe Let's go with the Flow union to match the existing unions. I'll talk it out with the Flow folks.

src/component/utils/isEventHandled.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Utility method for determining whether or not the value returned
+ * from a handler indicates that it was handled.
+ */
+function isEventHandled(value: DraftHandleValue): boolean {
+  return value === 'handled';
+}
+
+module.exports = isEventHandled;
```
johanneslumpe: @hellendag I used commonjs here because somehow it couldn't find the module when using ES6 modules. Is there anything special going on which prevents ES6 modules from working for normal code and not just types?

src/component/handlers/drag/DraftEditorDragHandler.js
```diff
@@ -23,6 +23,7 @@ const getUpdatedSelectionState = require('getUpdatedSelectionState');
 const nullthrows = require('nullthrows');
 
 import type SelectionState from 'SelectionState';
+const isEventHandled = require('isEventHandled');
```
guitardave24: why not `import {isEventHandled} from 'isEventHandled';`?

I assumed that import was now preferred over require and using that now as we go.

johanneslumpe: @guitardave24 see my comment below. Never had issues with es6 modules before but here the build process just didn't find the module when using `import`.

hellendag: For required modules, we still stick to `require` internally at FB. We use `import` syntax when importing Flow types only.

src/component/utils/isEventHandled.js
hellendag: This is what we do internally, so this is totally fine.

src/component/utils/isEventHandled.js
```diff
@@ -0,0 +1,26 @@
...
+'use strict';
+
+import type {DraftHandleValue} from 'DraftHandleValue';
+
+/**
+ * Utility method for determining whether or not the value returned
+ * from a handler indicates that it was handled.
+ */
+function isEventHandled(value: DraftHandleValue): boolean {
+  return value === 'handled';
```
hellendag: For (temporary) backwards compatibility, this should probably also deal with `value === true`. Then we can cut the cord in a subsequent breaking change.

src/model/constants/DraftHandleValue.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule DraftHandleValue
+ * @flow
+ */
+
+/*eslint-disable no-bitwise*/
```
hellendag: Did lint complain about the bitwise below? I think we should avoid doing that for exported Flow types. cc @zpao

johanneslumpe: This was already in there, so I cannot comment on that at the moment.

src/component/utils/isEventHandled.js
johanneslumpe: Ok cool. Will try to update this soon, so we can get this merged.

johanneslumpe:
@hellendag just added the check for `true` for backwards compatibility. Is this one good to go?

hellendag:
Looks good to me! Thanks @johanneslumpe!

PR #345, opened 2016-04-27
Add support for h4, h5, h6 block tags

mattmontgomery:
This adds support for h4, h5, and h6 block tags in `convertFromHTMLToContentBlocks.js`.

I see that #205 would solve for this as well, so maybe this is unnecessary. Given this is a simple change, I thought I'd submit the PR regardless.


mitermayer:
This is not required, since we already a PR to deal with that https://github.com/facebook/draft-js/pull/205/files

hellendag:
This should now be supported via the changes in https://github.com/facebook/draft-js/commit/e2037984f32de4d3c267f6d43e0588fd124c0168.

Issue #342, opened 2016-04-27
`customStyleMap` is too limiting


guitardave24:
`customStyleMap` works by identifying inline styles and iteratively building a style map by concatenating the maps together in order. For example, a character with BOLD and ITALIC would apply `font-weight: bold` and `font-style: italic`, which works well for thise case.

However, sometimes, you want to apply a single style based on having two overlapping ranges. For example, assuming the ordered set `['BOLD', 'ITALIC']` I want:
```js
style: {
  fontFamily: 'MyFont-BoldItalic';
}
```

This is because I don't control the weights via CSS transforms alone, but a custom font.

I thought of a couple hacky solutions:
1. Define and apply a new custom style `BOLD_ITALIC` when I detect overlapping ranges, and use the `customStyleMap` but this just feels like sadness.
2. Use a decorator, but sadly, I would have to switch entirely to using a decorator for single styles because otherewise `customStyleMap``creates a new, styled span within the decorator's span, overriding styles. 

The better solution, I think, may be to allow for a `customStyleFn` which takes a character and returns the styleMap you want to apply. This means that you'd lose the ability to "stack" styles via `customStyleMap`. Maybe this is a bit too heavy for this use case, but I'm not seeing anything more elegant right now.

hellendag:
The `customStyleFn` approach is interesting. I don't think you'd have to lose the ability to stack styles, you'd just have to do the combinations yourself.

An added benefit of this function would be avoiding the need to create a new `customStyleMap` object every time a new style name is added. This would be especially helpful for cases like color pickers.

Thinking aloud...

```js
function customStyleFn(style: DraftInlineStyle): Object {
  const output = {};
  const isBold = style.has('BOLD'):
  const isItalic = style.has('ITALIC');

  if (isBold) {
    if (isItalic) {
      output.fontFamily = 'MyFont-BoldItalic';
    } else {
      output.fontWeight = 'bold';
    }
  } else if (isItalic) {
    output.fontStyle = 'italic';
  }

  // A possibility for color picker style names...
  const color = style.filter((value) => value.startsWith('#')).first();
  if (color) {
    output.color = color;
  };

  // etc.
  return output;
}
```

Is this something like what you have in mind?

The library could provide a default `customStyleFn` to process existing maps and ease the transition.

guitardave24:
Yes, that's exactly what I was leaning towards. 

[...]

PR #354, opened 2016-04-30
Adds customStyleMap for more control over inline style ranges #342

guitardave24:
cc @hellendag

Note, I also considered making `customStyleMap` and `customStyleFn` mutually exclusive, where if the latter were specified, then the former would materialize a function with the current behavior and only pass down `customStyleMap` to the leaf component -- but, then you'd lose the default behavior, which seems a little aggressive.

I think as long as it's clear that `customStyleFn` will execute after any `customStyleMap` additions then this is fine.

The downside here, is that just specifying a `customStyleFn` means that the default draft styles will always apply and you'd have to overwrite them to stop it (which is the current situation anyway).

### DIFF
--- a/src/component/contents/DraftEditorLeaf.react.js
+++ b/src/component/contents/DraftEditorLeaf.react.js
@@ -150,9 +150,7 @@ class DraftEditorLeaf extends React.Component {
 
     if (customStyleFn) {
       const newStyles = customStyleFn(styleSet);
-      if (newStyles !== undefined) {
-        styleObj = Object.assign(styleObj, newStyles);
-      }
+      styleObj = Object.assign(styleObj, newStyles);
     }
 
     return (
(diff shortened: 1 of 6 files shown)


src/component/contents/DraftEditorLeaf.react.js
```diff
@@ -145,6 +148,13 @@ class DraftEditorLeaf extends React.Component {
       return Object.assign(map, style, mergedStyles);
     }, {});
 
+    if (customStyleFn) {
+      const newStyles = customStyleFn(styleSet);
+      if (newStyles !== undefined) {
```
hellendag: Any reason to restrict this check to `undefined`?

Relatedly, could this perform the `Object.assign` with `null` or `undefined` as supplied by the style function, since a no-op should be fine?

hellendag:
I like this, and I think that in the long term, this approach will make sense as a replacement for the style map given the superior flexibility.

guitardave24:
@hellendag comments addressed. Thanks

guitardave24:
@hellendag anything needed further on this PR for merge?

hellendag:
LGTM. Let's do it.

PR #356, opened 2016-05-01
Convert block type if block is empty or pasting at beginning - Fixes #302

### LINKED ISSUE #302: Incorrect paste behavior for bulleted lists
I think there's an issue handling `<ul>` pastes.

Take a list that looks like this:
```
* one
* two
```

If you select all text with the mouse cursor, and then paste it. 

The result looks like:
```
one
*two
```

The HTML given to the paste processor above starts with: `<meta charset='utf-8'><ul ...`

However, if you select via cursor selection and make sure to select the *entire* block and focus on the end of the last block (which may need an extra left cursor movement) the result looks like this:
```
* one
* two
```
The HTML given to the paste processor in that scenario starts with: `<meta charset='utf-8'><li ...`

This is what I would have expected to occur in the first scenario. 

I looked into the paste processor and it seems to be ignoring `<ul>` elements and only looking for `<li>`. See https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L304

The ul is processed as an unstyled block. Perhaps just adding `ul` to getBlockTypeForTag is the fix?

Happy to submit a PR if this is agreed.

FWIW, I noticed this behavior in Facebook Notes too, but Medium does what I expect for sake of argument.

johanneslumpe:
Potential fix for https://github.com/facebook/draft-js/issues/302

### DIFF
--- a/src/model/transaction/insertFragmentIntoContentState.js
+++ b/src/model/transaction/insertFragmentIntoContentState.js
@@ -100,10 +100,11 @@ function insertFragmentIntoContentState(
       var modifiedHead = block.merge({
         text: headText + appendToHead.getText(),
         characterList: headCharacters.concat(appendToHead.getCharacterList()),
-        type: (!text.length || targetOffset === 0) ? appendToHead.getType() : block.getType(),
+        type: headText ? block.getType() : appendToHead.getType(),
       });
 
       newBlockArr.push(modifiedHead);
+
       // Insert fragment blocks after the head and before the tail.
       fragment.slice(1, fragmentSize - 1).forEach(
         fragmentBlock => {
--- a/src/model/transaction/insertFragmentIntoContentState.js
+++ b/src/model/transaction/insertFragmentIntoContentState.js
@@ -100,10 +100,10 @@ function insertFragmentIntoContentState(
       var modifiedHead = block.merge({
         text: headText + appendToHead.getText(),
         characterList: headCharacters.concat(appendToHead.getCharacterList()),
+        type: (!text.length || targetOffset === 0) ? appendToHead.getType() : block.getType(),
       });
 
       newBlockArr.push(modifiedHead);
-
       // Insert fragment blocks after the head and before the tail.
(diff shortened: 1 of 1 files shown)


mitermayer:
Would be great to get this merged

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -100,10 +100,10 @@ function insertFragmentIntoContentState(
       var modifiedHead = block.merge({
         text: headText + appendToHead.getText(),
         characterList: headCharacters.concat(appendToHead.getCharacterList()),
+        type: (!text.length || targetOffset === 0) ? appendToHead.getType() : block.getType(),
```
spicyj: Can we do

```
headText ? block.getType() : appendToHead.getType()
```

here instead?

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -100,10 +100,10 @@ function insertFragmentIntoContentState(
...
       });
 
       newBlockArr.push(modifiedHead);
-
```
spicyj: And add back this newline?

src/model/transaction/insertFragmentIntoContentState.js
johanneslumpe: @spicyj That will change the insertion behavior. If we only check for `headText`, then we will always keep the type of the existing block, as long as that block contains text. If we insert at offset 0 though, I think that we should use the type of the new block instead of the existing one.

spicyj: `headText` is `text.slice(0, targetOffset)`, not all of the original text.

johanneslumpe: Oh duh...

johanneslumpe:
@spicyj updated!

spicyj:
Thanks!

@facebook-github-bot import

PR #361, opened 2016-05-03
handle xml carriage return, and zero width spaces

### LINKED ISSUE #274: `&#13` (Carriage Return from xml entity) will break `convertFromHTML`
when i use jsdom to process html => dom, 
some weird html page will have this issue (which claim they are html doc but return with some XML entities).

perform these changes solved my issue with `convertFromHTML`:
```javascript
html = html.trim().replace(REGEX_CR, '').replace(REGEX_NBSP, SPACE);
```
to
```javascript
var XML_RETURN = new RegExp('(&#13;|&#13)', 'g');
...
html = html.trim().replace(REGEX_CR, '').replace(REGEX_NBSP, SPACE).replace(XML_RETURN, '');
```

for time reason, i have not dive into this too much (other cases like ZERO WIDTH SPACE may break this too?). Just create this issue as a reminder for people who encounter same problem.

further reading: 
https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L374
http://stackoverflow.com/questions/1459170/what-is-13

LordLambda:
Fixes #274

ignores carriage returns from xml entities, and also their zero-width-spaced
bretheren.

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -45,6 +45,8 @@ var MAX_DEPTH = 4;
 var REGEX_CR = new RegExp('\r', 'g');
 var REGEX_LF = new RegExp('\n', 'g');
 var REGEX_NBSP = new RegExp(NBSP, 'g');
+var REGEX_CARRIAGE = new RegExp('(&#13;|&#13)', 'g');
+var REGEX_ZWS = new RegExp('(&#8203;|&#8203)', 'g');
 
 // Block tag flow is different because LIs do not have
 // a deterministic style ;_;
@@ -422,7 +424,9 @@ function getChunkForHTML(
   html = html
     .trim()
     .replace(REGEX_CR, '')
-    .replace(REGEX_NBSP, SPACE);
+    .replace(REGEX_NBSP, SPACE)
+    .replace(REGEX_CARRIAGE, '')
+    .replace(REGEX_ZWS, '');
 
   const supportedBlockTags = getBlockMapSupportedTags(blockRenderMap);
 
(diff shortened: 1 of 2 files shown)


hellendag:
Seems fine to me. Mind adding a couple test cases?

LordLambda:
Sure. I initially went against it due to that file not having any tests. However, I'll add in a test for my change and some of the other functionality.

LordLambda:
I actually found a place where there were tests for handling text just in a different place then where I was looking. I've added in those tests.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -45,6 +45,8 @@ var MAX_DEPTH = 4;
 var REGEX_CR = new RegExp('\r', 'g');
 var REGEX_LF = new RegExp('\n', 'g');
 var REGEX_NBSP = new RegExp(NBSP, 'g');
+var REGEX_CARRIAGE = new RegExp('(&#13;|&#13)', 'g');
+var REGEX_ZWS = new RegExp('(&#8203;|&#8203)', 'g');
```
hellendag: Sufficient to do `new RegExp('&#13;?', 'g')` etc?

LordLambda: Ah yea. That'd probably be a lot better. I'll go ahead, and do that.

hellendag:
@facebook-github-bot shipit

facebook-github-bot-9:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/590686081091725/int_phab) to review.

PR #362, opened 2016-05-03
Rich example: Remove P

hellendag:
I think this adds confusion, and it breaks spacing. I'd be happy to have a different example for this, but I don't think this is the right way to exemplify the block render map.


hellendag:
@facebook-github-bot shipit

mitermayer:
I think we can create a new example for the block render map, with the same PR that adds the documentation for it.

PR #374, opened 2016-05-06
Adding more attributes to convertFromHTMLToContentBlocks entity whitelist

mitermayer:
Making sure that entities also support other common attributes for importing from html and pasting content. This is very important for compatibility with already existing applications

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -57,7 +57,7 @@ var inlineTags = {
   s: 'STRIKETHROUGH',
   strike: 'STRIKETHROUGH',
   strong: 'BOLD',
-  u: 'UNDERLINE'
+  u: 'UNDERLINE',
 };
 
 var anchorAttr = [
@@ -65,7 +65,7 @@ var anchorAttr = [
   'href',
   'rel',
   'target',
-  'title'
+  'title',
 ];
 
 var lastBlock;
@@ -377,15 +377,14 @@ function genFragment(
   var entityId: ?string = null;
 
   while (child) {
-    if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
-      // had to cast to Object due to https://github.com/facebook/flow/issues/1730
-      const anchor: Object = child;
-
-      let entityConfig = {};
+    if (child instanceof HTMLAnchorElement && child.href && hasValidLinkText(child)) {
+      const anchor: HTMLAnchorElement = child;
+      const entityConfig = {};
 
       anchorAttr.forEach((attr) => {
-        if (anchor[attr]) {
-          entityConfig[attr] = anchor[attr];
+        const anchorAttribute = anchor.getAttribute(attr);
+        if (anchorAttribute) {
+          entityConfig[attr] = anchorAttribute;
         }
       });
 
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -57,12 +57,7 @@ var inlineTags = {
   s: 'STRIKETHROUGH',
   strike: 'STRIKETHROUGH',
   strong: 'BOLD',
-  u: 'UNDERLINE',
-/**
- *
- *
- * @returns {6 +60,14 @@ var inlineTags = {}
- */
+  u: 'UNDERLINE'
 };
 
 var anchorAttr = [
@@ -383,11 +378,11 @@ function genFragment(
   while (child) {
     if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
       // had to cast to Object due to https://github.com/facebook/flow/issues/1730
-      const anchor:Object = child;
+      const anchor: Object = child;
 
-      var entityConfig = {};
+      let entityConfig = {};
 
-      anchorAttr.forEach(attr => {
+      anchorAttr.forEach((attr) => {
         if (anchor[attr]) {
           entityConfig[attr] = anchor[attr];
         }
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -388,7 +388,7 @@ function genFragment(
         }
       });
 
-      entityConfig.url = new URI(entityConfig.href).toString();
+      entityConfig.url = new URI(anchor.href).toString();
 
       entityId = DraftEntity.create('LINK', 'MUTABLE', entityConfig);
     } else {
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -58,7 +58,20 @@ var inlineTags = {
   strike: 'STRIKETHROUGH',
   strong: 'BOLD',
   u: 'UNDERLINE',
+/**
+ *
+ *
+ * @returns {6 +60,14 @@ var inlineTags = {}
+ */
 };
+
+var anchorAttr = [
+  'className',
+  'href',
+  'rel',
+  'target',
+  'title'
+];
 
 var lastBlock;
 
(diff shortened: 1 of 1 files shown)


src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -60,6 +60,14 @@ var inlineTags = {
   u: 'UNDERLINE',
 };
 
+var anchorAttr = [
```
mitermayer: We could even extract most of the logic in this file to a 'whitelist' place

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -58,8 +58,21 @@ var inlineTags = {
   strike: 'STRIKETHROUGH',
   strong: 'BOLD',
   u: 'UNDERLINE',
+/**
+ *
+ *
+ * @returns {6 +60,14 @@ var inlineTags = {}
+ */
```
hellendag: What's this?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +379,23 @@ function genFragment(
...
 
   var entityId: ?string = null;
-  var href: ?string = null;
 
   while (child) {
     if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
-      href = new URI(child.href).toString();
-      entityId = DraftEntity.create('LINK', 'MUTABLE', {url: href});
+      // had to cast to Object due to https://github.com/facebook/flow/issues/1730
+      const anchor:Object = child;
```
hellendag: Space after colon.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +379,23 @@ function genFragment(
...
+
+      var entityConfig = {};
+
+      anchorAttr.forEach(attr => {
```
hellendag: Wrap the arrow function arg in parens.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +379,23 @@ function genFragment(
...
-  var href: ?string = null;
 
   while (child) {
     if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
-      href = new URI(child.href).toString();
-      entityId = DraftEntity.create('LINK', 'MUTABLE', {url: href});
+      // had to cast to Object due to https://github.com/facebook/flow/issues/1730
+      const anchor:Object = child;
+
+      var entityConfig = {};
```
hellendag: const

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +379,23 @@ function genFragment(
...
+
+      anchorAttr.forEach(attr => {
+        if (anchor[attr]) {
+          entityConfig[attr] = anchor[attr];
+        }
+      });
+
+      entityConfig.url = new URI(entityConfig.href).toString();
```
hellendag: We probably need to be certain that there is an `href` value here -- the URI module might shout if this is an empty string or something.

src/model/encoding/convertFromHTMLToContentBlocks.js
mitermayer: sorrry no idea whats that, must been vim messing around things. will update that! sorry not sure how did i not see it

mitermayer:
@hellendag updated based on your comments, used 'let' instead of 'const' since we needed to do the URI conversion. If was not for that could have used `reduce()` but could I think the current way its probably more clear.  
I think this part of the codebase should be re-written soon to make entities and inline styles pasting support configurable instead of hardcoded similar to the blockRenderMap. But then we can do it all in a bigger chunk and develop the 'schema'

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -57,9 +57,17 @@ var inlineTags = {
   s: 'STRIKETHROUGH',
   strike: 'STRIKETHROUGH',
   strong: 'BOLD',
-  u: 'UNDERLINE',
+  u: 'UNDERLINE'
```
hellendag: revert

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -57,9 +57,17 @@ var inlineTags = {
...
 };
 
+var anchorAttr = [
+  'className',
+  'href',
+  'rel',
+  'target',
+  'title'
```
hellendag: trailing comma

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +374,23 @@ function genFragment(
   }
 
   var entityId: ?string = null;
-  var href: ?string = null;
 
   while (child) {
     if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
-      href = new URI(child.href).toString();
-      entityId = DraftEntity.create('LINK', 'MUTABLE', {url: href});
+      // had to cast to Object due to https://github.com/facebook/flow/issues/1730
```
hellendag: // Casting to Object due to ...

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -366,12 +374,23 @@ function genFragment(
...
+      const anchor: Object = child;
+
+      let entityConfig = {};
```
hellendag: But `entityConfig` isn't reassigned anywhere, right? You just add fields to the object?

mitermayer: Sorry, It was a personal preference where I tend to prefer to not use const when adding properties to referenced Objects, will update it to const

src/model/encoding/convertFromHTMLToContentBlocks.js
mitermayer: Without casting we get this error:

```
src/model/encoding/convertFromHTMLToContentBlocks.js:387
387:         if (anchor[attr]) {
                 ^^^^^^^^^^^^ access of computed property/element. Indexable signature not found in
387:         if (anchor[attr]) {
                 ^^^^^^ Node
```

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -367,12 +375,22 @@ function genFragment(
   }
 
   var entityId: ?string = null;
-  var href: ?string = null;
 
   while (child) {
-    if (nodeName === 'a' && child.href && hasValidLinkText(child)) {
-      href = new URI(child.href).toString();
-      entityId = DraftEntity.create('LINK', 'MUTABLE', {url: href});
+    if (nodeName === 'a' && child instanceof Element && child.href && hasValidLinkText(child)) {
```
mitermayer: Managed to remove the Object casting but still need to do an Element assignment in order to have access to `getAttribute()` without failling on flow

since if trying to access property anchor[attr] directly causes the error:

```
access of computed property/element. Indexable signature not found in
```

mitermayer:
@hellendag done :)

hellendag:
This looks fine to me. Trying another Travis run.

mitermayer:
The build is green now

src/model/paste/__tests__/DraftPasteProcessor-test.js
```diff
@@ -293,7 +293,7 @@ describe('DraftPasteProcessor', function() {
     expect(output[0].getText()).toBe('This is a link, yep.');
     var entityId = output[0].getCharacterList().get(12).getEntity();
     var entity = DraftEntity.get(entityId);
-    expect(entity.getData().url).toBe('http://www.facebook.com/');
+    expect(entity.getData().url).toBe('http://www.facebook.com');
```
hellendag: Revert?

ghigt:
I would be interested to also have the attributes `data-*`.

src/model/paste/__tests__/DraftPasteProcessor-test.js
mitermayer: The link was `'This is a <a href="http://www.facebook.com">link</a>, yep.`

not sure how did we get to the backslash on the end.

mitermayer:
@hellendag reverted the tests change and fixed the issue, accessing the anchor element href directly preserved the slash as opposed to accessing the entityConfig.

mitermayer:
@hellendag anything else required for this one ?

PR #376, opened 2016-05-07
[WIP] Moving entities into content state #185

johanneslumpe:
This is just a very quick, hacky PR to showcase the general direction of how I want to approach things for #185. I mainly want to gather some feedback so we can discuss whether we want to keep `DraftEntity` and just use that to perform transactions on the state or whether we want to go a different route.

I have updated most of the tests to pass. One is failing due to me not having implemented `DraftEntity.mergeData` and two other ones are failing because they rely on an implementation which requires a `contentState` argument and I didn't refactor anything there yet.

I'm totally down to refactor anything here or to go a different direction. Just wanted to get the ball rolling. But one way or another: this is going to be a breaking change ;)


amireh:
Status update possible? Is there anyway we can help? 

Thanks for taking the initiative!

johanneslumpe:
@amireh I am waiting for initial feedback from @hellendag regarding this.

src/model/encoding/convertFromRawToDraftState.js
```diff
@@ -62,7 +71,9 @@ function convertFromRawToDraftState(
     }
   );
 
-  return ContentState.createFromBlockArray(contentBlocks);
+  const contentState = ContentState.createFromBlockArray(contentBlocks);
```
hellendag: Might make sense for `createFromBlockArray` to accept a second parameter for the entity map, rather than requiring the `set` below.

src/model/entity/DraftEntity.js
```diff
@@ -37,7 +35,8 @@ var instanceKey = 0;
  * generated via DraftEntity.create() and used to obtain entity metadata
  * via DraftEntity.get().
  */
-var DraftEntity = {
+const DraftEntity = {
```
hellendag: I think that if entities are being moved to `ContentState`, we can eliminate this module altogether -- all creating/updating behavior can be methods on `ContentState`.

src/model/entity/DraftEntity.js
```diff
@@ -46,62 +45,54 @@ var DraftEntity = {
...
   },
 
-  /**
-   * Retrieve the entity corresponding to the supplied key string.
-   */
-  get: function(key: string): DraftEntityInstance {
-    var instance = instances.get(key);
-    invariant(!!instance, 'Unknown DraftEntity key.');
-    return instance;
+  get: function(contentState: ContentState, key: String): DraftEntityInstance {
```
hellendag: Lowercase `string`

hellendag:
Thanks for your patience. :)

I think this is on the right track. A couple of comments here. Mostly I'm thinking we can just move everything directly into `ContentState` and throw out the `DraftEntity` module altogether.

johanneslumpe:
@hellendag I was on vacation last week. I hope I'll be able to take a look at this soon! :)

bkniffler:
Hey @johanneslumpe and @hellendag, this is awesome and in my opinion one of the more important things that need to be implemented atm. Not having entity state changes refresh the content state is a big pain point that gave me loads of headaches while developing plugins for draft-js-plugins, at the point were I change both component state and entity state to force atomic editor components to rerender, since forcing a rerender by setting selection doesn't work reliably.

Is there anything you need help with? Has there been any further discussions about this PR internally @hellendag?

johanneslumpe:
@bkniffler Unfortunately I still didn't have time to revisit this - got some personal things to deal with right now. I will get back on this eventually though!

johanneslumpe:
@hellendag the only thing left to fix is the `DraftPasteProcessor` or rather `convertFromHTMLtoContentBlocks`, which creates `LINK` entities. In order to do that I think we either have to give it access to the content state. But then again we won't be generating chunks/fragments but rather a new/updated `contentState` whenever we parse HTML. There is only a single call site to `convertFromHTMLToContentBlocks` though inside the `DraftPasteProcessor` and that again is only called in the `editOnPaste` handler:

```js
if (html) {
      var htmlFragment = DraftPasteProcessor.processHTML(
        html,
        this.props.blockRenderMap
      );
      if (htmlFragment) {
        var htmlMap = BlockMapBuilder.createFromArray(htmlFragment);
        this.update(insertFragment(this.props.editorState, htmlMap));
        return;
      }
    }
```
So instead of having a fragment here, we could just return a new `contentState` and then call `this.update` with the new state. What are your thoughts on this? Do you see any issues with that?

johanneslumpe:
@hellendag I went ahead and committed 012ea41 - it's one way to do it. we're generating additional objects now when parsing html now, but I'm not sure that that's a huge issue. Let's discuss what you think about it :)

src/model/transaction/updateEntityDataInContentState.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule updateEntityDataInContentState
```
hellendag: Module name should match exported function name. (Lint will complain about this internally.)

src/model/transaction/createEntityInContentState.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+'use strict';
+
+const addEntityToContentState = require('addEntityToContentState');
+const DraftEntityInstance = require('DraftEntityInstance');
+
+import type {DraftEntityMutability} from 'DraftEntityMutability';
+import type {DraftEntityType} from 'DraftEntityType';
+import type ContentState from 'ContentState';
```
hellendag: Module ordering pattern used internally (sorry for nitpicking):

```javascript
// alphabetized classes
const SomeClass = require('SomeClass');

// alphabetized functions
const someFunction = require('someFunction');

// alphabetized Flow imports
import type SomeFlowType from 'SomeFlowType';
```

src/model/transaction/addEntityToContentState.js
```diff
@@ -0,0 +1,30 @@
...
+const generateRandomKey = require('generateRandomKey');
+
+import type ContentState from 'ContentState';
+import type DraftEntityInstance from 'DraftEntityInstance';
+
+function addEntityToContentState(
+  contentState: ContentState,
+  instance: DraftEntityInstance
+): ContentState {
+  const newKey = generateRandomKey();
```
hellendag: I think it would be okay to keep using a regular incrementing counter for these IDs. The random key generator has mostly been useful for being able to spot a unique key within the DOM, etc., while here it's fine for it to just be a number.

src/model/immutable/EntityMap.js
```diff
@@ -0,0 +1,18 @@
...
+ * @providesModule EntityMap
+ * @flow
+ */
+
+'use strict';
+
+import type DraftEntityInstance from 'DraftEntityInstance';
+import type {OrderedMap} from 'immutable';
+
+export type EntityMap = OrderedMap<string, DraftEntityInstance>;
```
hellendag: Any reason this needs to be an `OrderedMap`? I would think that a regular `Map` is sufficient.

src/model/encoding/convertFromDraftStateToRaw.js
```diff
@@ -58,8 +57,9 @@ function convertFromDraftStateToRaw(
   // DraftEntity keys.
   var entityKeys = Object.keys(entityStorageMap);
   var flippedStorageMap = {};
+  const entityMap = contentState.getEntityMap();
   entityKeys.forEach((key, jj) => {
-    var entity = DraftEntity.get(DraftStringKey.unstringify(key));
+    var entity = entityMap.get(DraftStringKey.unstringify(key));
```
hellendag:
```javascript
const entity = contentState.getEntity(...);
```

src/model/decorators/DraftDecoratorType.js
```diff
@@ -25,7 +26,7 @@ export type DraftDecoratorType = {
   /**
    * Given a `ContentBlock`, return an immutable List of decorator keys.
    */
-  getDecorations(block: ContentBlock): List<?string>,
+  getDecorations(contentState: ContentState, block: ContentBlock): List<?string>,
```
hellendag: I'm thinking it may make sense to change the second argument to be a key string. That way we would never run the risk of passing in a `ContentBlock` that is not part of the `ContentState`. It's a breaking API change either way, so we might as well just totally break it. :)

src/component/handlers/edit/editOnPaste.js
```diff
@@ -146,9 +146,21 @@ function editOnPaste(e: SyntheticClipboardEvent): void {
         this.props.blockRenderMap
       );
       if (htmlFragment) {
-        var htmlMap = BlockMapBuilder.createFromArray(htmlFragment);
-        this.update(insertFragment(this.props.editorState, htmlMap));
-        return;
+        const { contentBlocks, contentState } = htmlFragment;
```
hellendag: Just to be clear, the idea here is that the returned `htmlFragment` has a `ContentState` with the new entities already added to it but not necessarily the relevant ranges of text -- those are included in the `contentBlocks` array?

If so, I'm wondering if `htmlFragment` could instead be something like

```javascript
{
  contentBlocks: Array<ContentBlock>,
  entityMap: EntityMap,
}
```

With `insert-fragment` accepting both a block map and an entity map for insertion.

What do you think?

src/component/handlers/edit/editOnInput.js
```diff
@@ -81,7 +80,7 @@ function editOnInput(): void {
   });
 
   const entityKey = block.getEntityAt(start);
-  const entity = entityKey && Entity.get(entityKey);
+  const entity = entityKey && content.getEntityMap().get(entityKey);
```
hellendag: `content.getEntity(entityKey)`

src/component/contents/__tests__/DraftEditorBlock.react-test.js
```diff
@@ -249,7 +250,7 @@ describe('DraftEditorBlock.react', () => {
       );
       var decorator = new Decorator();
 
-      var newTree = BlockTree.generate(helloBlock, decorator);
+      var newTree = BlockTree.generate(ContentState.createFromText(helloBlock.getText()), helloBlock, decorator);
```
hellendag: Internally we have an 80-col limit, so even if lint doesn't complain here we'll need to end up cleaning it up later.

src/Draft.js
```diff
@@ -43,7 +42,6 @@ const DraftPublic = {
   EditorState,
 
   CompositeDecorator: CompositeDraftDecorator,
-  Entity: DraftEntity,
   EntityInstance: DraftEntityInstance,
```
hellendag: Does this still need to be exposed?

src/Draft.js
```diff
@@ -22,7 +22,6 @@ const DefaultDraftInlineStyle = require('DefaultDraftInlineStyle');
 const DraftEditor = require('DraftEditor.react');
 const DraftEditorBlock = require('DraftEditorBlock.react');
 const DraftModifier = require('DraftModifier');
-const DraftEntity = require('DraftEntity');
```
hellendag: 🎉

examples/link/link.html
```diff
@@ -173,21 +176,22 @@
...
-      function findLinkEntities(contentBlock, callback) {
+      function findLinkEntities(contentState, contentBlock, callback) {
+        const entityMap = contentState.getEntityMap();
         contentBlock.findEntityRanges(
           (character) => {
             const entityKey = character.getEntity();
             return (
               entityKey !== null &&
-              Entity.get(entityKey).getType() === 'LINK'
+              entityMap.get(entityKey).getType() === 'LINK'
```
hellendag: Same comment as above.

examples/entity/entity.html
```diff
@@ -174,7 +175,7 @@
 
       const TokenSpan = (props) => {
         const style = getDecoratedStyle(
-          Entity.get(props.entityKey).getMutability()
+          props.contentState.getEntityMap().get(props.entityKey).getMutability()
```
hellendag:
```javascript
props.contentState.getEntity(props.entityKey).getMutability()
```

examples/entity/entity.html
```diff
@@ -149,14 +149,15 @@
...
+        return function(contentState, contentBlock, callback) {
+          const entityMap = contentState.getEntityMap();
           contentBlock.findEntityRanges(
             (character) => {
               const entityKey = character.getEntity();
               if (entityKey === null) {
                 return false;
               }
-              return Entity.get(entityKey).getMutability() === mutability;
+              return entityMap.get(entityKey).getMutability() === mutability;
```
hellendag: Could skip pulling out `entityMap`:

```javascript
return contentState.getEntity(entityKey).getMutability() === mutability;
```

hellendag:
I like this a whole lot and I think it's really close.

src/model/transaction/updateEntityDataInContentState.js
johanneslumpe: Yeah this is a complete oversight on my part...

src/Draft.js
johanneslumpe: I actually don't think so. I also noticed that I have a method `contentState.addEntity`, which takes a `DraftEntityInstance` but is actually not used... I'm not quite sure why I still have it but it might be a relict from porting over the methods of `DraftEntity` 1:1. Do you think we still need it for something?

src/model/transaction/addEntityToContentState.js
johanneslumpe: At the moment that doesn't work with the way we re-create a `ContentState` instance from raw state. There an entity map is created manually which is then passed to `ContentState.createFromBlockArray`. Since we have no `contentState` yet we cannot utilize `contentState.addEntity`. How about we do not pass the entity map to `ContentState.createFromBlockArray`, but create the content state and then reduce the raw entity map by calling `contentState.addEntity`? Then we could definitely use a numerical counter.

src/model/immutable/EntityMap.js
johanneslumpe: I chose an `OrderedMap` because of the fact that we need to know the last inserted key. I was under the impression that only an `OrderedMap` would be safe for that, since the order of iteration for `Map` is supposed to not be defined. If we can be sure that the last inserted key is always the actually last inserted key, then sure, we can use `Map` here.

src/component/handlers/edit/editOnPaste.js
johanneslumpe: You understood correctly. We could definitely just return an `EntityMap` instance and update `insertFragment` to support that. So we'd do `insertFragment(this.props.editorState, htmlMap, entityMap)`. Sounds good to me!

src/model/decorators/DraftDecoratorType.js
johanneslumpe: This would require us to also change the call sites for `BlockTree.generate` as those pass in the block which is forwarded to `getDecorations`. That happens while mapping over a content state's blocks. I agree that it would be nice to make sure that the lookup takes place within `getDecorations` using a block key and not a block instance. Having said that, since `getDecorations` is not a method which a consumer would call directly, is it really worth changing?

src/Draft.js
johanneslumpe: With the new changes I've made I think we can remove the export and the `addEntity` method. (If a consumer is not ever required to add an entity manually like that, which I think they shouldn't be)

bkniffler:
@johanneslumpe Whats the current state? Is there anything missing or is it ready for merge? I tried your branch on a local project and didn't encounter any issues.

johanneslumpe:
@bkniffler I asked some questions to @hellendag and I've been waiting on answers since. This actually works well as of now. There were potential changes to `src/model/decorators/DraftDecoratorType.js` but I haven't gone ahead with those because I wanted to discuss them first.

But this is working!

hellendag:
We're going to get this merged today. Taking another pass now.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -310,17 +314,18 @@ function genFragment(
   depth: number,
   blockRenderMap: DraftBlockRenderMap,
   inEntity?: string
-): Chunk {
+): {chunk: Chunk, entityMap: EntityMap} {
```
hellendag: This can be defined as a Flow type so that you don't have to repeat it.

src/model/encoding/convertFromRawToDraftState.js
```diff
@@ -32,13 +34,18 @@ function convertFromRawToDraftState(
...
       var {type, mutability, data} = encodedEntity;
-      var newKey = DraftEntity.create(type, mutability, data || {});
+      const instance = new DraftEntityInstance({type, mutability, data: data || {}});
+      const tempEntityMap = addEntityToEntityMap(updatedEntityMap, instance);
+      const newKey = tempEntityMap.keyseq().last();
       fromStorageToLocal[storageKey] = newKey;
-    }
+      return tempEntityMap;
+    },
+    OrderedMap()
```
hellendag: This could be defined once in the module.

src/model/immutable/ContentState.js
```diff
@@ -116,13 +135,47 @@ class ContentState extends ContentStateRecord {
...
+    data?: Object
+  ): ContentState {
+    return createEntityInContentState(this, type, mutability, data);
+  }
+
+  mergeEntityData(
+    key: string,
+    toMerge: {[key: string]: any}
+  ): ContentState {
+    return updateEntityDataInContentState(this, key, toMerge, true);
```
hellendag: As you know, I don't love boolean arguments and return values. :) But I want to get this unblocked, so we can just revisit if needed.

src/model/immutable/EntityMap.js
hellendag: Well, I guess the issue is still that there is a distinction between the order of insertion and the order of iteration. If these are the same for `OrderedMap`, then that sounds fair enough to me.

hellendag:
Okay, let's go ahead and do this, then refine as needed.

Thanks again for your patience! :)

PR #387, opened 2016-05-12
Adding missing documentation to BlockRenderMap and updating documentation for data conversion

mitermayer:
- Including documentation for BlockRenderMap
- Updating the data conversion

### DIFF
--- a/docs/APIReference-Data-Conversion.md
+++ b/docs/APIReference-Data-Conversion.md
@@ -49,7 +49,7 @@ other usage within an application.
 ### convertFromHTML
 
 ```
-convertFromHTML(html: string): ?Array<ContentBlock>
+convertFromHTML(html: string, blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap): ?Array<ContentBlock>
 ```
 
 Given an HTML fragment, convert it to an array of `ContentBlock` objects.
--- a/docs/APIReference-Data-Conversion.md
+++ b/docs/APIReference-Data-Conversion.md
@@ -49,7 +49,10 @@ other usage within an application.
 ### convertFromHTML
 
 ```
-convertFromHTML(html: string, blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap): ?Array<ContentBlock>
+convertFromHTML(
+  html: string,
+  blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap
+): ?Array<ContentBlock>
 ```
 
 Given an HTML fragment, convert it to an array of `ContentBlock` objects.
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -47,17 +47,14 @@ the editor blockRender props.
 
 ```js
 // The example below deliberatly only allows
-// 'heading-two' and 'paragraph' as valid block types and
-// updated the unstyled element to become a paragraph.
+// 'heading-two' as the only valid block type and
+// updates the unstyled element to also become a h2.
 const blockRenderMap = Immutable.Map({
   'header-two': {
    element: 'h2'
   },
-  'paragraph': {
-   element: 'p'
-  },
   'unstyled': {
-   element: 'p'
+    element: 'h2'
   }
 });
 
@@ -80,11 +77,8 @@ this can be done by using the DefaultDraftBlockRenderMap reference to create a n
 
 ```js
 const blockRenderMap = Immutable.Map({
-  'paragraph': {
-    element: 'p'
-  },
-  'unstyled': {
-    element: 'p'
+  'section': {
+    element: 'section'
   }
 });
 
@@ -120,7 +114,7 @@ to wrap any other custom block type
 *example of extending default block render map to use a react component for a custom block:*
 
 ```js
-class myCustomBlock {
+class MyCustomBlock extends React.Component {
   constructor(props) {
     super(props);
   }
@@ -135,9 +129,9 @@ class myCustomBlock {
 }
 
 const blockRenderMap = Immutable.Map({
-  'myCustomBlock': {
+  'MyCustomBlock': {
     element: 'section', // will be used during paste or html conversion to auto match your component
-    wrapper: <myCustomBlock {...this.props} />
+    wrapper: <MyCustomBlock {...this.props} />
   }
 });
 
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -20,13 +20,13 @@ by matching the Draft block render map with the matched tag.
 
 | HTML element    | Draft block type                        |
 | --------------- | --------------------------------------- |
-| `</h1>`         | header-one                              |
-| `</h2>`         | header-two                              |
-| `</h3>`         | header-three                            |
-| `</h4>`         | header-four                             |
-| `</h5>`         | header-five                             |
-| `</h6>`         | header-six                              |
-| `</h6>`         | header-six                              |
+| `<h1/>`         | header-one                              |
+| `<h2/>`         | header-two                              |
+| `<h3/>`         | header-three                            |
+| `<h4/>`         | header-four                             |
+| `<h5/>`         | header-five                             |
+| `<h6/>`         | header-six                              |
+| `<h6/>`         | header-six                              |
 | `<blockquote/>` | blockquote                              |
 | `<pre/>`        | code-block                              |
 | `<figure/>`     | atomic                                  |
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -1,0 +1,148 @@
+---
+id: advanced-topics-custom-block-render-map
+title: Custom Block Rendering
+layout: docs
+category: Advanced Topics
+next: advanced-topics-block-components
+permalink: docs/advanced-topics-custom-block-render-map.html
+---
+
+This article discusses how to customize Draft default block rendering.
+The block rendering is used to define supported block types and their respective
+renderers, as well as converting pasted content to known Draft block types.
+
+When pasting content or when using the
+[convertFromHTML](https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#convertfromhtml)
+Draft will then convert the pasted content to the respective block rendering type
+by matching the Draft block render map with the matched tag.
+
+## Draft default block render map
+
+| HTML element    | Draft block type                        |
+| --------------- | --------------------------------------- |
+| `</h1>`         | header-one                              |
+| `</h2>`         | header-two                              |
+| `</h3>`         | header-three                            |
+| `</h4>`         | header-four                             |
+| `</h5>`         | header-five                             |
+| `</h6>`         | header-six                              |
+| `</h6>`         | header-six                              |
+| `<blockquote/>` | blockquote                              |
+| `<pre/>`        | code-block                              |
+| `<figure/>`     | atomic                                  |
+| `<li/>`         | unordered-list-item,ordered-list-item** |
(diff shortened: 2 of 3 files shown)


docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,148 @@
...
+When pasting content or when using the
+[convertFromHTML](https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#convertfromhtml)
+Draft will then convert the pasted content to the respective block rendering type
+by matching the Draft block render map with the matched tag.
+
+## Draft default block render map
+
+| HTML element    | Draft block type                        |
+| --------------- | --------------------------------------- |
+| `</h1>`         | header-one                              |
```
johanneslumpe: @mitermayer I assume these are all supposed to be `<hX/>` with a trailing slash, no? :)

mitermayer: thanks for spotting that :) will update it now

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,148 @@
...
+    render() {
+      return (<Editor ...
+                blockRenderMap={extendedBlockRenderMap} />);
+    }
+}
+```
+
+## Custom block wrappers
+
+By default the html element is used to wrap block types however a react component
```
mitermayer: I released I should also talk about the grouping that happens with elements that share the same wrapper and are adjacents. Example consecutive `<li/>` will be on the same `<ul/>` or `<ol/>`

mitermayer: Should also create an example to when a custom wrapper could be used

johanneslumpe: @mitermayer might be a little bit off topic but what are your thoughts on providing a function for the wrapper which receives the current wrapper element, the current node to wrap and the previous node, so it would be possible to wrap adjacent items in different groups? Just an idea though, maybe it's crazy.

mitermayer: This functionality was already there, but your idea make sense. maybe could be included on a PR.

docs/APIReference-Data-Conversion.md
```diff
@@ -49,7 +49,7 @@ other usage within an application.
 ### convertFromHTML
 
 ```
-convertFromHTML(html: string): ?Array<ContentBlock>
+convertFromHTML(html: string, blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap): ?Array<ContentBlock>
```
hellendag: We've got an annoying 80-col limit internally, so if you wouldn't mind :)

```js
convertFromHTML(
  html: string,
  blockRenderMap?: DraftBlockRenderMap = ...
): ?Array<ContentBlock>
```

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,148 @@
...
+	'header-two': {
+		element: 'h2'
+	},
+	'paragraph': {
+		element: 'p'
+	},
+	'unstyled': {
+		element: 'p'
+	}
+});
```
hellendag: Another FB guideline: please use two spaces for each indent level, instead of tabs. (This applies throughout.)

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,148 @@
...
+
+// Include 'paragraph' as a valid block and updated the unstyled element but
+// keep support for other draft default block types
+const extendedBlockRenderMap = Draft.DefaultDraftBlockRenderMap.merge(blockRenderMap);
+
+class RichEditor() {
+    ...
+    render() {
+      return (<Editor ...
+                blockRenderMap={extendedBlockRenderMap} />);
```
hellendag:
```js
render() {
  return (
    <Editor
      ...
      blockRenderMap={...}
    />
  );
}
```

hellendag:
Some styling nits. Thanks for the doc!

docs/Advanced-Topics-Custom-Block-Render.md
mitermayer: I remember updating my editorconfig for that, sorry forgot to change for .MD files

mitermayer:
@hellendag Sorry for the styling, the .editorConfig was not picking up the md files. Should not be a problem anymore for further PR's

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,158 @@
...
+// updated the unstyled element to become a paragraph.
+const blockRenderMap = Immutable.Map({
+  'header-two': {
+   element: 'h2'
+  },
+  'paragraph': {
+   element: 'p'
+  },
+  'unstyled': {
+   element: 'p'
```
frenzzy: After the first line break I see in console:

`Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>. See DraftEditorContents > p > ... > DraftEditorBlock > div.`

Maybe you missed something?

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,158 @@
...
+        {this.props.children}
+      </div>
+    );
+  }
+}
+
+const blockRenderMap = Immutable.Map({
+  'myCustomBlock': {
+    element: 'section', // will be used during paste or html conversion to auto match your component
+    wrapper: <myCustomBlock {...this.props} />
```
ayal: should be MyCustomBlock? (capital 'M')

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,158 @@
...
+the html will be scanned for matching tag elements. A wrapper will be used when there is a definition for
+it on the _blockRenderMap_ to wrap that particular block type. For example:
+
+Draft uses wrappers to wrap `<li/>` inside either `<ol/>` or `<ul/>` but wrappers can also be used
+to wrap any other custom block type
+
+*example of extending default block render map to use a react component for a custom block:*
+
+```js
+class myCustomBlock {
```
ayal: shoudl extend 'Component' ?

docs/Advanced-Topics-Custom-Block-Render.md
mitermayer: Will do that :)

docs/Advanced-Topics-Custom-Block-Render.md
mitermayer: Thanks will also update that

docs/Advanced-Topics-Custom-Block-Render.md
mitermayer: This was a feature implemented prior to react 0.15 when this warning did not exist. will update the docs to show a different example

mitermayer:
@hellendag updated the docs, removed reference to 'paragraph' since after latest react it emits that warning. Please let me know what you think and if this PR need any other amends

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,152 @@
...
+const blockRenderMap = Immutable.Map({
+  'header-two': {
+   element: 'h2'
+  },
+  'unstyled': {
+    element: 'h2'
+  }
+});
+
+class RichEditor() {
```
acusti: I believe `class RichEditor() {` should be `class RichEditor extends React.Component {` (also at lines 89 and 141), assuming I understand the examples.

mitermayer: Ahh thanks for spotting this. Not sure how did I missed so many things when righting this documentation. :( but once again thanks for helping spoting this :)

hellendag:
Loaded this up on the website locally.

Looks like there are some whitespace issues with the table and some of the docblocks in the examples. Might require some CSS tweaks to make it render nicely there. :\

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,152 @@
...
+Draft uses wrappers to wrap `<li/>` inside either `<ol/>` or `<ul/>` but wrappers can also be used
+to wrap any other custom block type
+
+*example of extending default block render map to use a react component for a custom block:*
+
+```js
+class MyCustomBlock extends React.Component {
+  constructor(props) {
+    super(props);
+  }
```
flarnie: I don't think we need to include this 'constructor' in the example.

flarnie:
Hi @mitermayer! Thanks for adding this. I think it looks great - I commented on a nit but I think that it could be fixed in a follow-up. I'm going to spin up the site locally and verify that things look ok, and I plan on merging this if no blockers come up.

Some questions;

- We now have `blockRendererFn` and `blockRenderMap`, and they might seem similar to some users. What do you think of the idea of adding documentation about the differences and the way that these properties work together? Specifically, we could include an explanation of how each one affects the block that is output. Perhaps an example of using both. Or maybe just showing the structure:
```
+----------+wrapper +->set by blockRenderMap
|
|    +-----+element w/ props +-->set by blockRenderMap
|    |
|    |   +-+component w/ props +-->set by blockRendererFn
|    |   |
|    |   |
|    |   +-+
|    |
|    +-----+
|
|    // ... more blocks within the wrapper
|
+----------+
```

- Another thing we could add in the future; an example of using a custom ReactClass for the `element` in the `blockRenderMap`.

We could open issues for both those as follow-up items.

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,152 @@
...
+|     `<h4/>`     |               header-four               |
+|     `<h5/>`     |               header-five               |
+|     `<h6/>`     |               header-six                |
+|     `<h6/>`     |               header-six                |
+| `<blockquote/>` |               blockquote                |
+|    `<pre/>`     |               code-block                |
+|   `<figure/>`   |                 atomic                  |
+|     `<li/>`     | unordered-list-item,ordered-list-item** |
+|    `<div/>`     |               unstyled***               |
+
```
flarnie: For some reason this table is still not aligned in the rendered documentation. Also we can omit the dash after the asterisks in the comments below the table.
screen shot 2016-08-27 at 6 52 44 pm

mitermayer:
Thanks guys for the feedback, Please feel free to contribute to this PR, I am unable to contribute to Draft until after October 15th so feel free to update this PR.

flarnie:
Thanks @mitermayer! I'll see if I can fix the way that table renders then.

flarnie:
I [added some styles for the table](https://github.com/facebook/draft-js/pull/629), and I think that was the only blocker.

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -0,0 +1,152 @@
...
+        {this.props.children}
+      </div>
+    );
+  }
+}
+
+const blockRenderMap = Immutable.Map({
+  'MyCustomBlock': {
+    element: 'section', // will be used during paste or html conversion to auto match your component
+    wrapper: <MyCustomBlock {...this.props} />
```
AlastairTaft: Is this a mistake, I think it would cause a runtime error? What does the `this` refer to in this context?

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -80,11 +77,8 @@ this can be done by using the DefaultDraftBlockRenderMap reference to create a n
...
 const blockRenderMap = Immutable.Map({
-  'paragraph': {
-    element: 'p'
-  },
-  'unstyled': {
-    element: 'p'
+  'section': {
+    element: 'section'
   }
 });
```
mkhazov: you forgot to replace `'paragraph'` with '`section`' in the comment below:
```
 // Include 'paragraph' as a valid block and updated the unstyled element but
 // keep support for other draft default block types```

PR #389, opened 2016-05-12
Generate .js.flow files

### LINKED ISSUE #366: Expose flow.
Hey,
As I can see draft-js uses flow and has flow annotations.
Would it be possible to leave them in in the dist?
This would allow people who use flow to take advantage of them.

zpao:
> I've done this before, it'll only take a few minutes
> — @zpao, before he realized upgrading Babel was a prereq

The main goal here was to generate `.js.flow` files and fix #366. Then I needed to upgrade Babel, which meant using the new fbjs preset, which then needed to be updated as it wasn't as done as we thought since Draft has different considerations than fbjs. So now we need https://github.com/facebook/fbjs/pull/147 and https://github.com/facebook/fbjs/pull/148 to land there and ship before we can take this.

The [diff of lib/](https://gist.github.com/zpao/32c511aa1151aef1009eccb4f1cf0a5f) is pretty boring. Mostly it's a result of a few things:
- the helpers are different, this is particularly noticeable for classes
- code generation around comments got better - previously block comments near rewritten nodes (notably `require`) would be moved but now stay where they should
- conversely there are a bunch of newlines being inserted now. win some, lose some.
- a few errant `__esModule` markers are gone, which is good. Looks like those were a result of the type exports being erroneously detected as real exports. This also means that type-only "modules" are now empty. It might be good to filter those out in the future but in the mean time I don't think they hurt except that there are a few more files.


package.json
```diff
@@ -90,6 +92,7 @@
       "<rootDir>/node_modules/fbjs/lib/UserAgentData.js",
       "<rootDir>/node_modules/fbjs-scripts/",
       "<rootDir>/node_modules/immutable/",
+      "<rootDir>/node_modules/object-assign/",
```
spicyj: Why did this change?

spicyj:
👍  otherwise. Do you know if there's any difference with the new class helpers?

package.json
zpao: Code now has `require('object-assign')` because I brought that plugin over to the preset. Jest will mock that and so a bunch of tests fail.

zpao:
It looks like the biggest change is actually that the preset has "loose" classes turned on (which maybe it shouldn't…) but we weren't doing that previously in this project (but we were in most others). So there are probably some subtle things there.

ryyppy:
Hey @zpao I see that your blocking issues facebook/fbjs#147 and facebook/fbjs#148 are merged, is there any way to make this PR work now?

nikgraf:
Seems like the script is missing. I guess the PR is just a little outdated. Really looking forward to get all the flow-types …

```
Error: Cannot find module './scripts/module-map'
```

paulyoung:
I merged this with `master` locally, fixed the conflicts, and got the same result.

@zpao – did you forget to commit `./scripts/module-map`?

zpao:
I'll be coming back to this shortly.

zpao:
@facebook-github-bot import

Issue #448, opened 2016-06-08
handleBeforeInput and deletions


paulyoung:
I'm currently making some changes to the editor state in `handleBeforeInput`.

One issue I'm seeing is that `handleBeforeInput` doesn't get called for deletions, which leads to undesirable results.

While I could possibly work around this by using `handleKeyCommand` for the `delete` and `backspace` commands, that still doesn't account for text removed by cutting (⌘ + x) on OS X.

What would be the recommended way to address this?

hellendag:
Right, `handleBeforeInput` is only intended to handle character insertions, not removal. For deletion behavior, `handleKeyCommand` is the approach I would recommend.

The Cut command is trickier. One option could be to add the command to the key binding and allow the caller to handle it manually, though this delegates a lot of responsibility and risk -- see the complexity of [editOnCut](https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnCut.js). :)

Can you provide more details about the behavior of your editor? What kinds of state changes do your `handleBeforeInput` and `handleKeyCommand` make?

paulyoung:
@hellendag I'm handing off the input to a lexer, parser, (and soon type checker) then providing syntax highlighting by updating the `CharacterList` and using `replaceText` or `replaceBlockFragment`. I'm also underlining the text at the error position and providing error information.

I initially used a decorator as mentioned in https://github.com/facebook/draft-js/issues/47#issuecomment-223049922 but it was too limiting. So far using `handleBeforeInput` I can:

- surface error messages beyond the decorated text in a reasonable way
- access the previous lexemes and potentially only update a subset of the text, instead of processing everything again on every keystroke

It also seems like a better approach for adding more features such as autocomplete.

hellendag:
Interesting!

One question: would it be overkill to always perform your highlighting behavior in `onChange` if the `ContentState` has changed?

If not, you'd then always be able to know that some content is different, regardless of how it changed. Perhaps you could compare the `blockMap` objects of the distinct `ContentState` objects to identify differing lines, then perform your processing on only those lines.

FWIW, for Facebook mention autocomplete, we always trigger autocompletion in the parent component's  `componentDidUpdate` if the selection is collapsed. Then we just pay attention to the selected block (at the selection offset, of course) and can ignore everything else. Further, we can use information about when the selection state changes so that we can trigger autocompletion if the user just moves the cursor to another position in the editor.

paulyoung:
[...]

Issue #427, opened 2016-05-31
Advice on setting ContentState blockMap


paulyoung:
I'm expanding on [this example](https://github.com/facebook/draft-js/blob/8689bcd0a9b3aff402db0720789f3d4028645ef1/docs/QuickStart-API-Basics.md#controlling-rich-text) a little bit and running into an issue.

This works just fine (effectively a no-op):

```javascript
onChange(editorState) {
  const contentState = editorState.getCurrentContent();
  const blockMap = contentState.getBlockMap();
  const newContentState = contentState.set('blockMap', blockMap);
  const newEditorState = EditorState.push(editorState, newContentState, 'change-inline-style');
  this.setState({
    editorState: newEditorState,
  });
}
```

but this produces some strange behavior (still effectively a no-op):

```javascript
onChange(editorState) {
  const contentState = editorState.getCurrentContent();
  const blockMap = contentState.getBlockMap();
  const newBlockMap = blockMap.map((contentBlock) => {
    return contentBlock;
  });
  const newContentState = contentState.set('blockMap', newBlockMap);
  const newEditorState = EditorState.push(editorState, newContentState, 'change-inline-style');
  this.setState({
    editorState: newEditorState,
  });
}
```

More specifically – using the latter, I can type `h` but by the time I've typed `he` it appears as `ehe` and the cursor is after the first letter. This continues creating palindromes like so:

| Typed | Displayed |
| --- | --- |
| h | h |
| he | ehe |
| hel | lehel |
| hell | llehell |
| hello | ollehello |

paulyoung:
FWIW, I've also tried various approaches using the static methods on `Modifier`, all with the same results.

hellendag:
This is a true no-op:

```js
const blockMap = contentState.getBlockMap();
const newContentState = contentState.set('blockMap', blockMap);
```

After this point, `contentState` and `newContentState` are the same object, since `blockMap` is unchanged. The `set` operation is a no-op. This, however...

```js
const newBlockMap = blockMap.map((contentBlock) => {
  return contentBlock;
});
const newContentState = contentState.set('blockMap', newBlockMap);
```

...is not a no-op, since `newBlockMap` is not the same object as the original `blockMap`. Thus, `contentState` and `newContentState` are also distinct objects.

[...]

PR #453, opened 2016-06-09
Add support for preventing native character insertion

paulyoung:
@hellendag this intends to address the issue we discussed in https://github.com/facebook/draft-js/issues/427 and https://github.com/facebook/draft-js/issues/448.

The intended use is something like this:

```javascript
onChange(editorState) {
  const contentState = editorState.getCurrentContent();
  const newBlockMap = blockMap.map((contentBlock) => {
    // perform some updates here
    return contentBlock;
  });
  const newContentState = contentState.set('blockMap', newBlockMap);
  const newEditorState = EditorState.push(editorState, newContentState, 'insert-fragment');
  this.setState({
    editorState: EditorState.preventNativeInsertion(newEditorState, true),
  });
}
```

Please review and let me know if you have any feedback.

Thanks!

### DIFF
--- a/docs/APIReference-EditorState.md
+++ b/docs/APIReference-EditorState.md
@@ -110,11 +110,6 @@ The list below includes the most commonly used instance methods for `EditorState
       <pre>static setInlineStyleOverride(inlineStyleOverride): EditorState</pre>
     </a>
   </li>
-  <li>
-    <a href="#preventnativeinsertion">
-      <pre>static preventNativeInsertion(editorState, preventNativeInsertion): EditorState</pre>
-    </a>
-  </li>
 </ul>
 
 *Properties*
@@ -168,11 +163,6 @@ The list below includes the most commonly used instance methods for `EditorState
   <li>
     <a href="#nativelyrenderedcontent">
       <pre>nativelyRenderedContent</pre>
-    </a>
-  </li>
-  <li>
-    <a href="#preventnativeinsertion">
-      <pre>preventNativeInsertion</pre>
     </a>
   </li>
   <li>
@@ -359,19 +349,6 @@ static setInlineStyleOverride(inlineStyleOverride: DraftInlineStyle): EditorStat
 Returns a new `EditorState` object with the specified `DraftInlineStyle` applied
 as the set of inline styles to be applied to the next inserted characters.
 
-### preventNativeInsertion
-
-```
-static preventNativeInsertion(
-  editorState: EditorState,
-  preventNativeInsertion: boolean
-): EditorState
-```
-Returns a new `EditorState` object with the specified value applied.
-
-This is intended to be used when EditorState has been updated from within
-`onChange`, to prevent duplicate character insertion.
-
 ## Properties and Getters
 
 In most cases, the instance and static methods above should be sufficient to
@@ -484,17 +461,6 @@ In doing so, we can avoid extra re-renders and preserve spellcheck highlighting.
 When allowing native rendering behavior, it is appropriate to use the
 `nativelyRenderedContent` property to indicate that no re-render is necessary
 for this `EditorState`.
-
-### preventNativeInsertion
-
-```
-preventNativeInsertion: boolean;
-mustPreventNativeInsertion()
-```
-Whether to prevent native insertion.
-
-You should not set this property manually -- see
-[preventNativeInsertion()](#preventnativeinsertion).
 
 ### redoStack
 
--- a/docs/APIReference-EditorState.md
+++ b/docs/APIReference-EditorState.md
@@ -110,6 +110,11 @@ The list below includes the most commonly used instance methods for `EditorState
       <pre>static setInlineStyleOverride(inlineStyleOverride): EditorState</pre>
     </a>
   </li>
+  <li>
+    <a href="#preventnativeinsertion">
+      <pre>static preventNativeInsertion(editorState, preventNativeInsertion): EditorState</pre>
+    </a>
+  </li>
 </ul>
 
 *Properties*
@@ -163,6 +168,11 @@ The list below includes the most commonly used instance methods for `EditorState
   <li>
     <a href="#nativelyrenderedcontent">
       <pre>nativelyRenderedContent</pre>
+    </a>
+  </li>
+  <li>
+    <a href="#preventnativeinsertion">
+      <pre>preventNativeInsertion</pre>
     </a>
   </li>
   <li>
@@ -349,6 +359,19 @@ static setInlineStyleOverride(inlineStyleOverride: DraftInlineStyle): EditorStat
 Returns a new `EditorState` object with the specified `DraftInlineStyle` applied
 as the set of inline styles to be applied to the next inserted characters.
 
+### preventNativeInsertion
+
+```
+static preventNativeInsertion(
+  editorState: EditorState,
+  preventNativeInsertion: boolean
+): EditorState
+```
+Returns a new `EditorState` object with the specified value applied.
+
+This is intended to be used when EditorState has been updated from within
+`onChange`, to prevent duplicate character insertion.
+
 ## Properties and Getters
 
 In most cases, the instance and static methods above should be sufficient to
@@ -461,6 +484,17 @@ In doing so, we can avoid extra re-renders and preserve spellcheck highlighting.
 When allowing native rendering behavior, it is appropriate to use the
 `nativelyRenderedContent` property to indicate that no re-render is necessary
 for this `EditorState`.
+
+### preventNativeInsertion
+
+```
+preventNativeInsertion: boolean;
+mustPreventNativeInsertion()
+```
+Whether to prevent native insertion.
+
+You should not set this property manually -- see
+[preventNativeInsertion()](#preventnativeinsertion).
 
 ### redoStack
 
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -116,6 +116,8 @@ export type DraftEditorProps = {
     isInternal: DraftDragType
   ) => boolean,
 
+  allowNativeInsertion?: (chars: string) => boolean,
+
   /**
    * Non-cancelable event triggers.
    */
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -142,7 +142,6 @@ function editOnBeforeInput(e: SyntheticInputEvent): void {
   );
 
   if (
-    editorState.mustPreventNativeInsertion() ||
     mustPreventDefaultForCharacter(chars) ||
     originalFingerprint !== newFingerprint ||
     (
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -124,7 +124,12 @@ function editOnBeforeInput(e: SyntheticInputEvent): void {
     )
   );
 
-  if (!this.props.allowNativeInsertion(chars) || !mayAllowNative) {
+  var allowNativeInsertion = this.props.allowNativeInsertion;
+
+  if (
+    (allowNativeInsertion && !allowNativeInsertion(chars)) ||
+    !mayAllowNative
+  ) {
     e.preventDefault();
     this.update(newEditorState);
     return;
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -142,6 +142,7 @@ function editOnBeforeInput(e: SyntheticInputEvent): void {
   );
 
   if (
+    editorState.mustPreventNativeInsertion() ||
     mustPreventDefaultForCharacter(chars) ||
     originalFingerprint !== newFingerprint ||
     (
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -124,7 +124,7 @@ function editOnBeforeInput(e: SyntheticInputEvent): void {
     )
   );
 
-  if (!mayAllowNative) {
+  if (!this.props.allowNativeInsertion(chars) || !mayAllowNative) {
     e.preventDefault();
     this.update(newEditorState);
     return;
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -356,7 +356,8 @@ class EditorState {
   static push(
     editorState: EditorState,
     contentState: ContentState,
-    changeType: EditorChangeType
+    changeType: EditorChangeType,
+    preventNativeInsertion: boolean
   ): EditorState {
     if (editorState.getCurrentContent() === contentState) {
       return editorState;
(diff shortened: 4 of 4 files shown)


docs/APIReference-EditorState.md
```diff
@@ -349,6 +359,19 @@ static setInlineStyleOverride(inlineStyleOverride: DraftInlineStyle): EditorStat
 Returns a new `EditorState` object with the specified `DraftInlineStyle` applied
 as the set of inline styles to be applied to the next inserted characters.
 
+### forceSelection
```
paulyoung: I'll fix this and force push.

docs/APIReference-EditorState.md
```diff
@@ -462,6 +485,17 @@ When allowing native rendering behavior, it is appropriate to use the
 `nativelyRenderedContent` property to indicate that no re-render is necessary
 for this `EditorState`.
 
+### forceSelection
```
paulyoung: And this.

paulyoung:
Come to think of it, this probably doesn't need to be re-set every time `onChange` is called.

Instead, I think it could be set when providing the initial `editorState` in which case `preventNativeInsertion: boolean` could also be added to the `Editor` `props` for convenience.

spicyj:
Once you set this, it's on forever? Does it make sense to automatically set it to false after an update?

paulyoung:
> Once you set this, it's on forever?

Currently, whatever value is set remains set until it is explicitly changed. I'm not at all opposed to that changing.

> Does it make sense to automatically set it to false after an update?

@spicyj where do you think that should happen?

spicyj:
Not sure, sorry. Just seemed weird from a first glance.

paulyoung:
@hellendag (and @spicyj)  – if this at least inspires a more desirable solution from either of you I'd be happy to work on that if I could get some direction.

nikgraf:
@paulyoung What about having a fourth argument in `EditorState.push` which by default is `false`? This way you don't have to deal we automatic resets.

I wonder what implications this has for the undo/redo history. Have you tried them in combination with this change?

paulyoung:
@nikgraf – thanks for the input. I _really_ appreciate it.

> I wonder what implications this has for the undo/redo history. Have you tried them in combination with this change?

With the current changeset, it does appear that things aren't fully taken care of. It seems that:
  - both the native and programatic insertion still create an entry (2 items in the stack for 1 change)
  - it's not possible to undo (and likely redo)

> What about having a fourth argument in EditorState.push which by default is false? This way you don't have to deal we automatic resets.

I'll explore that tomorrow. Thanks again for the idea!

paulyoung:
> - it's not possible to undo (and likely redo)

To clarify – ⌘ + z doesn't do anything

paulyoung:
I just pushed a couple of commits in line with @nikgraf's comments.

I still had to update the original code example to update the selection state:

```javascript
onChange(editorState) {
    const contentState = editorState.getCurrentContent();
    const selectionState = editorState.getSelection();
    const newBlockMap = blockMap.map((contentBlock) => {
      // perform some updates here
      return contentBlock;
    });
    const newContentState = contentState.set('blockMap', newBlockMap);
    const newEditorState = EditorState.push(editorState, newContentState, 'insert-fragment', true);
    this.setState({
      editorState: EditorState.acceptSelection(newEditorState, selectionState),
    });
  }
```

This _almost_ works as expected, the only exceptions being a couple of anomalies with the cursor position on undo/redo.

If anyone could take a look and has any advice, I'd really appreciate it.

Once again, my use case is described in https://github.com/facebook/draft-js/issues/448#issuecomment-224648304

paulyoung:
@hellendag – I'd love to keep this going and arrive at an acceptable solution, but think I may need some help/input from you (or someone who has more experience with Draft.js) to get there.

If there's any way I can make that easy for you, or you have any suggestions, please let me know.

hellendag:
I don't think this belongs on `EditorState`. This setting concerns the component and its event management, not the model. This seems like it belongs as a prop on the component.

spicyj:
It seems similar to forceSelection to me – don't you think?

hellendag:
> It seems similar to forceSelection to me – don't you think?

Interesting point. `EditorState` does encode a handful of properties that are based on the behavior of the rendered editor rather than being purely model state. So maybe it's fine, though in hindsight it feels awkward and I wonder if there's a better way to represent that kind of state. Oh well.

A question on usage, whether it's a state value or a prop: how will I know when to set this? I need to set it before `onBeforeInput` fires, but how can I be sure when I am going to need it? Would one typically just turn it on and leave it on?

> FWIW – I considered an alternative approach which allowed setting a new prop on Editor:

This actually sounds like a pretty good approach to me. The function prop doesn't *handle* the insertion, but it determines whether the value should be allowed natively. So that way, rather than setting a boolean value in the overall state of the editor, you would essentially perform filtering on values as they arrive, or use some other information to determine how to handle the character. A contrived example:

```js
allowNativeInsertion(chars: string): boolean {
  if (chars === 'c') {
    // we want to handle `c` manually, no native insertion allowed
    return false;
  }
  return true;
}

// within your `Editor`
<Editor
  ...
  allowNativeInsertion={allowNativeInsertion}
  editorState={editorState}
/>

// within `onBeforeInput`
if (
  !this.props.allowNativeInsertion(chars) ||
  ...
) {
  // handle manually
  return;
}
```

This seems okay to me, and it gives you an opportunity to selectively decide how to manage the incoming insertion.

paulyoung:
Thanks for the thoughtful replies! I'll go back and try this approach this afternoon.

paulyoung:
I've removed the functionality around this that I added to `EditorState` and added a function called `allowNativeInsertion` to the `Editor` props as discussed.

The way this is called in `editOnBeforeInput` may require some discussion.

This appears to work well except for undo/redo again. I attempted to handle this in https://github.com/facebook/draft-js/pull/453/commits/30791d6d0d6d0a288773cf766dab67a0ede53d97 via `EditorState` but now I'm not sure how to approach it.

Now, in conjunction with the code in my original comment https://github.com/facebook/draft-js/pull/453#issue-159311158, items on the `undoStack` have a `blockMap` where the `text` is an empty string (`""`).

paulyoung:
I may have bandwidth to take a stab at fixing undo/redo with the latest changes soon.

If anyone has any ideas, I'd love to hear them!

paulyoung:
I took a quick look at this again and I'm having a bit of trouble seeing how to affect undo/redo now that `EditorState` isn't involved.

rygine:
This issue comes up a lot in our Slack. Would be nice to have this implemented soon.

hellendag:
@paulyoung Would you mind clarifying a bit what the undo/redo issues are for your current solution?

When you perform an undo, what happens? Is the issue that the `ContentState` with the character insertion becomes an unwanted boundary state?

I appreciate the simplicity of this change, so it would be great to get this to work. :)

paulyoung:
@hellendag I decided to first examine the undo stack before applying the changes in this pull request.

This appears to exhibit the same behavior (so while it doesn't seem like a problem caused by the code I've introduced) it's a problem that needs to be addressed _somehow_ nonetheless.

Below is the `onChange` handler I used to log the undo stack. I decided that selection state wasn't relevant to the discussion just yet.

```javascript
_onChange(editorState) {
    const contentState = editorState.getCurrentContent();
    const selectionState = editorState.getSelection();
    const newBlockMap = blockMap.map((contentBlock) => {
      // perform some updates here
      return contentBlock;
    });
    const newContentState = contentState.set('blockMap', newBlockMap);
    const newEditorState = EditorState.push(editorState, newContentState, 'insert-fragment', true);
    console.log(JSON.stringify(newEditorState.toJS().undoStack.map((x) => x.blockMap), null, 2));
    this.setState({
      editorState: EditorState.acceptSelection(newEditorState, selectionState),
    });
  }
```

If there's a way that I can change this to "fix" the undo/redo stack that would be great, although I think a baked-in solution that "just works" would be nice.

Here is what was logged for each action (expand to view):

<details>
<summary>Initial page load</summary>
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
</details>

<details>
<summary>Focus</summary>
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
</details>

<details>
<summary>Type "a"</summary>
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
</details>

<details>
<summary>Type "b" (note that "b" is repeated because my changes aren't applied)</summary>
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "ab",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "abb",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "ab",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "ab",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
</details>

<details>
<summary>Blur</summary>
```json
[
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "abb",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "abb",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "ab",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "ab",
      "characterList": [
        {
          "style": [],
          "entity": null
        },
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "a",
      "characterList": [
        {
          "style": [],
          "entity": null
        }
      ],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  },
  {
    "5msnk": {
      "key": "5msnk",
      "type": "unstyled",
      "text": "",
      "characterList": [],
      "depth": 0
    }
  }
]
```
</details>

paulyoung:
It seems to me that since the approach taken with the latest changes doesn't support a way to address the issues with the undo/redo stack, I should revert to the `forceSelection`-style method I tried earlier and see if I can figure out undo/redo from there.

hellendag:
Still not completely sure I understand, but it seems like in the logged output, for each step there is an extra `ContentState` object in the undo stack that it would be preferable not to have. Am I understanding correctly? Is it that you would like to discard the top of the undo stack before pushing on the new `ContentState` in your `onChange`?

paulyoung:
I'm going to get back to this soon, I promise 🙇

paulyoung:
Quick update on this – discarding the top of the undo stack _does_ appear to address the issues I was having with duplicate characters, when paired with this changes in this branch.

I still can't get undo/redo to work, although that appears to be a problem with the way I'm handling `onChange`, so going to look into that.

spicyj:
My #667 should fix this cleanly so let's go with that.

paulyoung:
Oh, awesome @spicyj! Would love to see this land soon!

AnilRedshift:
@paulyoung I still would like this PR to go through. Half of the point of draftjs is that contenteditable is essentially unspecced. I am very hesitant to rely on the browser to do the right thing here. I would prefer to at the very minimum give the ability for the caller to disable native behavior.

PR #471, opened 2016-06-15
Implemented insertAtomicBlockBefore and insertAtomicBlockAfter

delijah:
The function "insertAtomicBlock" within the module "AtomicBlockUtils" does not really make sense to me. To insert an atomic block, it splits an existing block and deletes text of the preceding block etc.

This is not what i want.

What i want is simply to insert a new atomic block, without affecting other blocks, but the order of the blockMap.
In this case you have to decide always if you want to insert the new atomic block before or after an other block. This is why there are to new functions: "insertAtomicBlockBefore" and "insertAtomicBlockAfter"

Issues i am not sure about:
- Do you agree with me that this is a useful feature?
- I am not sure if i have to do some entity handling since the functions move a new block inbetween two already existing blocks (which might have entities already)
- Please check my code since this is the first time i am contributing to the draft-js project
- Do i have to implement some tests?

### DIFF
--- a/docs/APIReference-AtomicBlockUtils.md
+++ b/docs/APIReference-AtomicBlockUtils.md
@@ -24,3 +24,23 @@ insertAtomicBlock: function(
   character: string
 ): EditorState
 ```
+
+### insertAtomicBlockBefore
+
+```
+insertAtomicBlockBefore: function(
+  editorState: EditorState,
+  entityKey: string,
+  character: string
+): EditorState
+```
+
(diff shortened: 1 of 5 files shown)


facebook-github-bot-6:

delijah:
It looks like i do not understand the insertAtomicBlock function to it's detail yet. I'll create a pull request again when i do so and have another proposal...

Superoryco:
#713 

Hi @delijah,

Will your branch solve my issue?

Issue #476, opened 2016-06-20
Double click link in Firefox broke selection


ouchxp:
In the link example. After double click the link, it will not select the link or the word, instead of selecting the text before the link. Having this issue only in Firefox.
![image](https://cloud.githubusercontent.com/assets/1253729/16180873/73c09bea-36e5-11e6-8dc6-1944d6b54059.png)

The problem is because `getSelectionOffsetKeyForNode` function does not handle the link decorator properly. The decorator node (`a` tag) may not have a `data-offset-key` attribute. But it's children could have. So it's not correct to only get `data-offset-key` attribute from decorator node without checking it's children.
The related tests are assuming decorator to have `data-offset-key` attribute, which is wrong as well.

ouchxp:
I've create a pull request to fix this issue https://github.com/facebook/draft-js/pull/477

ouchxp:
Patch merged. Close this issue.

PR #477, opened 2016-06-20
Decorator do not have `data-offset-key` attribute. Fix `getSelectionOffsetKeyForNode` function

ouchxp:
Decorator do not have `data-offset-key` attribute. Fix the implementation of `getSelectionOffsetKeyForNode` function to resolve the `data-offset-key` from decorator's children and fix tests.

Related issue https://github.com/facebook/draft-js/issues/476

### DIFF
--- a/src/component/selection/__tests__/getDraftEditorSelection-test.js
+++ b/src/component/selection/__tests__/getDraftEditorSelection-test.js
@@ -131,7 +131,8 @@ describe('getDraftEditorSelection', function() {
       .map(
         function(decoratorKey, ii) {
           var span = document.createElement('span');
-          span.setAttribute('data-offset-key', '' + decoratorKey);
+          // decorator do not have data-offset-key attribute
+          span.setAttribute('decorator-key', '' + decoratorKey);
           span.appendChild(leafs[(ii * 2)]);
           span.appendChild(leafs[(ii * 2) + 1]);
           return span;
--- a/src/component/selection/__tests__/getDraftEditorSelection-test.js
+++ b/src/component/selection/__tests__/getDraftEditorSelection-test.js
@@ -131,7 +131,7 @@ describe('getDraftEditorSelection', function() {
       .map(
         function(decoratorKey, ii) {
           var span = document.createElement('span');
-          // decorator do not have data-offset-key attribute
+          // Decorators may not have `data-offset-key` attribute
           span.setAttribute('decorator-key', '' + decoratorKey);
           span.appendChild(leafs[(ii * 2)]);
           span.appendChild(leafs[(ii * 2) + 1]);
--- a/src/component/selection/getSelectionOffsetKeyForNode.js
+++ b/src/component/selection/getSelectionOffsetKeyForNode.js
@@ -17,7 +17,15 @@
  * Get offset key from a node.
  */
 function getSelectionOffsetKeyForNode(node: Node): ?string {
-  return node instanceof Element ? node.getAttribute('data-offset-key') : null;
+  if (node instanceof Element) {
+    var offsetKey = node.getAttribute('data-offset-key');
+    if (offsetKey) return offsetKey;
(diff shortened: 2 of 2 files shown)


hellendag:
Weird. What `SelectionState` is generated when double-clicking the link? What happens when stepping through https://github.com/facebook/draft-js/blob/master/src/component/selection/getDraftEditorSelectionWithNodes.js#L34?

ouchxp:
@hellendag 
Using the example in this issue https://github.com/facebook/draft-js/issues/476

`global.getSelection()` returns 
```
Selection { anchorNode: <div.public-DraftStyleDefault-block.public-DraftStyleDefault-ltr>, anchorOffset: 1, focusNode: <div.public-DraftStyleDefault-block.public-DraftStyleDefault-ltr>, focusOffset: 2, isCollapsed: false, rangeCount: 1, caretBidiLevel: 0 }
```
Which indicates selection is the second child of the block, the link tag.

Inside `getDraftEditorSelectionWithNodes` function, it calls `getPointForNonTextNode` to get the `focusOffset`
Then in `getPointForNonTextNode` use `getSelectionOffsetKeyForNode` to find out the selected node is Leaf/Text or a Parent of Leaf. (This is the place it couldn't retrieve `data-offset-key` attribute from link decorator. So it consider the decorator as a Leaf/Text node.)

Finally, the `SelectionState` returned by `getDraftEditorSelectionWithNodes` is
```
Object { anchorKey: "9o6md", anchorOffset: 49, focusKey: "9o6md", focusOffset: 6, isBackward: true, hasFocus: true }
```

ouchxp:
@hellendag I've check the Facebook Note, it does not have this problem because they added the `data-offset-key` to the link decorator. It's another way to make it work.

src/component/selection/__tests__/getDraftEditorSelection-test.js
```diff
@@ -131,7 +131,8 @@ describe('getDraftEditorSelection', function() {
       .map(
         function(decoratorKey, ii) {
           var span = document.createElement('span');
-          span.setAttribute('data-offset-key', '' + decoratorKey);
+          // decorator do not have data-offset-key attribute
```
hellendag:
```javascript
// Decorators may not have `data-offset-key` attribute
```

src/component/selection/getSelectionOffsetKeyForNode.js
```diff
@@ -17,7 +17,15 @@
  * Get offset key from a node.
```
hellendag: Probably a good idea to flesh out the docblock for this function, now that it does more than just check the current node.

src/component/selection/getSelectionOffsetKeyForNode.js
```diff
@@ -17,7 +17,15 @@
...
  */
 function getSelectionOffsetKeyForNode(node: Node): ?string {
-  return node instanceof Element ? node.getAttribute('data-offset-key') : null;
+  if (node instanceof Element) {
+    var offsetKey = node.getAttribute('data-offset-key');
+    if (offsetKey) return offsetKey;
```
hellendag: Styling nit. We should have eslint catch this.

```javascript
if (offsetKey) {
  return offsetKey;
}
```

Same for the `childOffsetKey` case below.

src/component/selection/getSelectionOffsetKeyForNode.js
```diff
@@ -17,7 +17,15 @@
...
+    for (var ii = 0; ii < node.children.length; ii++) {
```
hellendag: I'd prefer `childNodes` to `children` here, even though it means iterating over nodes that will return null when recursing.

hellendag:
Thanks for the details. I think that sounds reasonable.

ouchxp:
Thanks for the comments, I've fix these issues you've point out. Please feel free to leave more comments if there are still something need to be change.

hellendag:
Cool, let's go with it.

PR #482, opened 2016-06-22
Fix insertAtomicBlock function

delijah:
Hey!

I'd like to enhance the insertAtomicBlock function. Before i start with my explanations, i try to define what is an atomic block in my eyes. Maybe i have a wrong understanding of atomic blocks, so this pull request starts on the wrong edge? Maybe not? We'll see.

In my eyes an atomic block does not only contain text, it can contain images, videos etc. Maybe there is no editable text in it, maybe there is. But from the view of the editor, this is just a block. The editor does not know what is inside. I would consider an atomic block, what for example for CKEditor is a widget. Actually this is where i am coming from, i'd like to implement "widgets" in a draft-js editor.

How i think the insertAtomicBlock function should work:
- After calling insertAtomicBlock, an atomic block should be inserted at the current selection
- If the current selection is at the beginning of a block, the new atomic block should be inserted before that block
- If the current selection is inbetween a text in a block, the block should be splitted and the new atomic block should be inserted inbetween the splitted blocks
- If the current selection is at the end of a block, the new atomic block should be inserted after that block
- There is no need for inserting an extra empty block before or after an atomic block
- The insertAtomicBlock function mustn't affect the styling of the blocks before and/or after
- An atomic block can have an entity or not. It mustn't be mandatory to define/create an entity for an atomic block
- I you want to insert an atomic block after/before an empty block, you need a function insertAtomicBlockAfter/insertAtomicBlockBefore, since you cannot determine the insertion direction by the selection

According to this i've created the this pull request.

Why? Problems with the current insertAtomicBlock implementation:
- It creates an extra empty block before or after the inserted block
- It affects the styling of blocks before and/or after the inserted atomic block
- It is mandatory to create an entity for an atomic block
- You cannot insert an atomic block before or after an empty block

In my eyes an atomic block should also be able to have a focus. Same principle than widgets in CKEditor for example. But i'll leave this to another pull request...

Please let me know what you think. Would be great to see this feature in the master.

Cheers, delijah

### DIFF
--- a/src/model/modifier/AtomicBlockUtils.js
+++ b/src/model/modifier/AtomicBlockUtils.js
@@ -52,7 +52,7 @@ const AtomicBlockUtils = {
       'atomic'
     );
 
-    const charData = CharacterMetadata.create({entity: entityKey});
+    const charData = CharacterMetadata.create(entityKey ? {entity: entityKey} : undefined);
 
     const fragmentArray = [
       new ContentBlock({
--- a/src/model/modifier/AtomicBlockUtils.js
+++ b/src/model/modifier/AtomicBlockUtils.js
@@ -61,12 +61,6 @@ const AtomicBlockUtils = {
         text: character,
         characterList: List(Repeat(charData, character.length)),
       }),
-      new ContentBlock({
-        key: generateRandomKey(),
-        type: 'unstyled',
-        text: '',
-        characterList: List(),
-      }),
     ];
 
     const fragment = BlockMapBuilder.createFromArray(fragmentArray);
--- a/src/model/modifier/AtomicBlockUtils.js
+++ b/src/model/modifier/AtomicBlockUtils.js
@@ -29,6 +29,22 @@ const {
   Repeat,
 } = Immutable;
 
+function createAtomicBlock(
+  entityKey: string,
+  character: string
+): ContentBlock {
+  const charData = CharacterMetadata.create(
+    entityKey ? {entity: entityKey} : undefined
+  );
+
+  return new ContentBlock({
+    key: generateRandomKey(),
+    type: 'atomic',
+    text: character,
+    characterList: List(Repeat(charData, character.length)),
+  });
+}
+
 const AtomicBlockUtils = {
   insertAtomicBlock: function(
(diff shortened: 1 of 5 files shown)


hellendag:
Thanks for the detailed explanation!

> It creates an extra empty block before or after the inserted block

This is not ideal, but is present largely due to unresolved selection issues around atomic blocks, especially if the inserted block is the first or last block in the editor. If this can be eliminated, I'm open to it.

> It is mandatory to create an entity for an atomic block

This may be avoidable with the change to add block metadata, but for now, I'm not sure how you'd represent the features needed for your block without an entity. What do you propose there?

> You cannot insert an atomic block before or after an empty block

Yeah, that sounds like a bug. :)

delijah:
Hey mr. hellendag, thank you for your response!

> It creates an extra empty block before or after the inserted block
> > This is not ideal, but is present largely due to unresolved selection issues around atomic blocks, especially if the inserted block is the first or last block in the editor. If this can be eliminated, I'm open to it.

I think to fix the selection issues, it takes a lot of thinking and a lot of coding. So i would prefer to look at it separately. Then i would suggest the following behaviour for the moment:
- If the atomic block is inserted at the beginning of the editor, insert an empty block before
- If the atomic block is inserted at the end of the editor, insert an empty block after
- If the atomic block is inserted somewhere in the content / inbetween blocks, do not insert an empty block before and/or after

Or do you think it would make more sense to have a look at selection issues right away?

> It is mandatory to create an entity for an atomic block
> > This may be avoidable with the change to add block metadata, but for now, I'm not sure how you'd represent the features needed for your block without an entity. What do you propose there?

The thing is i am using block metadata already in my project, did my own (poor) implementation of it :) This is why i came across this issue. But i think in general, the draft-js core can not know if an entity for an atomic block is needed. I think it is wrong to make it mandatory by default. Just let the the caller make the decision if an entity should be created/applied or not.

Let me know what you think.

lacker:
Hey, thanks for submitting this pull request! It seems like this is not quite the ideal way to handle it, and the pull request has stalled, so I am going to close it. Thanks again.

Issue #395, opened 2016-05-14
is the `paragraph` block type removed now?


ianstormtaylor:
I have been using the `paragraph` block type instead of unstyled, but in upgrading to the lastest version from `0.4.0` (aka somewhere in the recent releases) it seems like the `paragraph` concept for blocks was removed? And in it's place `unstyled` is the only one?

Just wanted to double check if this is true. Might be good to add a breakage note to the release notes if they are retroactively editable.

Thanks!

@mitermayer you might be able to answer this one since it looks like you were creating a bunch of PRs around it.

mitermayer:
paragraph support is still present but it needs to be done using the blockRenderMap, apologies for not having the documentation yet on the website.  The documentation for it is open as a pull-request, but you can read it from here https://github.com/facebook/draft-js/pull/387/files

will paste it here so that you can see

```
const blockRenderMap = Immutable.Map({
 	'paragraph': {
 		element: 'p'
 	},
 	'unstyled': {
 		element: 'p'
 	}
 });
 
 // Include 'paragraph' as a valid block and updated the unstyled element but
// keep support for other draft default block types
const extendedBlockRenderMap = Draft.DefaultDraftBlockRenderMap.merge(blockRenderMap);

class RichEditor() {
    ...
    render() {
      return (<Editor ...
                blockRenderMap={extendedBlockRenderMap} />);
    }
}
```

ianstormtaylor:
Why have the paragraph supported internally but then make it not work until adding it to the `blockRenderMap`? Wouldn't it be better if it's considered a "core" type to just support it fully? Or if it's not a core type to remove it fully?

gimmixAT:
Hi, I am currently having a similar problem. Are you sure the overriding paragraph & unstyled with 'p' works? Because the resulting DOM, at least for me, is invalid (note the div in the p):

```
<div data-contents="true" data-reactid=".0.0.0.2.0.$editor0.0.0">
   <p class="" data-block="true" data-editor="1r4ao" data-offset-key="et022-0-0" data-reactid=".0.0.0.2.0.$editor0.0.0.$et022">
      <div data-offset-key="et022-0-0" class="public-DraftStyleDefault-block public-DraftStyleDefault-ltr" data-reactid=".0.0.0.2.0.$editor0.0.0.$et022.$et022">
[...]

PR #488, opened 2016-06-25
Some blocks should be treated as unstyled even when not defined on the blockRenderMap

mitermayer:
- This will guarantee if user does not define paragraphs will still split them into different blocks

This fixes https://github.com/facebook/draft-js/issues/395

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -132,11 +132,17 @@ function getListBlockType(
 function getBlockMapSupportedTags(
   blockRenderMap: DraftBlockRenderMap
 ): Array<string> {
+  // Some blocks must be treated as unstyled when not present on the blockRenderMap
   const unstyledElement = blockRenderMap.get('unstyled').element;
-  return blockRenderMap
+  const defaultUnstyledSet = new Immutable.Set(['p']);
+  const userDefinedSupportedBlockSet = (
+    blockRenderMap
     .map((config) => config.element)
     .valueSeq()
     .toSet()
+  );
+
+  return defaultUnstyledSet.merge(userDefinedSupportedBlockSet)
     .filter((tag) => tag && tag !== unstyledElement)
     .toArray()
     .sort();


src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -132,11 +132,17 @@ function getListBlockType(
 function getBlockMapSupportedTags(
   blockRenderMap: DraftBlockRenderMap
 ): Array<string> {
+  // Some blocks must be treated as unstyled when not present on the blockRenderMap
   const unstyledElement = blockRenderMap.get('unstyled').element;
-  return blockRenderMap
+  const defaultUnstyledSet = new Immutable.Set(['p']);
```
mitermayer: We need this since even though we don't support paragrah directly internally we still want to treated pasted paragraph content as different unstyled blocks

hellendag: Would this be better resolved by allowing the configuration to specify the set of elements that should map *to* a given block type, not just block type to element? I.e. `unstyled: { element: ..., wrapper: ..., inputTags: ['div', 'p']}`

mitermayer: Yeah, agreed I think that's the best option. That makes it one step closer to a schema, could also supply a function. In some cases custom elements will only match a particular type of html elements. Maybe a user will have a different block type for certain divs with different classes. 

I can update this PR to cover this.  What do you reckon ?

howardjing: Are there any updates on this? I can take a stab at implementing this as well if you don't have time @mitermayer.

mitermayer:
Hi Howardjing, please feel free to finalize this PR, I will only be able to work on it after 15th of October.

adamlong5:
@howardjing in case you missed the comment above from @mitermayer ^^

howardjing:
@adamlong5 thanks -- the PR is here: https://github.com/facebook/draft-js/pull/584

src/model/encoding/convertFromHTMLToContentBlocks.js
mitermayer: hi @howardjing please feel free to pick this up, i am unable to contribute to Draft until the 14th of October when i will resume work in here :)

howardjing: Hi @mitermayer the PR is here: https://github.com/facebook/draft-js/pull/584 -- everything works correctly when I play around with it locally. If you know who I can talk to to get the PR reviewed, that would be great. Thanks!

mitermayer:
will close this PR in favour or https://github.com/facebook/draft-js/pull/584

PR #502, opened 2016-06-30
Moving AtomicBlocks

delijah:
To be able to drag and drop atomic blocks around within the editor, it is necessary to have a way to move them within the blockMap. To achieve this goal, this pull request implements two methods in AtomicBlockUtils: moveAtomicBlockBefore and moveAtomicBlockAfter.

I was not completely sure how to name the new EditorChangeType for this action. Feel free to find some better names...

Would be great to see this feature in the master!

Cheers, delijah


docs/APIReference-AtomicBlockUtils.md
```diff
@@ -24,3 +24,21 @@ insertAtomicBlock: function(
...
+```
+
+### moveAtomicBlockAfter
+
+```
+moveAtomicBlockAfter: function(
+  editorState: EditorState,
+  contentBlock: ContentBlock
+): EditorState
+```
```
hellendag: These functions are a bit confusing. Am I moving the specified `ContentBlock` or am I moving something before/after the specified `ContentBlock`?

I think the best thing for API clarity is to specify which block is being moved and which block is the target to move it before/after.

src/model/immutable/EditorChangeType.js
```diff
@@ -20,6 +20,7 @@ export type EditorChangeType = (
   'backspace-character' |
   'change-block-type' |
   'change-inline-style' |
+  'change-fragment' |
```
hellendag: This is also confusing. What does it mean to "change" a fragment? I think it's probably fine to have a `'move-block'` change type.

src/model/modifier/AtomicBlockUtils.js
```diff
@@ -84,6 +86,48 @@ const AtomicBlockUtils = {
...
+
+    const withMovedAtomicBlock = moveBlockBeforeInContentState(
+      contentState,
+      selectionState,
+      contentBlock
+    );
+
+    const newContent = withMovedAtomicBlock.merge({
+      selectionBefore: selectionState,
+      selectionAfter: withMovedAtomicBlock.getSelectionAfter().set('hasFocus', true),
```
hellendag: Is this necessary? I see that it's being done in the existing method, just wondering if it's needed there or here.

docs/APIReference-AtomicBlockUtils.md
hellendag: Ah, I think I understand -- this is essentially following the pattern of `insertAtomicBlock`, assuming that the `SelectionState` specifies the block to move. I guess I find myself not liking that either. :)

src/model/transaction/moveBlockBeforeInContentState.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule moveBlockBeforeInContentState
```
hellendag: Looks like a lot of similarity in code between this module and the one above. Can they be combined?

src/model/modifier/__tests__/AtomicBlockUtils-test.js
```diff
@@ -252,5 +369,167 @@ describe('AtomicBlockUtils', () => {
...
+      // Try to move atomic block to the bottom
+      expect(function() {
+        moveAtomicBlockAfter(
+          atomicEditor,
+          lastBlock
+        );
+      }).toThrow(
+        new Error('Selection range must be within same block.')
+      );
+    });
```
hellendag: Thanks for the extensive test specs!

docs/APIReference-AtomicBlockUtils.md
delijah: Yes exactly, i've tried to keep the same style for AtomicBlock functions and took insertAtomicBlock function as basis. But i absolutely agree with you, the function declarations are not very intuitive at the moment. So i've been thinking about a better way, also considered other Draft.js functions (like Modifier.moveText).

Possible solution:
-- 

```
moveAtomicBlock: function(
  editorState: EditorState,
  atomicBlock: ContentBlock,
  targetRange: SelectionState
): EditorState
```

Add a function "moveAtomicBlock". With this function you can move atomic blocks before, after or in between existing blocks.
This function does not use the actual selection as a source for the move-operation, but uses the atomicBlock, passed as a second argument, as a source for the move-operation. The atomicBlock will be moved to the targetRange (third argument). Rules:
- If targetRange.anchorOffset is 0, move the block before the block mentioned in the target range.
- If the targetRange.focusOffset is set to the end of the target block, move the atomic block after the block mentioned in the targetRange.
- If the targetRange is collapsed and targetRange.anchorOffset and targetRange.focusOffset is somewhere within the target block, split the target block and move the atomicBlock in between those.
- If the targetRange isn't collapsed replace the selection with the atomicBlock. This can also work if the targetRange includes more than one block.

Since it is not possible to determine if the atomic block should be moved before or after an empty block, we also need the functions "moveAtomicBlockBefore" and "moveAtomicBlockAfter"...

```
moveAtomicBlockBefore: function(
  editorState: EditorState,
  atomicBlock: ContentBlock,
  targetBlock: ContentBlock
): EditorState
```

Adjusted the function "moveAtomicBlockBefore". With this function you can move atomic blocks (second argument) before another block (third argument).

```
moveAtomicBlockAfter: function(
  editorState: EditorState,
  atomicBlock: ContentBlock,
  targetBlock: ContentBlock
): EditorState
```

Adjusted the function "moveAtomicBlockAfter". With this function you can move atomic blocks (second argument) after another block (third argument).

src/model/immutable/EditorChangeType.js
delijah: Agree on that. `'move-block'` makes totally sense to me.

src/model/modifier/AtomicBlockUtils.js
delijah: Not sure if i understand what's on your mind :)

src/model/modifier/__tests__/AtomicBlockUtils-test.js
delijah: Welcome :)

src/model/transaction/moveBlockBeforeInContentState.js
delijah: Bueno, i'll give it a try!

docs/APIReference-AtomicBlockUtils.md
delijah: Or we forget about the before and after functions and implement a fourth (optional) argument insertionDirection to the moveAtomicBlock function, which overrides the algorithm which decides if the atomic block should be inserted before or after the targetRange.

```
type DraftInsertionDirection = 'before' | 'after';

moveAtomicBlock: function(
  editorState: EditorState,
  atomicBlock: ContentBlock,
  targetRange: SelectionState,
  insertionDireciton?: DraftInsertionDirection
): EditorState
```

delijah: Start to like that last version more and more. My fingertips can't wait to implement this! Haha.

hellendag: I like this too.

hellendag:
I'd say move forward with the `DraftInsertionMode` approach -- I think it sounds like a good way to do this. :)

delijah:
Hey!!! I've just rebased this branch with master, obviously github considers this as closed since i had to remove the remote to push it again :( I guess i have to open another pull request...

PR #508, opened 2016-07-04
removes window.prompt from media example

### LINKED ISSUE #463: Bug in example
Hi, 

I just found a bug in one example:

https://github.com/facebook/draft-js/blob/master/examples/media/media.html

If you click the button "add image" and don't type anything, this.state.editorState will be set to null because the lines: 

 _addMedia(type) {
          const src = window.prompt('Enter a URL');
          if (!src) {
            return;
          }

          const entityKey = Entity.create(type, 'IMMUTABLE', {src});

          return AtomicBlockUtils.insertAtomicBlock(
            this.state.editorState,
            entityKey,
            ' '
          );
        }

will return null if src is not set (specifically if(!src) line). Then afterwards you can't upload any images since the whole thing is broken, because this.state.editorState is set to null already. 

I 'd change

to:
      if (!src) {
            return this.state.editorState;
        }

I can submit a PR if this is not my illusion

LordLambda:
fixes #463

this fixes two things:
1. it fixes #463 where if you inserted a blank string for a piece
of media the editor state would be set to null.
2. the media example no longer uses window.prompt which previously
broke selection state. instead using an input like the link example.

it should be noted though that with this commit it's possible to
insert a blank url for media. i initially was going to add protection
for this, but after looking at the link example it appears the link
example doesn't care if you enter a blank link.

looking at the media objects generated for mp3/mp4 the play bar is still
there, and image is a whitespace. i'll happily add in a safeguard if
it's determined that's "unwanted" behavior. however I decided to stick
with consistency with the other examples for this commit.

### DIFF
--- a/examples/media/media.html
+++ b/examples/media/media.html
@@ -43,20 +43,25 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
           this.state = {
             editorState: EditorState.createEmpty(),
+            showURLInput: false,
+            url: '',
+            urlType: '',
           };
 
           this.focus = () => this.refs.editor.focus();
-          this.onChange = (editorState) => this.setState({editorState});
           this.logState = () => {
             const content = this.state.editorState.getCurrentContent();
             console.log(convertToRaw(content));
           };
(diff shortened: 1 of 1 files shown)


hellendag:
Looks good. Thanks!

Issue #430, opened 2016-05-31
Need ability to "not continue" an inline style or mutable entity


guitardave24:
Right now, the editor simply looks to the left of the cursor to figure out what the style should be when inserting text (https://github.com/facebook/draft-js/blob/master/src/model/immutable/EditorState.js#L597). This works for "continuing" bold/italic/etc which is usually what you want. But sometimes you might not. It also uses the similar logic for mutable entities.

This might be complicated to describe, so I will explain with a use case: Imagine you want to implement inline code `like this`. So when you type a tick, some text, and another tick, it converts 'foo' into `foo`, by replacing the text and applying an inline style. However, as you continue typing, it will extend that code snippet because the cursor is picking up the inline style.

There are workarounds like: using immutable/segmented entities (mutable also exhibit this behavior) or adding an extra space when replacing the text and applying the inline code. But, really, what I want is the ability to edit "inside" of the snippet, but not have it continue at the boundaries.

I currently have a hacky workaround which handles `onBeforeInput` globally for certain inline styles/entities and manually inserts text when it borders on one of these types of inline styles.

Not sure if this is a widespread, known or otherwise specific use case, but worth logging.

spicyj:
I haven't looked at exactly how it's implemented, but if you go to the example on the homepage https://facebook.github.io/draft-js/, click Bold, type some text, click Bold again, then type more, the second piece of text is not bold. Is that what you want?

guitardave24:
Hm, that is a good point. I suppose applying a style to the collapsed, new range may solve it, though I am certain I tried that. Let me see. Thanks @spicyj

guitardave24:
Ah, yes. It uses toggleInlineStyle which uses `setInlineStyleOverride` (https://facebook.github.io/draft-js/docs/api-reference-editor-state.html#setinlinestyleoverride).

I've switched to using this as it's exposed. Thanks for the tip.

guitardave24:
Actually, re-opening because I think it's still valid.

In the above case, I actually *never* want it to continue from the edge. Otherwise, you end up with weird edge cases such as being unable to get out of it without some hotkey or UI affordance. I can solve the case for continuing to type after the transformation, but once I move the cursor left and right again (for example), it picks up the style and you can't get out of it.

guitardave24:
Here's an example:
![bugs](https://cloud.githubusercontent.com/assets/1735553/15694664/90b97244-2754-11e6-9f96-b10ff99e3e19.gif)

[...]

PR #510, opened 2016-07-05
Need ability to "not continue" a mutable entity #430

iandoe:
Sometimes, we do not want MUTABLE entities to be continued when they are on the left of
the cursor, for example when creating a LINK entity that we do not want to continue after
pressing the space key, this proposal adds a "Contiguous" setting set to true by default that
we can set to false to get the desired behavior

Please note this is a WIP and not yet ready to be merged
- [x] Rework example for link.html
- [x] Rewrite tests with selection states
- [ ] Figure out why exported state does not output contiguous prop

### DIFF
--- a/examples/link/link.html
+++ b/examples/link/link.html
@@ -87,7 +87,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
         _confirmLink(e) {
           e.preventDefault();
           const {editorState, urlValue} = this.state;
-          const entityKey = Entity.create('LINK', 'MUTABLE', {url: urlValue});
+          const entityKey = Entity.create('LINK', 'MUTABLE', {url: urlValue}, false);
           this.setState({
             editorState: RichUtils.toggleLink(
               editorState,
--- a/src/model/entity/DraftEntity.js
+++ b/src/model/entity/DraftEntity.js
@@ -52,7 +52,7 @@ var DraftEntity = {
     data?: Object
   ): string {
     return DraftEntity.add(
-      new DraftEntityInstance({type, mutability, contiguous: true, data: data || {}})
+      new DraftEntityInstance({type, mutability, contiguous: contiguous || true, data: data || {}})
     );
   },
 
--- a/src/model/entity/DraftEntity.js
+++ b/src/model/entity/DraftEntity.js
@@ -51,8 +51,9 @@ var DraftEntity = {
     data?: Object,
     contiguous?: bool,
   ): string {
+    var isContiguous = (typeof contiguous !== undefined) ? contiguous : true;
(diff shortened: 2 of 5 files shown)


facebook-github-bot-1:

facebook-github-bot-9:

iandoe:
The word `contiguous` might not be the best, i am open to suggestions on a better term.

amireh:
This is great, thank you! I will test this in our editor as soon as I get the chance and perhaps help with the remaining points.

FWIW, I thought "contiguous" to be very descriptive but then I knew what kind of problem this was solving so that opinion may be skewed.

spicyj:
I don't think the word "contiguous" is intuitive here but am at a loss for better ideas. I wonder if it makes sense to always have entities behave like this? Not sure if there are use cases when you would want an "expand-y" entity.

iandoe:
@spicyj I think entities like **Bold** and _Italic_ for example are expected to expand as it is the most common UX so far in wysisyg editors.

hellendag:
Well, **bold** and _italic_ are handled internally as inline styles and do not require entities. There has been some previous discussion about whether they should actually be treated as distinct from decorators at all, though certainly they shouldn't require entities.

Is this not manageable by using a `handleBeforeInput` with the space character? If a space char should be inserted, can you manually perform the insertion, without the entity applied to the new character?

Stylistically, btw, I am *strongly* anti-boolean-arguments. :)

aligon:
Are the plans to merge this pull request?

I have a use case that "non-contiguous" entities would solve. There are some other work arounds that I found in https://github.com/facebook/draft-js/issues/430. For now I'll implement one of those. It'd be nice to know if this will ever be merged or not.

flarnie:
I'd rather not add a boolean option argument, and I'm also hesitant to add any more API surface to Draft. For now going to close this, but long term we probably will consider an API redesign that takes this into account.

javierfernandes:
Is there any other way to accomplish this since the PR was not merged ?
I'm also one of those that need this feature. For cases that you don't want the inlineStyle to be continued it is very annoying for the end user.

thibaudcolas:
True, that's annoying! I agree it would be better if Draft.js supported this directly since this is quite a common use case, but in the past the framework has been quite slow to evolve. Hence the many workarounds in #430. 

For the behavior you're showing, yes I wouldn't be surprised if the current inline style was wrong (when inline styles aren't continued), but I would also suspect that the cursor's color is handled at the browser level, and the browser simply has the selection placed inside the `span` that has those styles. I don't think there is a way to work around this unfortunately.

If I was to implement this, I think I would try to use the same font color everywhere to work around the issue, or replicate how Dropbox Paper allows continuing styles but doesn't do it when entering styles with Markdown-style shortcuts:

paper-shortcuts

PR #514, opened 2016-07-06
Update Flow comment supression regex

zpao:
**Summary**

I'm not sure why we need to change this. The magic bit seems to be needing to match the newline, the rest of the changes just bring use more in line with the internal flowconfig. Maybe @jeffmo can tell us why?

**Test Plan**

`npm run flow`

### DIFF
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -16,7 +16,7 @@
 module.system=haste
 esproposal.class_static_fields=enable
 suppress_type=$FlowIssue
-suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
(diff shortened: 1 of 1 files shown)


spicyj:
Maybe without the newline, the comment needs to be on the same line as the offending code?

zpao:
Seems to work fine internally without the newline. ¯\_(ツ)_/¯

jeffmo:
Pretty weird... I'm not sure what would've changed here :/

zpao:
Going to land this since it gets us back to green. We can revisit how we want that suppression to look and try to figure out the issue later.

PR #517, opened 2016-07-06
Update fbjs dep to 0.8.3

zpao:
This should ensure that Flow will pass and nobody should get stuck with
a version of fbjs that isn't compatible with the version of Flow we use.


spicyj:
k

zpao:
Only remaining Flow failures will be gone after #514

PR #518, opened 2016-07-06
Update Travis CI config

zpao:
**Summary**

- reorder for better readability
- add the Node v4 build requirements
- start pruning and updating npm deps so the cache doesn't get stale

**Test Plan**

- `travis-lint` only reports issue with `apt` key but is a bug with their tool
- testing it live on Travis here now


hellendag:
makeitso

Issue #504, opened 2016-07-01
option to clone entities on paste


afraser:
I just noticed that if I paste a block with custom entities in it, the newly created block has entities with the same key.

It looks like these might be the relevant places:
https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnPaste.js#L59
https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnPaste.js#L164

Maybe this could be handled with new `DraftEntityMutability` values?

eg.
MUTABLE_UNIQUE
IMMUTABLE_UNIQUE
SEGMENTED_UNIQUE

Thoughts?

hellendag:
Good question. This is in the gray area between feature and bug. :) In some scenarios, one might want the same entity to apply to the pasted content, but this is a clear case where one would not.

One thing I have been thinking about is whether it makes sense to define a "paste config" object that can be passed as a prop to `Editor`. This would allow `stripPastedStyles` to have a proper home rather than being a first-class prop, would allow better flexibility for keeping/discarding certain styles, and would make something like this pretty trivial to configure -- e.g. `pasteUniqueEntities: true` to force duplication of all entity objects for the pasted output.

Has this only been an issue for you within paste behavior, or have you encountered any other trouble with duplicated entities?

afraser:
This has only been an issue with pasting. I can't really think of any other ways that draft might need to handle duplicating entities for me, but then again, I'm pretty new to using it.

A similar case could me made for pasting media. A user who copies and pastes a media object into a different part of the document may intend to modify that object. e.g. pasting a video player embed and wanting to change the start/end time.

I like the idea of a `pasteConfig` it solves my problem more elegantly and makes a home for other paste configurations. What about the other paste handlers: `handlePastedText` and `handlePastedFiles` ...would it make sense to move them in as well?

afraser:
@hellendag I've forked the repo and am beginning to work on this. 

Would this likely be accepted in a PR?

I addition to `pasteUniqueEntities`, I'm going to be doing some additional work to fix pasting of inline equation entities. Here's my strategy

1. Ensure 'internal paste' is detected for selections within a block that contain an Entity
    * Modify Draft Entity id scheme so we can detect when an Entity is pasted based off of it's DOM id
    * If Draft Entity ids are found in the pasted HTML, use the internal clipboard.
[...]

Issue #390, opened 2016-05-13
How to create custom inline components without text


inyono:
I'm currently stuck at creating custom inline components. I want to have a component for inline LaTeX equations. I tried to follow [Creating and Retrieving Entities](https://facebook.github.io/draft-js/docs/advanced-topics-entities.html#creating-and-retrieving-entities) but I'm not sure what to pass `Modifier.applyEntity`. Something like this didn't work as intended:

``` .js
addInlineLatexEquation() {
  const { editorState } = this.state

  const entityKey = Entity.create(
    'INLINE_LATEX_EQUATION',
    'IMMUTABLE',
     { src: '' }
  )

  const contentState = editorState.getCurrentContent()
  const selectionState = editorState.getSelection()

  const newContent = Modifier.applyEntity(
    contentState,
    selectionState,
    entityKey
  )

  const newEditorState = EditorState.push(editorState, newContent, 'apply-entity')

  this.setState({
    editorState: newEditorState
  })
}
```

I now can create an inline LaTeX component by selecting text, but I rather want to have a behavior similar to the block components. I.e., I want to add an inline component at the current cursor position without any text (basically `AtomicBlockUtils.insertAtomicBlock` for inline components).

mull:
Just going to share the snippet I shared on Slack. This

1. Inserts an empty space before the current selection
2. Adds a blank character with an entity (which I then render customly through a decorator)
3. Adds a blank character afterwards

```js
import { Entity, Modifier, EditorState } from 'draft-js';

export function insert(editorState) {
  const currentContent = editorState.getCurrentContent();
  const entity = Entity.create(
    'TEX',
    'IMMUTABLE',
    {content: "$ 1 + 1 $"}
  );

  const selection = editorState.getSelection();

  const firstBlank = Modifier.insertText(
    currentContent,
    selection,
    " ",
    null,
    null
  );

  const withEntity = Modifier.insertText(
    firstBlank,
    selection,
    " ",
    null,
    entity
  );

  const withBlank = Modifier.insertText(
    withEntity,
    selection,
    " ",
    null,
    null,
  );

  return EditorState.push(
    editorState,
    withBlank,
    'insert-text'
  );
}
```

jussch:
@mull Thanks for the example snippet.

[...]

PR #525, opened 2016-07-08
add pasteUniqueEntities prop to DraftEditor

afraser:
**Summary**

Fixes: https://github.com/facebook/draft-js/issues/504 and https://github.com/facebook/draft-js/issues/390

I'd love to get some feedback on what I can do to get this accepted.

Some use cases for this:
- equations (implemented here)
- chemical diagrams
- sparklines
- cloze sentences (i.e. fill in the blanks) where the blanks have potential correct solutions
- contextual notes ala google docs

TODO:
- [ ] Distill the equation example.

**Test Plan**
- [x] ensure entities are not cloned when `pasteUniqueEntities={false}`
- [x] ensure entities are cloned when `pasteUniqueEntities={true}`
- [x] ensure entity references are pasted in the right place
- [ ] ensure that pasting plain text (non-internal paste) works as expected
- [ ] ensure that pasting plain HTML (non-internal paste) works as expected


facebook-github-bot-9:

facebook-github-bot-5:

afraser:
Another solution I think I might try instead of the `draftEntityKeyPrefix` solution is to leave it up to the user to implement the entity decorator such that it renders the editorKey on the DOM somewhere if they want `pasteUniqueEntitiies` for entities that are rendered with content that isn't the same as what's in the plain text. This way there's no need for this `else`: https://github.com/facebook/draft-js/pull/525/files#diff-3e4f7f7129a394d16c3592dd894fd31eR135

afraser:
Update: I now use `editorKey` instead of `draftEntityPrefix` to detect internal pasting of entities in selections that are within a block. This requires the user to include the editorKey somewhere in the markup of their component.

ericbiewener:
The example doesn't work well for me. If I move the carat with my arrow keys, as soon as I reach the equation it jumps back to the beginning.

flarnie:
Hi - first of all, sorry that this PR got stale with no review. Looking at it not, I would be open to merging something similar but with a different approach.

Rather than adding a prop, should we just change the default behavior here? When would you ever want a copy-pasted section to refer to the same entities as in the original section?

I would need to verify that none of Facebook's use cases rely on the current behavior, but I doubt they do.

So for now closing this PR but if @afraser or others are still interested in fixing this please open a new issue/PR.  Thanks!

afraser:
@flarnie Thanks for getting back to me on this. I've long since moved on from the project in question, but I totally get what you're saying about changing the approach.

PR #530, opened 2016-07-11
Use block level metadata into fragment insertion

vierno:
**Summary**

Fixes a situation when you insert new blocks with metadata,
eg using Modifier.replaceWithFragment. If the fragment have data,
it wouldn't be present on the resulting ContentState.

insertFragmentIntoContentState now uses the metadata of the
block or blocks inserted.

Also, included tests for insertFragmentIntoContentState.

**Test Plan**

It's possible to observe the results of Modifier.replaceWithFragment when the fragment have block-level-metadata, before the fix, the resulting ContentState will not contain the metadata from the new blocks, and after the fix, the metadata will be there.

### DIFF
--- a/src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
+++ b/src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @emails oncall+ui_infra
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
+var Immutable = require('immutable');
+var insertFragmentIntoContentState = require('insertFragmentIntoContentState');
+var getSampleStateForTesting = require('getSampleStateForTesting');
+var BlockMapBuilder = require('BlockMapBuilder');
+var ContentBlock = require('ContentBlock');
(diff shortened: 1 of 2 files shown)


src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
```diff
@@ -0,0 +1,143 @@
...
+    return BlockMapBuilder.createFromArray(fragmentArray);
+  }
+
+  function createMultiblockFragment() {
+    var fragmentArray = [
+      new ContentBlock({
+        key: 'j',
+        type: 'unstyled',
+        text: 'xx',
+        characterList: Immutable.List(),
```
hellendag: Nitpicking, and this should probably not affect anything in the test, but `characterList` should be populated by `text.length` number of `CharacterMetadata` objects. `CharacterMetadata.EMPTY` would do the trick.

src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
```diff
@@ -0,0 +1,143 @@
...
+    return BlockMapBuilder.createFromArray(fragmentArray);
+  }
+
+  it('must throw if no fragment is provided', () => {
+    var fragment = BlockMapBuilder.createFromArray([]);
+    expect(() => {
+      insertFragmentIntoContentState(
+        content,
+        selection,
+        fragment);
```
hellendag: Styling nit for the closing paren:

```javascript
someFunction(
   lorem,
   ipsum,
   dolorsit
);

hellendag:
Thanks for the test!

I'll go ahead and merge, but I left a couple nitpick comments that you can tweak in a followup if you'd like.

PR #533, opened 2016-07-12
Accept pasted mailto: links

mjomble:
**Summary**

`mailto:` links were being stripped to plaintext on paste because they failed the protocol check.

For now I added it to the list of accepted protocols, but I was wondering if it might be reasonable to remove this check altogether and allow pasting links with any protocol? Or perhaps making this configurable.

**Test Plan**

There are no tests for `convertFromHTMLToContentBlocks` yet. I'd prefer to add cases to an existing test file rather than starting a new one, unless absolutely necessary.

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -286,7 +286,7 @@ function hasValidLinkText(link: Node): boolean {
     'Link must be an HTMLAnchorElement.'
   );
   var protocol = link.protocol;
-  return protocol === 'http:' || protocol === 'https:';
+  return protocol === 'http:' || protocol === 'https:' || protocol === 'mailto:';
 }
 
 function genFragment(


hellendag:
Seems reasonable enough to me, thanks.

There is a test for `DraftPasteProcessor`, at https://github.com/facebook/draft-js/blob/master/src/model/paste/__tests__/DraftPasteProcessor-test.js. The test should probably just be renamed and moved entirely, since all important logic was moved to `convertFromHTMLToContentBlocks`. cc @clowbrow

If you wouldn't mind adding a test case there in a followup, that would be great. :)

mjomble:
I'll look into adding the test case soon.

Also, there is an unanswered question here :)

> I was wondering if it might be reasonable to remove this check altogether and allow pasting links with any protocol? Or perhaps making this configurable.

Was there a specific reason why this check was added? I noticed it goes back to the first revision on GitHub.

spicyj:
We should definitely not allow javascript: or vbscript: or data: by default as they allow XSS.

mjomble:
Aaah, yes. Excellent point :)

PR #535, opened 2016-07-12
ESLint upgrade, use new fb lint config, fixes

zpao:
## Summary

This doesn't do the strict enforcement of lint, but it does get us back to zero warnings.

I'm in the process of shipping a standalone config that is (virtually) identical to our internal config, minus custom rules. There isn't much impact for Draft except that we can remove the config parts that were duplicated, leaving us with just a simple extends entry.

There are a few code changes to make the `consistent-return` rule pass. One case I just disabled the rule (it doesn't understand that `invariant` throws). Then a couple explicit `return undefined`s. The last was in a `[].some` call. The implicit `undefined` is treated the same as `false` so I used that more broadly. I can add comments if it'll help with the code intent (which I didn't look super closely at).

## Test Plan

- `npm link` from `eslint-config-fbjs` and run lint. Modified a Flow Object Type to use a semicolon and saw that it failed lint for the `babel/flow-object-type` rule (only specified in the new config).
- `npm test`
- `npm run flow`

## Dependencies

- [ ] https://github.com/facebook/fbjs/pull/160


src/component/utils/getDefaultKeyBinding.js
```diff
@@ -45,6 +45,7 @@ function getZCommand(
   if (hasCommandModifier(e)) {
     return e.shiftKey ? 'redo' : 'undo';
   }
+  return undefined;
```
spicyj: Can you make this `null` like the examples below?

src/component/selection/getRangeClientRects.js
```diff
@@ -55,6 +56,7 @@ function getRangeClientRectsChrome(range: Range): Array<ClientRect> {
     'Found an unexpected detached subtree when getting range client rects.'
   );
 }
+/* eslint-enable consistent-return */
```
spicyj: Is this because of the invariant at the end? Should we just add a `return [];` at the bottom instead?

src/component/handlers/edit/getFragmentFromSelection.js
```diff
@@ -19,12 +19,15 @@ import type EditorState from 'EditorState';
...
 function getFragmentFromSelection(editorState: EditorState): ?BlockMap {
   var selectionState = editorState.getSelection();
-  if (!selectionState.isCollapsed()) {
-    return getContentStateFragment(
-      editorState.getCurrentContent(),
-      selectionState
-    );
+
+  if (selectionState.isCollapsed()) {
+    return undefined;
```
spicyj: Can we make this `null` too? It looks like we only check truthiness of this value before indexing into it.

spicyj:
lgtm

.eslintignore
```diff
@@ -1,6 +1,5 @@
 lib/
 dist/
 docs/
-examples/
```
zpao: I think I'll just strip this commit and do separately (or make an issue for somebody new to contribute). That example is actually doing more and running eslint in webpack. I don't see a good reason to do that with top-level linting but it's already the case so meh.

src/component/selection/getRangeClientRects.js
zpao: Yup (see PR description 😛).

I just threw the `[]` in there & Flow complains that it's unreachable code because it does know about `invariant`.

I'll leave it as is.

src/component/utils/getDefaultKeyBinding.js
zpao: 👍 Going to match the style of the fn below this too (early return `null`)

PR #545, opened 2016-07-18
fixes selection-state with -webkit-user-modify

SecurityInsanity:
**Summary**
turns out chrome (at least) would incorrectly not erase the placeholders
with -webkit-user-modify being applied where it was.
this was totally an oversight on my part not testing after taking
the lucidity designs comment in PR 509.

Testing this time in my normal suite of browsers (firefox, chrome, safari)
this commit moving the selection state to just target .public/DraftEditor/content
fixes the selection state problem.

**Test-Plan**
1. Build the project.
2. Open up the rich content example.
3. Notice the placeholder text goes away when you type.

### DIFF
--- a/src/component/base/DraftEditor.css
+++ b/src/component/base/DraftEditor.css
@@ -11,6 +11,9 @@
 .public/DraftEditor/content {
   height: inherit;
   text-align: initial;
+}
+
+.public/DraftEditor/content {
   -webkit-user-modify: read-write-plaintext-only;
   user-modify: read-write-plaintext-only;
 }


jvaill:
Hey, I also noticed that the placeholder is not correctly erased, but I'm wondering if the problem stems from something else. I haven't dug in too much, but `onChange` isn't firing at all for me in Chrome with the editor on `https://facebook.github.io/draft-js/` or the rich text example. When I click one of the formatting buttons, the placeholder correctly goes away - just fyi!

SecurityInsanity:
Hey @jvaill is that incorrect behavior happening with this pull request checked out?

If your testing on master/on the draft-js website webkit-user-modify caused lots of weird edge case behavior.

spicyj:
No idea why this works, but I'll take it.

jvaill:
@SecurityInsanity Just tried it, you're right that this fixes all of the issues! Apologies for sounding the alarm too soon. Nice fix. 👍

SecurityInsanity:
@spicyj after an hour of testing this after I determined this is why it worked:

magic

_No but really_, looking into it @jvaill 's minor digging seems correct. Chrome doesn't detect the "rich" content area as changing. Maybe it's because the flag "plaintext-only" only causes it to fire events for those when applied to a node? I still don't fully understand it, but the root cause seems to be "-webkit-user-modify" causing `onChange` to not fire.

PR #551, opened 2016-07-21
Moving atomic blocks

delijah:
To be able to drag and drop atomic blocks around within the editor, it is necessary to have a way to move them within the blockMap. To achieve this goal, this pull request adds the function moveAtomicBlock to AtomicBlockUtils.

This is a copy of #502 since i've mistakenly deleted the pull request branch of that one. Sorry.


src/model/transaction/moveBlockInContentState.js
```diff
@@ -0,0 +1,75 @@
...
+  insertionMode: DraftInsertionType
+): ContentState {
+  invariant(
+    blockToBeMoved.getKey() !== targetBlock.getKey(),
+    'Block cannot be moved next to itself.'
+  );
+
+  var blockMap = contentState.getBlockMap();
+  var blockBefore = blockMap.toSeq().takeUntil(v => v === targetBlock).last();
+  var blockAfter = blockMap.toSeq().skipUntil(v => v === targetBlock).skip(1).first();
```
hellendag: const targetKey = targetBlock.getKey();
  const blockBefore = contentState.getBlockBefore(targetKey);
  const blockAfter = contentState.getBlockAfter(targetKey);

src/model/modifier/AtomicBlockUtils.js
```diff
@@ -84,6 +88,84 @@ const AtomicBlockUtils = {
...
+    const contentState = editorState.getCurrentContent();
+    const selectionState = editorState.getSelection();
+
+    var withMovedAtomicBlock;
+
+    if (insertionMode === 'before' || insertionMode === 'after') {
+      const targetBlock = contentState.getBlockForKey(
+        insertionMode === 'before' ?
+          targetRange.getAnchorKey() :
+          targetRange.getFocusKey()
```
hellendag: Use `getStartKey()` and `getEndKey()` for these instead, since we're looking at the before/after order of blocks irrespective of selection direction.

src/model/modifier/AtomicBlockUtils.js
```diff
@@ -84,6 +88,84 @@ const AtomicBlockUtils = {
...
+  moveAtomicBlock: function(
+    editorState: EditorState,
+    atomicBlock: ContentBlock,
+    targetRange: SelectionState,
+    insertionMode?: DraftInsertionType
+  ): EditorState {
+    const contentState = editorState.getCurrentContent();
+    const selectionState = editorState.getSelection();
+
+    var withMovedAtomicBlock;
```
hellendag: let

delijah:
Hey, any updates on this?

src/model/transaction/moveBlockInContentState.js
```diff
@@ -0,0 +1,81 @@
...
+): ContentState {
+  invariant(
+    blockToBeMoved.getKey() !== targetBlock.getKey(),
+    'Block cannot be moved next to itself.'
+  );
+
+  invariant(
+    insertionMode !== 'replace',
+    'Replacing blocks is not supported.'
+  );
```
hellendag: Why is it included in the enum?

src/model/transaction/moveBlockInContentState.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule moveBlockInContentState
+ * @typechecks
```
hellendag: This can be removed.

hellendag:
One question around the `DraftInsertionType` enum. Otherwise I think this is fine.

src/model/transaction/moveBlockInContentState.js
delijah: replace is needed in AtomicBlockUtils but not here. Actually i could just implement replace here as well...

delijah: Actually i am not really sure if i should implement the support for "replace" here, since at the moment there is no use case for it.

Actual use case:
- in "moveBlockInContentState" the insertionMode can be "before" or "after"
- in "AtomicBlockUtils" the insertionMode can be "before", "after" or "replace"

So i see 3 solutions:
1. implement "replace" in "moveBlockInContentState" even if there is no use case at the moment
2. implement 2 different types for "insertionMode", one to use in "moveBlockInContentState" and one to use in "AtomicBlockUtils"
3. we just leave it as it is at the moment

What do you think?

rodrigopivi:
Is there any chance this will get merged soon? dragging atomic blocks seem an important feature

src/model/transaction/moveBlockInContentState.js
delijah: Not sure if i understand correctly. You mean the whole comment section can be removed?

src/model/transaction/moveBlockInContentState.js
delijah: I suggest to just leave it as it is at the moment. It makes completely sense in my opinion.

jan4984:
I'm interesting in the working drag and drop demo, or complete d&d support in draft-js. Moving a block can be easily done with the OrderedMap APIs.

src/model/transaction/moveBlockInContentState.js
hellendag: I think I was referring to the `@typechecks` directive.

src/model/transaction/moveBlockInContentState.js
hellendag: That sounds fine.

hellendag:
Sorry for the extremely long wait. Let's go ahead with this. :)

PR #559, opened 2016-07-24
Added test cases for pasting mailto: and javascript: links

mjomble:
As discussed on #533, I've added a test for pasting mailto: links.

As I also found out from that discussion that the main reason for checking the protocol is to avoid javascript: links etc, I've also added a test case for that.
And corrected the indentation on a few lines I noticed.


hellendag:
Thanks!

PR #566, opened 2016-07-27
Expose path to Draft.css for usage by CSS pre-processors

darobin:
**Summary**

Some CSS pre-processors understand `node_modules` and are able to `@import` without requiring an absolute path. However, without assistance from the module all that they can resolve is its root directory. This leads to `@import` statements like the following:

``` css
@import "draft-js/dist/Draft.css";
```

While this works, it requires internal knowledge of the package's layout. This requires digging around and if the layout changes, this breaks.

These pre-processors also recognise a `style` field in `package.json`, which is the equivalent of `main` for CSS. This allows one to import thusly:

``` css
@import "draft-js";
```

It's a small change, but it helps!

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "PATENTS"
   ],
   "main": "lib/Draft.js",
+  "style": "dist/Draft.css",
   "repository": "facebook/draft-js",
   "license": "BSD-3-Clause",
   "scripts": {


zpao:
Interesting. What preprocessors work like this? (so we can test on our end and ensure we have some knowledge about what will and won't work)

darobin:
Typically anything built on top of https://github.com/postcss/postcss-import (of which there are [a few](https://www.npmjs.com/browse/depended/postcss-import)). If you want to test the simplest is probably to just use `postcss-import` directly.

hellendag:
I'm fine with this. @zpao, any concerns?

flarnie:
Not sure why CI never finished - I think we could merge this if it's not too late. Adding it to my queue.

flarnie:
PS - I might have trouble merging this because it's been open for so long. If @darobin you are able to rebase it onto master and force push, that should re-trigger CI and make it easier to merge.

darobin:
Hi — sorry, I no longer use draft-js or for that matter work in development. I don't have anything set up on this machine. It's a one-line change to `package.json` though so it shouldn't be too hard to bring it back :)

darobin:
Actually @flarnie, GitHub says it's mergeable without a hitch so you should be fine there. Just ignore CI (or kick it over in the Travis UI), this really shouldn't break tests.

flarnie:
Agreed - the only hitch is that our scripts for pulling changes into the Facebook codebase may try to verify that CI is passing on the branch. If so then I'll make a new branch.

Good luck with whatever you are doing now! :) 🌅

flarnie:
Internally I'm getting a merge conflict and unable to rebase it - our system expects that rebasing and conflict resolution will happen on Github, and I haven't been able to access this branch to rebase it.

I'll just make a new PR with this change, hope that's ok. Credit goes to @darobin for this change, and sorry again it went so long without review.

PR #571, opened 2016-07-29
Add p tags to block render map

### LINKED ISSUE #523: `convertFromHTML` merged <p> tags into one <p> tag
I'm having some problems with the `convertFromHTML`, it doesn't render the right data as I expected.

For example,
_html_: `<p>Paragraph 1</p>, <p>Paragraph 2</p>, <p>Paragraph 3</p>` 

_editorState_: 
`EditorState.createWithContent(ContentState.createFromBlockArray(convertFromHTML(_html_)), Decorator)`

The html at this stage render correctly as expected.
> Paragraph 1
> Paragraph 2
> Paragraph 3

However, when I save data and refresh the page, it displays as below (all in one line)

> Paragraph 1 Paragraph 2 Paragraph 3

Then, I clicked on the editor and suddenly the html turned into
`<p>Paragraph 1 Paragraph 2 Paragraph 3</p>`

Does anyone has similar problem like this? Thanks.

CLowbrow:
**Summary**

Fixes #523 

There was some kind of diversion with the test cases using their own block map. I took that out and put p in the block map and everything worked. @mitermayer seems to have dealt with a lot of this last so would be good to get his opinion. 

This will alter paste behavior for people but they can change it back by passing a custom block map. 

**Test Plan**

`Draft.convertFromHTML("<p>A</p><p>B</p><p>C</p>");` returns 3 blocks

### DIFF
--- a/src/model/immutable/DefaultDraftBlockRenderMap.js
+++ b/src/model/immutable/DefaultDraftBlockRenderMap.js
@@ -61,4 +61,7 @@ module.exports = Map({
   'unstyled': {
     element: 'div',
   },
+  'paragraph': {
+    element: 'p',
+  },
 });
--- a/src/model/paste/__tests__/DraftPasteProcessor-test.js
+++ b/src/model/paste/__tests__/DraftPasteProcessor-test.js
@@ -14,38 +14,8 @@
 jest.disableAutomock();
 
 var DraftEntity = require('DraftEntity');
-var Immutable = require('immutable');
 
 var DraftPasteProcessor = require('DraftPasteProcessor');
-var CUSTOM_BLOCK_MAP = Immutable.Map({
-  'header-one': {
-    element: 'h1',
-  },
-  'header-two': {
-    element: 'h2',
-  },
-  'header-three': {
-    element: 'h3',
-  },
-  'unordered-list-item': {
-    element: 'li',
-  },
-  'ordered-list-item': {
-    element: 'li',
-  },
-  'blockquote': {
-    element: 'blockquote',
-  },
-  'code-block': {
-    element: 'pre',
-  },
-  'paragraph': {
-    element: 'p',
-  },
-  'unstyled': {
-    element: 'div',
-  },
-});
 
 describe('DraftPasteProcessor', function() {
   function assertInlineStyles(block, comparison) {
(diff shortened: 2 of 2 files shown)


facebook-github-bot-9:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

CLowbrow:
@isaac this harkens back to https://github.com/facebook/draft-js/pull/205 which I stopped tracking before the end. Is this something we want?

Also, that cla form errors for me so I emailed the account 😆

spicyj:
Hah, @amasad was having trouble with the CLA too.

@jamesgpearce Another former employee. Did we find out why the CLA page was erroring?

howardjing:
With this approach, it looks like rendered blocks will be paragraphs that contain divs, which I think is invalid html. Is that a concern? It looks fine in chrome / firefox / safari though.

```
<p data-block="true" data-editor="fg6o8" data-offset-key="eint1-0-0">
  <div data-offset-key="eint1-0-0" class="public-DraftStyleDefault-block public-DraftStyleDefault-ltr">
    <span data-offset-key="eint1-0-0">
      <span data-text="true">some text here</span>
    </span>
  </div>
</p>
```

CLowbrow:
That's not great. Should I change it to: 

```
'paragraph': {
   element: 'div',
},
```

or is that not good/too hacky?

spicyj:
Definitely don't want div inside p. React DOM should actually warn you about that.

howardjing:
It looks like there would need to be a separate blockRenderMap for `convertFromHTMLtoContentBlocks` in that case, because otherwise `p` would not be a supported tag in `getChunkForHTML` [here](https://github.com/facebook/draft-js/blob/4cb881139e34b9bac84ed5847e010e3a495bd56f/src/model/encoding/convertFromHTMLToContentBlocks.js#L150)

So something like

```javascript
const DefaultDraftBlockRenderMap = Map({
  // ...
  'paragraph': {
    element: 'div',
  },
  // ...
});

const DefaultHTMLBlockRenderMap = DefaultDraftBlockRenderMap.set('paragraph', { 
  element: 'p',
});

function convertFromHTMLtoContentBlocks(
  html: string,
  DOMBuilder: Function = getSafeBodyFromHTML,
  blockRenderMap?: DraftBlockRenderMap = DefaultHTMLBlockRenderMap,
): ?Array<ContentBlock> {
  // ...
}
```

But yeah it does seem pretty hacky.

howardjing:
It looks like https://github.com/facebook/draft-js/pull/488/files is addressing the same issue as this PR -- the suggestion of having blockRenderMap take a list of valid `inputTypes` seems like a sane idea.

mitermayer:
Hi @CLowbrow, This issue has been fix by the PR https://github.com/facebook/draft-js/pull/828

PR #572, opened 2016-07-29
Android will loose focus if the contenteditable field is redrawn.  This forces keyboard close.

butchmarshall:
Android will loose focus if the contenteditable="true" field is also redrawn.

This restricts forcing redraw to the only the element instead of redrawing the contenteditable="true" element.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -57,6 +57,7 @@ const handlerMap = {
 
 type State = {
   containerKey: number,
+  contentsKey: number,
 };
 
 /**
@@ -114,7 +115,7 @@ class DraftEditor extends React.Component {
   blur: () => void;
   setMode: (mode: DraftEditorModes) => void;
   exitCurrentMode: () => void;
-  restoreEditorDOM: (scrollPosition: DraftScrollPosition) => void;
+  restoreEditorDOM: (scrollPosition: DraftScrollPosition, keyToUse: string) => void;
   setRenderGuard: () => void;
   removeRenderGuard: () => void;
   setClipboard: (clipboard?: BlockMap) => void;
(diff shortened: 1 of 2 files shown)


spicyj:
Hm. I'm worried that there could be other children of the contenteditable div in the DOM that wouldn't get removed if we only rekey the contents component.

butchmarshall:
I was wondering if there were any children @spicyj  - thanks for the clarification.

Would a more targeted updating method work?  

Right now I'm thinking:

- add a second key
- determine if we're inputting from a virtual-keyboard
- re-key only content="true" div if this is the case?
- else re-key everything

codejet:
Is there any chance of this getting merged sometime soon?

butchmarshall:
Hey @codejet, the concern here is we're not sure if this is the proper fix.  It fixes the keyboard issue, but due to a more targeted redraw we need to make sure it doesn't cause any other issues.

hellendag:
I don't think there should be any other children of the `contentEditable` div. As it is, I think I only pulled `DraftEditorContents` out into a separate component for `shouldComponentUpdate` purposes.

I'm actually wondering if it would always be safe to redraw *only* the `DraftEditorContents` when resetting the DOM.

- The reset behavior is only needed for *cut* and *composition* events.
- These events should only occur to contents that are definitely within the contentEditable.
- These events should only occur on descendants of `DraftEditorContents`.

The case where I can see problems occurring is where a cut or composition discards the entire `DraftEditorContents` for some reason.

Would you mind giving this a try and going through the cut/composition parts of the manual test plan to see if it works?

spicyj:
It seems like select all then cut would delete the contents completely…

dortzur:
Dear god, please merge this

dortzur:
FYI Googlers: I was able to sidestep this problem by by disabling autocorrect on the contentdeditable element on Android.
https://davidwalsh.name/disable-autocorrect

tonygentilcore:
@hellendag Your suggestion about *only* redrawing `DraftEditorContents` works like a charm. Here's a PR for that approach: https://github.com/facebook/draft-js/pull/907

aaren-cordova:
May want to also check out https://github.com/facebook/draft-js/pull/907, looks like it tackles the same issue with a similar fix.

flarnie:
We ended up merging a similar PR - https://github.com/facebook/draft-js/pull/907 so closing this.

Issue #385, opened 2016-05-12
Rendering isomorphically 


jamesjjk:
Draft-js is great! But we are rendering our app isomorpically, draft-js creates a checksum error when doing so? Using react 14.8. Any way we can avoid this?

spicyj:
What in the HTML differs?

jamesjjk:
```
 (client) true" data-editor="dulcp" data-offset-ke
 (server) true" data-editor="dj7ps" data-offset-ke

```

spicyj:
Interesting, okay. Maybe we can create and add that editor key after initial render or only when you copy…

jamesjjk:
Yep, any work around for this?

spicyj:
You could… stub out Math.random while doing the initial render? Other than that, not really. You could also look at not rendering the Draft component at all on the server.

jamesjjk:
@spicyj it would be cool to get a PR done to get the Math.random replaced as a key. Do you think FB will assign this as a bug? I will work with it as is for now, and consider not rendering isomorphically. Although thats the part I'm really keen on as I have 100% coverage at the moment.

Thanks!

spicyj:
It's mostly not clear what to replace it with. Certainly willing to accept good alternatives but this isn't very high priority for me.

jamesjjk:
I understand this key is used to define editor instances? Can we provide a prop for the key, else use math.random?

juhaelee:
@spicyj I'm also getting the following HTML differs in SSR:

```
(client) er" id="placeholder-87s16" data-reactid=
(server) er" id="placeholder-eur98" data-reactid=
```

on react v15 and latest (0.7.0) draft.js

reyhansofian:
hey guys, have you try this [method](http://stackoverflow.com/a/33521172/2763662) ?
Just wrap the react markup with `div`. I just tried, and it works.

gabriel-miranda:
Any thoughts on this? That method that @reyhansofian mentioned doesnt work :/

Keepcase:
@spicyj, are there any new updates (any discussions internally on how to tackle this)?
@reyhansofian, that method did not work.
@hellendag, do you have any ideas regarding a solution/workaround for this?

I'm getting the same checksum errors while using Draft.js with SSR (react 15.1.0 and draft-js 0.7.0):

```
(client) "true" data-editor="2ec27" data-offset-k
(server) "true" data-editor="3vera" data-offset-k
```

[...]

PR #582, opened 2016-08-06
Prefer rolling index over random key generation. Prefix editor key with

khankuan:
**Summary**

Ref: https://github.com/facebook/draft-js/issues/385

The current key generation method prohibits server-side rendering to have the same checksum when the editor is re-rendered on client side. The solution here is to keep editorKey in editorState so that developers have some control to retain the same editor key by sending the serialized editorState to client side.

There is also a performance concern on the current method to generate random keys. The method allows up to 2^24 keys in the seenKeys map. During generation, the method retries if a key previously used is generated again. This causes the process to retry more often as more keys get generated (evenly freezing) and would be a huge concern on server side rendering. The proposed method is to use a non-random but incrementing variable.


khankuan:
@hellendag would be great to get comments on the PR :)

khankuan:
The PR might still be buggy for SSR so i'm dropping it first. Will look deeper into usage of random keys.

PR #587, opened 2016-08-09
Fix nested editable elements in readOnly editor

hellendag:
**Summary**

The introduction of `user-modify: read-write-plaintext-only;` ends up breaking readOnly editors with nested editable content, as the cursor jumps back to the top-level editor when focus changes.

Fix this by applying the `user-modify` value only if the content node is editable.

**Test Plan**

  - Run TeX editor.
  - Click a formula. Type inside the textarea. Verify that I can do so.

### DIFF
--- a/src/component/base/DraftEditor.css
+++ b/src/component/base/DraftEditor.css
@@ -13,7 +13,7 @@
   text-align: initial;
 }
 
-.public/DraftEditor/content {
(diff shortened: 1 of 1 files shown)


spicyj:
I guess because user-modify should be read-only? I have a minor preference using a class name for this instead of an attribute selector but I suppose this also works.

PR #588, opened 2016-08-09
Preserve depth of all block types when converting to raw

### LINKED ISSUE #381: Depth of (custom) blocks is not preserved when converting to raw
As can be seen on the console demo, the conversion to raw doesn't serialize depth properly. (In the first raw state I have entered it manually.)
Console demo

If this needs more information drop me a line on Slack (draftjs.slack.com - @viclick) or contact me by any other means.

bryanrsmith:
**Summary**
Previously, `convertFromDraftStateToRaw` stripped the depth of all non-list blocks, making it difficult to implement custom block types that support depth/indentation. This seems a bit arbitrary since `convertFromDraftStateToRaw` is the only code in Draft with this restriction; the rest of the code does allow depth to be set on any block, and convertFromRawToDraftState` does allow depths to be specified at import time on any block type.

This change causes `convertFromDraftStateToRaw` to always export the block depth regardless of block type.

**Test Plan**

All existing tests pass, and I don't believe this change warrants any additional test coverage.

fixes #381

### DIFF
--- a/src/model/encoding/convertFromDraftStateToRaw.js
+++ b/src/model/encoding/convertFromDraftStateToRaw.js
@@ -18,7 +18,6 @@ var DraftStringKey = require('DraftStringKey');
 var encodeEntityRanges = require('encodeEntityRanges');
 var encodeInlineStyleRanges = require('encodeInlineStyleRanges');
 
-import type ContentBlock from 'ContentBlock';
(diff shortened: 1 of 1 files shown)


hellendag:
Yeah, this is reasonable. Thanks.

PR #595, opened 2016-08-12
Include object-assign in deps

### LINKED ISSUE #593: Cannot resolve module 'object-assign'
when update to 0.8.0  
it's report this error
and i should into draft-js and install object-assign  ,it's none error ,
am i miss something or it's just a bug?

zpao:
compiled code now requires `object-assign` due to a new transform

Fixes #593

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
   },
   "dependencies": {
     "fbjs": "^0.8.3",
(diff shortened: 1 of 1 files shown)


zpao:
Note: we should probably ship this as 0.8.1 ASAP.

PR #601, opened 2016-08-15
Add contentState argument to decorator

mizchi:
This is a proposal and document and test are imperfect yet.

**Summary**

Add contentState to decorator callback.

Example

```js
    strategy(contentBlock, callback, newContent) {
       // ... use newContent to decide what contentBlock is
    }
```

** Reason **

I need to know whole newState in decorator to decide itself but has no clue now.

Example

```
--- start-block
yey, here is block what i want to decorate
--- end-block
```

These are 3 block but middle one need to know upper and below lines.

My personal trigger is here http://stackoverflow.com/questions/38893815/how-do-i-make-multiline-hilighter-on-draft-js

... but I don't know this is correct way. Anyone has better way? or need this?

### DIFF
--- a/src/model/decorators/CompositeDraftDecorator.js
+++ b/src/model/decorators/CompositeDraftDecorator.js
@@ -16,6 +16,7 @@
 var Immutable = require('immutable');
 
 import type ContentBlock from 'ContentBlock';
+import type ContentState from 'ContentState';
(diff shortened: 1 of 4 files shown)


facebook-github-bot-5:

hellendag:
I believe this should be resolved by #376.

PR #613, opened 2016-08-23
RichUtils onBackspace: remove entire atomic block

### LINKED ISSUE #554: BlockData is not being removed after deleting atomic block
When the cursor is placed immediately after an atomic block and the user presses the backspace button, the preceding atomic block is removed, but the block data for that atomic block is kept resp. assigned to the following block. Example:

Cursor is placed at the very beginning of TextBlock2
```
|-----------------|
| TextBlock1      |
| AtomicBlock     |
| |TextBlock2     |
|-----------------|
```

User is pressing backspace, atomic block is being removed
```
|-----------------|
| TextBlock1      |
| |TextBlock2     |
|-----------------|
```

Now TextBlock2 has got the block data that was previously attached to the atomic block, which is wrong in my opinion. TextBlock2 has no use for the data, that was meant for use by the atomic block.

I am using RichUtils.onBackspace resp. RichUtils.handleKeyCommand (which takes care of removing atomic blocks).

qrohlf:
**Summary**

Previously, RichUtils "removed" atomic blocks via removeRangeFromContentState. This was problematic, because it preserves the atomic block properties and merges the end block content, resulting in behavior where atomic block data would transfer to unstyled blocks.

This change addresses those issues by removing the atomic block entirely from the blockMap and preserving the existing selection.

**Test Plan**

given the following blockMap in an Editor which is using RichUtils' handleKeyCommand

```
[
  {
    "key": "2124c",
    "text": "abc",
    "type": "unstyled",
    "depth": 0,
    "inlineStyleRanges": [],
    "entityRanges": [],
    "data": {}
  },
  {
    "key": "6ud8",
    "text": "",
    "type": "atomic",
    "depth": 0,
    "inlineStyleRanges": [],
    "entityRanges": [],
    "data": {
      "type": "image",
      "src": 1684980,
      "operation": null
    }
  },
  {
    "key": "a8vr5",
    "text": "def",
    "type": "unstyled",
    "depth": 0,
    "inlineStyleRanges": [],
    "entityRanges": [],
    "data": {}
  }
]
```

positioning the cursor before 'd' and pressing backspace should result in a blockMap which looks like this, with the cursor positioned before the 'd'

```
[
  {
    "key": "2124c",
    "text": "abc",
    "type": "unstyled",
    "depth": 0,
    "inlineStyleRanges": [],
    "entityRanges": [],
    "data": {}
  },
  {
    "key": "a8vr5",
    "text": "def",
    "type": "unstyled",
    "depth": 0,
    "inlineStyleRanges": [],
    "entityRanges": [],
    "data": {}
  }
]
```

*you can see what happened before this change [here](https://gist.github.com/qrohlf/4d3fa3c26c06503cdc10c2b03fa11937) for comparison*

Since the `if` conditions in onBackspace ensure that this code will run only in the case where the selection is collapsed and positioned at the start of a block which is preceded by an atomic block, no other behavior is impacted.  

Fixes #554

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -119,26 +119,10 @@ const RichTextEditorUtil = {
     var blockBefore = content.getBlockBefore(startKey);
 
     if (blockBefore && blockBefore.getType() === 'atomic') {
-      var atomicBlockTarget = selection.merge({
-        anchorKey: blockBefore.getKey(),
-        anchorOffset: 0,
-      });
-      var asCurrentStyle = DraftModifier.setBlockType(
-        content,
-        atomicBlockTarget,
-        content.getBlockForKey(startKey).getType()
-      );
-      var withoutAtomicBlock = DraftModifier.removeRange(
-        asCurrentStyle,
-        atomicBlockTarget,
-        'backward'
-      );
+      const blockMap = content.getBlockMap().delete(blockBefore.getKey())
+      var withoutAtomicBlock = content.merge({blockMap, selectionAfter: selection});
       if (withoutAtomicBlock !== content) {
-        return EditorState.push(
-          editorState,
-          withoutAtomicBlock,
-          'remove-range'
-        );
+        return EditorState.push(editorState, withoutAtomicBlock, 'remove-range');
       }
     }
 


facebook-github-bot-4:

facebook-github-bot-4:

qrohlf:
@hellendag or @spicyj, anything I can do to help move this thing along?

hellendag:
Seems reasonable, thanks!

J22Melody:
Cool! May I ask when will the next stable version including this fix publish?

leeching:
In my using case, checking `blockBefore` is not enough. Considering the following situation:

original html

```html
<figure><img src="1.jpg" /></figure>
```

rendered html by `convertToHTML` api of draft-js

```html
<figure class="" data-block="true" data-editor="1pbbq" data-offset-key="7jm7f-0-0">
    <div data-offset-key="7jm7f-0-0" class="public-DraftStyleDefault-block public-DraftStyleDefault-ltr">
        <img src="http://cf.dtcj.com/Fp7FCDZpSct1bJgfgkbTYLrHdxq7" height="100">
    </div>
</figure>
```

If the cursor is after img tag but still in figure tag, it does not meet the condition `blockBefore.getType() === 'atomic'` and the code in if block will be skipped. It is better to check whether the cursor is in a atomic-type block as well.

PR #629, opened 2016-09-03
Add table styles to documentation

flarnie:
We have [a PR to add a new page to the docs,] and it includes a table
showing the mapping of block types to elements. However, we don't have
good styles for tables in the documentation.

[a PR to add a new page to the docs,]: https://github.com/facebook/draft-js/pull/387

Adding these styles should unblock that PR so we can merge it and update
the Draft.js website.

I've attached a 'before' and 'after' screenshot of the new docs to show how these styles will help:
**Before:**
before_styles

**After:**
after_styles

### DIFF
--- a/website/src/draft-js/css/draft.css
+++ b/website/src/draft-js/css/draft.css
@@ -106,6 +106,30 @@ a:focus {
   outline-offset: -2px;
 }
 
+table {
+  border-collapse: collapse;
+  margin: 15px 16px;
+}
+
+table tr {
+  background-color: #fff;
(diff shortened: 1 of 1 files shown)


stopachka:
looks great!

flarnie:
It seems like the CSS change wasn't actually applied. I landed the documentation change assuming the table styles were there, but they are not. Do we need to do some other step here?

stopachka:
Cheers @flarnie -- looks like I moved so fast I didn't click "squash and merge" :P.

PR #630, opened 2016-09-04
update handleKeyCommand docs to return DraftHandleValue

davidchang:
Updating documentation to say handleKeyCommand returns 'handled' or 'not-handled' (the DraftHandleValue type) as opposed to true or false (boolean type)


stopachka:
nice!

Issue #491, opened 2016-06-26
Should allow user to define undo boundary states


mitermayer:
From the API documentation page it is recommended that all changes on the contentState to be applied using 'EditorState.push' the problem is that it will always create a new undoLevel. 

Internally we have boundary states, where it does not generate undo levels. Ideally we should be able to extend it somehow.

this is the section that deals with it at the moment on file `src/model/immutable/EditorState.js`

```
570 /**
571  * Return whether a change should be considered a boundary state, given
572  * the previous change type. Allows us to discard potential boundary states
573  * during standard typing or deletion behavior.
574  */
575 function mustBecomeBoundary(
576   editorState: EditorState,
577   changeType: EditorChangeType
578 ): boolean {
579   var lastChangeType = editorState.getLastChangeType();
580   return (
581     changeType !== lastChangeType ||
582     (
583       changeType !== 'insert-characters' &&
584       changeType !== 'backspace-character' &&
585       changeType !== 'delete-character'
586     )
587   );
588 }
```

sarmstrong:
+1, I need this when performing bulk operations using EditorState.push and I want to restore to the point before the bulk operation. I'd also be cool w/ a simple boolean flag that would exclude it from undo history.

PR #635, opened 2016-09-07
Add a way to prevent replaced content to be added onto UndoStack

AlbertHilb:
This pull request is a proposal to fix issue #491.

It adds an optional boolean parameter, `boundaryState`,  to `EditorState` `push` method.
If `boudaryState` is given, depending on its value, the replaced content is considered a boundary state and added onto `UndoStack` or not.
Otherwise, as before, the choice of adding replaced content onto `UndoStack` is based of the `changeType`.

### DIFF
--- a/docs/APIReference-EditorState.md
+++ b/docs/APIReference-EditorState.md
@@ -72,7 +72,7 @@ The list below includes the most commonly used instance methods for `EditorState
   </li>
   <li>
     <a href="#push">
-      <pre>static push(editorState, contentState, changeType): EditorState</pre>
+      <pre>static push(editorState, contentState, changeType, ?boundaryState): EditorState</pre>
     </a>
   </li>
   <li>
@@ -262,12 +262,14 @@ if you need custom configuration not available via the methods above.
 static push(
   editorState: EditorState,
   contentState: ContentState,
-  changeType: EditorChangeType
+  changeType: EditorChangeType,
+  ?boundaryState: boolean
 ): EditorState
 ```
 Returns a new `EditorState` object with the specified `ContentState` applied
-as the new `currentContent`. Based on the `changeType`, this `ContentState`
-may be regarded as a boundary state for undo/redo behavior.
+as the new `currentContent`. Based on `boundaryState`, if given, or on
+the `changeType`, if not, this `ContentState`may be regarded as a boundary
+state for undo/redo behavior.
 
 All content changes must be applied to the EditorState with this method.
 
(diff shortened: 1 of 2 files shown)


hellendag:
Seems useful!

I think the optional boolean argument is a little awkward, though. (I always feel that way about boolean arguments.) It means that there are three states:

1. Explicitly make it a boundary
1. Explicitly make it *not* a boundary
1. Defer to core logic

I'm not sure the second option here is even valuable. (Unless you have a use case for it?)

Perhaps it may make more sense to be explicit:

```javascript
push(...): EditorState;
pushBoundaryState(...): EditorState;
```

In this way, you are being as clear as possible about what you're doing. Under the hood, `push` can use `pushBoundary`.

What do you think?

AlbertHilb:
I need exactly the behavior described in the second point. :-)

I'm working to a simple scientific editor which can be used to write documents containing math formulas. It is similar to the draft-js TeX example, just with few additional features.

To insert a new piece of mathematics, users must:

1. Click on a button in the toolbar.   
    The main editor is placed in `readOnly` mode and a custom block is inserted in the document. It contains a widget to set the TeX source of the formula and one to render it.
1. Write the math content using the dedicated widget.  
    When they have finished the input widget is hidden, the source is stored as block metadata and the main editor is put again in read-write mode.

Let's call A the initial content state, B the state after point 1 and C the final state.
In my view, those listed above are two distinct phases of the same action. So, on triggering undo when the editor is in the state C, the content of the editor should return to the state A.
To make this happen, I need a way to prevent the state B from being added to the `UndoStack`.

About the shape of the API, I have no definite preference.
If you prefer to not use boolean parameters, I'm open to other solutions.

AlbertHilb:
To avoid using optional boolean parameter, I'm thinking around something like this.

We can introduce a new enum type:

```
type ReplacedContentStateDestiny (
  'undostack' |
  'discard' |
  'auto'
)
```

and a new `EditorState` static method with signature:

```
static replaceContent(
  editorState: EditorState,
  newContentState: ContentState,
  changeType: EditorChangeType,
  replacedContentDestiny?: ReplacedContentStateDestiny
): EditorState
```

Then `push` method can be rewritten using `replaceContent` and deprecated in view of its future removal.

stha:
:+1:

flarnie:
Looking this over now - I don't think we have a use case for this at Facebook, and it doesn't seem to come up often for folks building text editors. I'm not sure we want to add the extra complexity right now. I'd like to close this for now, but once we fix some of the open bugs with Draft it could be worth coming back to. I'd also be curious to talk to @mitermayer more about https://github.com/facebook/draft-js/issues/491 - maybe I'm missing some functionality this enables.

Issue #638, opened 2016-09-09
Cannot paste from clipboard in Firefox


robbertbrak:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**

* Copy an image to the clipboard (for example by taking a screenshot).
* Go to http://facebook.github.io/draft-js/ in Firefox.
* Focus the editor and paste the clipboard contents in the editor.
* Result: the console shows an error: "TypeError: appendToHead is undefined". The callback function `handlePastedFiles` is not called.

**What is the expected behavior?**

I expected there to be no errors and that I would be able to access the clipboard contents in the `handlePastedFiles` callback function.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

v0.8.1, Firefox 48.

flarnie:
Fix just landed, and will be out in the next release. Thanks for reporting this issue!

PR #654, opened 2016-09-14
Edit on paste truthy fix

tylercraft:
**Summary**

Fix for #638

**Test Plan**

textBlocks is defined on line 88 `let textBlocks: Array<string> = [];` and therefore is always truthy. So this if statement was always inserting a fragment even when it shouldn't.

### DIFF
--- a/src/component/handlers/edit/editOnPaste.js
+++ b/src/component/handlers/edit/editOnPaste.js
@@ -32,7 +32,6 @@ const isEventHandled = require('isEventHandled');
 function editOnPaste(e: SyntheticClipboardEvent): void {
   e.preventDefault();
   var data = new DataTransfer(e.clipboardData);
-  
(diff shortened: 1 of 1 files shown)


src/component/handlers/edit/editOnPaste.js
```diff
@@ -32,7 +32,7 @@ const isEventHandled = require('isEventHandled');
 function editOnPaste(e: SyntheticClipboardEvent): void {
   e.preventDefault();
   var data = new DataTransfer(e.clipboardData);
-  //debugger;
+
```
flarnie: I'm curious about how that got in there? Seems like the linter would catch it.

flarnie:
This seems pretty reasonable. I'm going to check it out and manually test. Also looking into why ESLint is failing, here and on master.

PR #667, opened 2016-09-19
Fix modifying state during onChange triggered by beforeinput

### LINKED ISSUE #92: Typed char gets duplicated if also changing block type in onChange
I want to change the block type in onChange, however the typed character gets duplicated when doing that. Backspaces is not duplicated however so I can still remove one character at a time, not just type them.

Doesn't the editor support pushing another content change before being rerendered with its updated one?

spicyj:
Fixes #92.

If an onChange handler replaces every "x" with an "abc", then when you type "x", we trigger onChange from beforeinput and rerender immediately with the content "abc", then the browser still inserts an "x" character!

This fix is surprisingly simple: instead of triggering an update during beforeinput, we simply wait until the input event to do so. This means that when we rerender, the browser would have already inserted the "x" character and React will correctly delete it when rerendering.

Test Plan: In plaintext example, change the handler to

```
this.onChange = (editorState) => {
  var content = editorState.getCurrentContent();
  return this.setState({
    editorState: EditorState.set(editorState, {
      currentContent: Modifier.removeInlineStyle(content, content.getSelectionAfter(), 'OVERFLOW');
    }),
  });
};
```

then type. Characters don't get duplicated like they did before.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -87,6 +87,7 @@ class DraftEditor extends React.Component {
   _editorKey: string;
   _placeholderAccessibilityID: string;
   _latestEditorState: EditorState;
+  _pendingStateFromBeforeInput: ?EditorState;
 
   /**
    * Define proxies that can route events to the current handler.
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -155,14 +155,16 @@ function editOnBeforeInput(editor: DraftEditor, e: SyntheticInputEvent): void {
     )
   ) {
     e.preventDefault();
+    editor.update(newEditorState);
   } else {
-    // The native event is allowed to occur.
-    newEditorState = EditorState.set(newEditorState, {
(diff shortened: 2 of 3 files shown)


spicyj:
cc @hellendag – This fix felt too obvious to me. Am I missing some reason why it can't work?

spicyj:
This doesn't work; typing `#a` in the tweet example only inserts a `#`. Need to investigate.

spicyj:
Fixed – I had a silly mistake. :)

@stopachka I choose you! Think this makes sense? And @hellendag too if you have time.

hellendag:
Thanks! This seems reasonable to me!

I may not have considered this because I regard the post-mutation nature of the `input` event to be generally useless for Draft purposes (except for things like spellcheck, as used here), but this seems like a pretty clever way to use it.

This is enough of a core input change that it's worth staying on your toes to make sure it doesn't lead to any mysterious editing issues.

🦄

paulyoung:
If there's anything I can do to help move this along, please let me know.

davidgeilfus:
Hi, any idea when this will be merged? Thank you

PR #689, opened 2016-09-26
Fixed using IME at the beginning of a soft newline.

houkanshan:
**Summary**

In Webkit browsers, composition at the beginning of a soft newline will create new text nodes. 
Which makes the block leafs & selection into a strange state and lead to bugs. 
I guess it's just like the issue of composition at the beginning of a leaf.

Here is one of the bugs:

**Step**:
1. Open Facebook notes editor (because it can insert soft newline).
2. Type any words like `abcd`.
3. Press `shift+return` to insert a soft newline.
4. Enable IME (Chinese pinyin for example), type `caogao`, and press space.
5. Type `.`, you'll see the first line been removed.
6. Press `cmd+z` will recover the first line. (It was the `editOnInput` handle for `.`, that replace the whole block with the text after the `\n`)  

This GIF shows it.

draft

And the new text nodes it created after the composition:

screen shot 2016-09-26 at 10 53 48 pm

To make it easy to test,    26dfa8e enabled soft new line in rich example. I can remove this commit after you tested it. If you'd like to keep it, I can sync it with the homepage.

**Test Plan**

Same as the **step** above.

Thanks for review!

**EDIT**:
Latest Chrome has fixed this bug, but it still exist in Chrome 54 & Safari 10.0.3

### DIFF
--- a/examples/draft-0-10-0/rich/RichEditor.css
+++ b/examples/draft-0-10-0/rich/RichEditor.css
@@ -36,6 +36,10 @@
   padding: 10px 20px;
 }
 
+.RichEditor-editor .public-DraftStyleDefault-block {
+  margin-bottom: 16px;
+}
+
 .RichEditor-editor .public-DraftStyleDefault-pre {
   background-color: rgba(0, 0, 0, 0.05);
   font-family: 'Inconsolata', 'Menlo', 'Consolas', monospace;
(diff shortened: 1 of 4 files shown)


facebook-github-bot-7:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

facebook-github-bot-7:

flarnie:
Hi @houkanshan - thanks for submitting for this! I really appreciate the detailed instructions for reproducing one of the bugs.

We have released a couple of versions of Draft with bug-fixes for IME, and I'm no longer able to reproduce the bug you described. We may have fixed this behavior when making other improvements - but I know there are still open problems with IME, and so please reopen this or ping me if there are other problems with soft newlines which this could fix. :)

houkanshan:
@flarnie Thank you for your reply! 

I've tested on Chrome 54 & Safari 10.0.3 (with macOS Sierra). The first line won't be removed, but the caret will jump back to the head, as the gif shows:
bug1

PR #694, opened 2016-09-30
Add `select` to atomic block component props

AlbertHilb:
Add `select` to the properties passed to the component used to render atomic blocks.
It is a boolean parameter, true when the block is inside selection, false otherwise.
It can be used, for example, to customize atomic blocks aspect when they are selected.

The main change is that `DraftEditorContents` re-renders every time the editor state
changes and the old selection or the new one contain atomic blocks.

This PR is intended, essentially, to test if there is interest around such a proposal.
If yes, there are few aspects to be discussed.


flarnie:
Hi - Sorry we took so long to review this.

Thanks for making this proposal - for now we prefer to lean on the side of avoiding any new APIs unless there is a strong need. I would say that because this PR is stale and nobody else has jumped on to comment, there may not be enough interest to warrant adding this.

Please feel free to open an issue to discuss the use case for this more.

Issue #697, opened 2016-10-02
Tex example is broken -- JS errors immediately upon page load


ericbiewener:
The tex example immediately throws an error when trying to open `localhost:3000` in my browser:

`Uncaught TypeError: Cannot read property 'get' of undefined`

<img width="1440" alt="screen shot 2016-10-02 at 3 35 51 pm" src="https://cloud.githubusercontent.com/assets/253298/19024311/fb3f4d2e-88b5-11e6-8fdf-90ab1331a48d.png">
<img width="1440" alt="screen shot 2016-10-02 at 3 36 01 pm" src="https://cloud.githubusercontent.com/assets/253298/19024312/fb535da0-88b5-11e6-9486-5b97eb7e3a29.png">

*To Reproduce*
1. `cd examples/tex`
2. `npm install`
3. `npm start`
4. Open `localhost:3000` in browser.

mzbac:
I am guessing this is caused by data structure change of moving entity data  to contentBlock

flarnie:
Fix just landed - thanks for flagging this @ericbiewener.

PR #699, opened 2016-10-03
Fix/tex example

mzbac:
this pull request fixed broken Tex example by removing outdated syntax that changed
in #376.

refer issue to https://github.com/facebook/draft-js/issues/697


flarnie:
Thanks for fixing this!

PR #705, opened 2016-10-07
Upgrade babel-preset-fbjs

zpao:
This fixes 2 issues when Draft is a dependency of a project that is
using Flow.
- `typeof` import now rewritten correctly
- `__DEV__` declaration is now inlined

Here's the diff on master right now: https://gist.github.com/zpao/641b31feae17be7e3e1fabc9aecd10a3

Note: this doesn't impact anything internal and is only for the npm package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "devDependencies": {
     "babel-core": "^6.8.0",
     "babel-eslint": "^6.1.2",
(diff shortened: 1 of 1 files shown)


davidchang:
this looks good to me, would merge, but i don't have the ability to do so (and i think zpao just headed off on vacation)... can someone else merge this?

PR #709, opened 2016-10-10
Re-apply current inline style after splitting a block with the return key

### LINKED ISSUE #672: Inline style is incorrectly toggled when starting a new block
To reproduce:
- Go to the example editor on http://facebook.github.io/draft-js/.
- Select 'Bold' and type some text.
- Deselect 'Bold'.
- Press enter to start a new block.
- 'Bold' is unexpectly selected again. When you type something, the text is bold.

The other way around doesn't work either:
- Go to the example editor on http://facebook.github.io/draft-js/.
- Type some text (without any inline styles).
- Select 'Bold'.
- Press enter to start a new block.
- 'Bold' is unexpectly deselected again. When you type something, the text is not bold.

It seems that the 'current' inline style does not survive when you start a new block. I might be able to work around this by doing something with `inlineStyleOverride`, but I think this is a bug that needs to be fixed in the framework.

robbertbrak:
Fixes #672.

### DIFF
--- a/src/component/handlers/edit/commands/keyCommandInsertNewline.js
+++ b/src/component/handlers/edit/commands/keyCommandInsertNewline.js
@@ -20,7 +20,16 @@ function keyCommandInsertNewline(editorState: EditorState): EditorState {
     editorState.getCurrentContent(),
     editorState.getSelection()
   );
-  return EditorState.push(editorState, contentState, 'split-block');
+
+  var newEditorState = EditorState.push(
+    editorState,
+    contentState,
+    'split-block');
+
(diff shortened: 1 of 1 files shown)


davidchang:
thanks for the PR @robbertbrak - this looks okay to me, but i'm just unclear as to if this is the best way to handle this. can someone else also take a look and chime in? cc @flarnie @stopachka @tylercraft @hellendag

flarnie:
Looks like [we merged another approach to fixing the problem](https://github.com/facebook/draft-js/pull/892) - thanks anyway for submitting this PR!

PR #727, opened 2016-10-19
Fix repeated backspace, other internal components

spicyj:
I can't describe entirely why the DraftEditor.react change here is necessary, but this is closer to what we had before #666.

The editOnSelect change fixes holding down Backspace in Firefox -- before this, the cursor would jump to the start of the input.


spicyj:
(accepted/landed internally already)

PR #731, opened 2016-10-21
Replace reference to headers to headings H-elements

### LINKED ISSUE #724: Semantics: "Header" should be "Heading" in rich.html
I'd like to point out minor semantics issues I see in `rich.html`.

Currently the example refers to `h1`–`h6` as “headers”, when really they should be called “headings”.

This perpetuates incorrect semantics when referring to these tags. “Headers” are generally known to be layout blocks that contain header information, while “Headings” are defined as block-level text titles. While people casually use these two terms interchangeably, they really should be better defined to avoid confusion.

jihdeh:
This PR fixes #724 ,  which aims to replace references to `H1 - H2` as headers to headings.

### DIFF
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -20,13 +20,12 @@ by matching the Draft block render map with the matched tag.
 
 |  HTML element   |            Draft block type             |
 | --------------- | --------------------------------------- |
-|     `<h1/>`     |               header-one                |
-|     `<h2/>`     |               header-two                |
-|     `<h3/>`     |              header-three               |
-|     `<h4/>`     |               header-four               |
-|     `<h5/>`     |               header-five               |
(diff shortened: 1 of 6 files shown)


davidchang:
would this require a major version update, since we'd be renaming all of the heading block type? this would be really problematic for others trying to load old ContentStates that expect the block type to say "header" instead of "heading", right?

jihdeh:
@davidchang Yes, it would require a major update if it would be causing that sort of issue.

@tylercraft what do you think?

tylercraft:
@jihdeh - Thank you for taking this on, but after looking at what it would take to migrate, this will be a bigger issue than I initially thought. While I agree that this is an issue, it will be too big one for migrations for just a semantic change. This is completely my fault and something I should have double checked prior to replying to your comment. I'm sorry if this took up much of your time and hope that you will still contribute to draft in the future.

PR #734, opened 2016-10-22
add image support for convertFromHTML, along with new example

tylercraft:
**Summary**

Adds support for  tags to be passed in to the convertFromHTML method. It treats them similarly as links. It outputs the src of the image tag as plain text by default if no custom decorator is specified.

**Test Plan**

Passes all tests. Created a new example page in the examples folder to demo.


examples/convertFromHTML/convert.html
```diff
@@ -0,0 +1,178 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+FACEBOOK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+-->
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Draft • Media Editor</title>
```
tylercraft: Good call, will update

flarnie:
Funny - the '\' got removed by Github from your PR summary.

As we talked about, we should move this to branch off of 0.9.1 and release it as part of a 0.9.2 with some of the other recent bug fixes. 

Also need to check the example in Chrome - something is not working there.

Thanks for working on this!

flarnie:
Ok - maybe @spicyj or @hellendag can help figure this out.

When I pull down this commit to manually test, the `media` example doesn't work. See attached gif.
For @tylercraft it does work locally. Also, if I comment out his changes, it still doesn't work on that branch, but works on master.

I'll look at it again tomorrow too, and see if I can figure it out.
failing_media_example

spicyj:
Do you both have https://github.com/facebook/draft-js/commit/449fea4723498143ae3c99a2a1702c6dcedae78f in your tree? Some similar stuff was broken in www before that.

flarnie:
Ok - it sounds like that bug in the media example is just related to https://github.com/facebook/draft-js/commit/07b1eb847fb801d61b41a7a03c7c26c80cca5478 which has since been fixed.

Issue #732, opened 2016-10-21
How to Retrieve Metadata for a Range of Text (Get the Current Href from Existing Link)?


superKalo:
Hello,

I have a question about retrieving metadata for a range of text.

I took the [link example](https://github.com/facebook/draft-js/blob/master/examples/link/link.html). Let's say we add a link to a range of text. I want the next time when I select the text and click "Add Link" - to retrieve the current link href and to pre-fill the input.

![selection](https://cloud.githubusercontent.com/assets/2548061/19595953/f538ea64-9794-11e6-82a2-a667d9ffb568.png)

So, basically I assume I need to [hook-up additional logic to the `_promptForLink` method](https://github.com/facebook/draft-js/blob/master/examples/link/link.html#L72) - instead of always having empty string for `urlValue`, to pre-fill with the current url for this text range (if it exists).

I read about how to [retrieve entities](https://facebook.github.io/draft-js/docs/advanced-topics-entities.html#creating-and-retrieving-entities), however - I must have missed something, because I can't figure out how to retrieve it. This doesn't work:

```javascript
_promptForLink(e) {
  e.preventDefault();
  const {editorState} = this.state;
  const selection = editorState.getSelection();
  if (!selection.isCollapsed()) {

    const contentState = editorState.getCurrentContent();

    // TODO: I can't really figure out what I need to pass to getBlockForKey
    const startKey = editorState.getSelection().getStartKey();
    const blockWithLinkAtBeginning = contentState.getBlockForKey(startKey);

    const linkKey = blockWithLinkAtBeginning.getEntityAt(0);
    const linkInstance = Entity.get(linkKey);
    const {href} = linkInstance.getData();

    this.setState({
      showURLInput: true,
      urlValue: href || '', // TODO: Pre-filll with the existing value here
    }, () => {
      setTimeout(() => this.refs.url.focus(), 0);
    });
  }
}
```

Can you please advice me how to do that?

ericbiewener:
You're almost there, but you need to grab the `url` property off `linkInstance.getData()`, not `href`. `getData` does not provide you with DOM metadata, but rather the metadata of the `Entity` at the time it was created. In the link example, you can see that this property is named `url` [here](https://github.com/facebook/draft-js/blob/master/examples/link/link.html#L93).

superKalo:
Thank you @ericbiewener !

[...]

PR #735, opened 2016-10-24
Feature / Link Example: Pre-fill the URL Input with the Current URL (if any)

superKalo:
**Summary**

When adding a new link on top of another link, pre-fill the input with the current url value. Use-case described in details here: https://github.com/facebook/draft-js/issues/732

I believe the example enhancement will be useful for other developers too.

**Test Plan**

I loaded the example and manually tested on Chrome / Safari / Firefox / Opera latest.

### DIFF
--- a/examples/link/link.html
+++ b/examples/link/link.html
@@ -76,8 +76,9 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
           if (!selection.isCollapsed()) {
             const contentState = editorState.getCurrentContent();
             const startKey = editorState.getSelection().getStartKey();
+            const startOffset = editorState.getSelection().getStartOffset();
             const blockWithLinkAtBeginning = contentState.getBlockForKey(startKey);
-            const linkKey = blockWithLinkAtBeginning.getEntityAt(0);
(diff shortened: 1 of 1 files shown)


superKalo:
Roger that! I just signed it.

leeching:
Thanks @superKalo ! Your code helps me a lot.

There is one line:

```js
const linkKey = blockWithLinkAtBeginning.getEntityAt(0);
```

It is better to provide `this.state.editorState.getSelection().getStartOffset()` instead of `0`. 

Consider the following situation: 

```html
<a href="http://github.com">github</a><a href="http://www.google.com>google</a>
```

If someone selects `google` text, returned url will be 'http://github.com'.

superKalo:
Yes, @leeching , you're correct.

I spot that as well (while being on the issue) and that's exactly how I did it in the pull request, see line [79](https://github.com/facebook/draft-js/pull/735/files#diff-eff3b85f7cf75258cc7223f10a408f8dR79).

So in the current pull request I did, this use-case is covered.

davidchang:
(y) this looks great, thanks for updating the example @superKalo . merging!

superKalo:
Great! Thanks for accepting it, @davidchang

PR #740, opened 2016-10-25
Change handlers to pass `editor` as an argument

spicyj:
This allows Flow to type them better and catch bugs.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -181,7 +181,7 @@ class DraftEditor extends React.Component {
     return (e) => {
       if (!this.props.readOnly) {
         const method = this._handler && this._handler[eventName];
-        method && method.call(this, e);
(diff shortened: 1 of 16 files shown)


flarnie:
Thanks for pinging about getting a review! This absolutely makes sense - using the 'editor' explicitly instead of assuming that the handlers will be called with the editor as the context is more clear and eliminates the risk that they would be called with the wrong context.

This seems to be a refactor that only affects internal APIs. I'd still like to read through the surrounding code and do some manual testing before we squash and merge. Will do that later tonight, possibly also in the morning.

flarnie:
Manual testing looks good, and `grep -nr 'this' ./src/component/handlers` only returns comments. 

This reminds me again how cool the architecture of the handlers is - it's still the delegation pattern, just more explicit about what the context is.

Thanks again for doing this!

PR #746, opened 2016-10-28
Ensure `ContentState.createFromBlockArray` creates with an empty array

### LINKED ISSUE #720: `ContentState.createFromBlockArray` fails on empty array
We have an app that’s initialising a `ContentState` object through a block array, and we’re using the `createFromBlockArray` method to do this. Unfortunately it fails if the block array is empty, which you can see in this fiddle: https://jsfiddle.net/m6z0xn4r/27/

```js
const {ContentState} = Draft;
const contentState = ContentState.createFromBlockArray([]);
```

The error appears to be that [this line assumes that the `blockMap` will always have a first value](https://github.com/facebook/draft-js/blob/master/src/model/immutable/ContentState.js#L175). This is easy enough to work around obviously, but it feels like a rough edge that’d be nice to smooth over.

makenosound:
Fixes #720 by passing through an empty `SelectionState` when the created `blockMap` is empty.

### DIFF
--- a/src/model/immutable/ContentState.js
+++ b/src/model/immutable/ContentState.js
@@ -172,7 +172,9 @@ class ContentState extends ContentStateRecord {
     entityMap: ?OrderedMap
   ): ContentState {
     var blockMap = BlockMapBuilder.createFromArray(blocks);
-    var selectionState = SelectionState.createEmpty(blockMap.first().getKey());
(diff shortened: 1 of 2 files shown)


davidchang:
lgtm, merging! thanks @makenosound !

Issue #443, opened 2016-06-05
Indenting list block of one type inside the other


jpuri:
If I create a OL of 2 items and change depth of 2nd item to 2, like this:
```
1. ITEM1
      2. ITEM2
```

If I change type of 2nd item to UL it looses it depth:
```
1. ITEM1

2. ITEM2
```

But now if I try to indent second item, it works
```
1. ITEM1

    2. ITEM2
```

While indenting its checked that block above is list, but its not checked if its list of same type. Also, please find gif attached. I can create a PR for the fix.

jpuri:
![indenting](https://cloud.githubusercontent.com/assets/2182307/15806439/02588d3a-2b61-11e6-96fa-556f52b983ae.gif)

hellendag:
Thanks for the GIF -- very helpful. :)

I don't recall whether this was intentional, but block depth is always reset when setting type: https://github.com/facebook/draft-js/blob/master/src/model/transaction/setBlockTypeForContentState.js#L35

I'd be open to a PR to change this, since it seems reasonable to preserve depth on list items.

ericbiewener:
My results are a bit different than what the GIF shows. In the GIF, re-indenting the now ordered `<li>` at the end causes it to become nested with the `<li>` of the unordered list. However, when I try to recreate this, the lists do not nest. In fact, I can't created truly nested lists--from the perspective of the DOM--with Draft.js at all. Instead, all nesting seems to be achieved solely through the use of class names, e.g. `.public-DraftStyleDefault-depth1`, which in combination with adding margins via CSS can be made to appear nested.

While this is a clever approach that surely simplifies things a lot, it has the disadvantage of not being able to support ordered list numbering when the `<li>`s of the ordered list have children `<li>`s:

![image](https://cloud.githubusercontent.com/assets/253298/19733825/12b3ad5e-9b5a-11e6-9fcb-2b83d6839310.png)

And the corresponding DOM:

![image](https://cloud.githubusercontent.com/assets/253298/19733859/399cdb84-9b5a-11e6-8809-3c466fceeaa1.png)

ericbiewener:
Adding on to my previous comment, I see that the Facebook notes editor can supprt the following:

![image](https://cloud.githubusercontent.com/assets/253298/19742728/3e7f6a24-9b7b-11e6-944a-535cebd246f6.png)

And the corresponding DOM:

[...]

PR #752, opened 2016-10-29
Use the same wrapper for both ordered and unordered list. Maintain bl…

ericbiewener:
**Summary**

Adds support for nesting ordered & unordered lists within one another, preserving the block depth when toggling between the two. This fixes https://github.com/facebook/draft-js/issues/443 in the process.

The enhancement is simple. Since DraftJS already uses classes in combination with the `::before` pseudo element to custom render ordered list numbering rather than depending on default browser functionality, there isn't actually any reason to render an `<ol>` vs `<ul>`. By giving both block types the same wrapper, we automatically gain support for nesting one within the other.

To maintaining block depth with toggling between the two list types, we simply merge in the block's current depth rather than 0.

**Test Plan**

Open up the `rich.html` example and create an ordered list. Indent a list item and convert it to an unordered list. It will remain nested at the same depth. Play around further, adding ordered and unordered lists throughout the list tree. Ordered list numbering will remain correct, restarting at new levels of depth and continuing from the last ordered list item at the same depth.

image

### DIFF
--- a/src/component/utils/DraftStyleDefault.css
+++ b/src/component/utils/DraftStyleDefault.css
@@ -35,8 +35,7 @@
 /**
  * Default spacing for list container elements. Override with CSS as needed.
  */
-.public/DraftStyleDefault/ul,
-.public/DraftStyleDefault/ol {
+.public/DraftStyleDefault/list {
   margin: 16px 0;
   padding: 0;
 }
--- a/src/model/immutable/DefaultDraftBlockRenderMap.js
+++ b/src/model/immutable/DefaultDraftBlockRenderMap.js
@@ -17,8 +17,7 @@ const React = require('React');
 
 const cx = require('cx');
 
-const UL_WRAP = <ul className={cx('public/DraftStyleDefault/ul')} />;
-const OL_WRAP = <ol className={cx('public/DraftStyleDefault/ol')} />;
+const LIST_WRAP = <ul className={cx('public/DraftStyleDefault/list')} />;
 const PRE_WRAP = <pre className={cx('public/DraftStyleDefault/pre')} />;
 
 module.exports = Map({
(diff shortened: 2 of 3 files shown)


davidchang:
thanks for the PR @ericbiewener - would this cause introduce inconsistency in conversions to HTML?

davidchang:
thanks for the article @ericbiewener - this actually looks all right to me.

can someone else take a look? @flarnie @stopachka @tylercraft @hellendag?

mzbac:
any update for this? really looking forward this fix

src/component/utils/DraftStyleDefault.css
```diff
@@ -35,8 +35,7 @@
 /**
  * Default spacing for list container elements. Override with CSS as needed.
  */
-.public/DraftStyleDefault/ul,
-.public/DraftStyleDefault/ol {
+.public/DraftStyleDefault/list {
```
nowen094: +.public/DraftStyleDefault/list {

DawnWright:
@ericbiewener I just took this fix for a spin, and noticed that it breaks copy-paste of lists from one editor into another: all lists become unordered. This is because Draft pastes from HTML, and with your changes `<ul>` is used to wrap both ordered and unordered lists. So when it's converted back to blocks, all the `<li>` elements are assumed to be unordered items.
https://github.com/facebook/draft-js/issues/1163

flarnie:
Thanks for making this PR! I like the idea of fixing #443 but would prefer the approach Isaac suggested, of just not resetting the depth when the block type changes from one list to another.

It doesn't make any visual difference to use an `ol` vs. `ul` but it has the potential to create inconsistencies like the one @DawnWright pointed out, and I'd also prefer we use semantic HTML when possible. That might also benefit things like screen readers, or styles folks have added to their editors on top of Draft.js.

Please feel free to open a PR with a different approach.

Issue #270, opened 2016-04-05
List item nesting not preserved  on dedent


Paul-Wray:
During list item indenting, draft ensures correct nesting by checking that an indented item will have a parent one level lower. But when an item is dedented, its descendants stay at the same level, giving children two or more levels deeper. The descendants should also be dedented I would think.

remko:
FYI, Google Docs and OpenOffice seem to allow independent indent/dedent of items, regardless of next/previous items. So the current dedent behavior of Draft is consistent with that, but the indent behavior isn't.

hellendag:
Yeah, this is something I remember spotting a while back.

I think Quip also allows free indenting.

I'd be fine with the Draft core also allowing this, since I don't think the framework needs to be opinionated about it.

For Draft-based editors at FB, I think we didn't want to allow arbitrary indentation. `RichUtils` could be the right place to do some enforcement here, since it's not required that clients use it.

PR #762, opened 2016-11-04
Allow indentation beyond one level deeper than block above.

ericbiewener:
**Summary**

Solves https://github.com/facebook/draft-js/issues/270. Removes restriction on indenting beyond one level deeper than the block above. Also allows indenting the very first item in a list. And all I had to do was delete code!

image

**Test Plan**

1. Try indenting the first item in a list.
2. Try indenting a list item more than one level deeper than the item above it.


flarnie:
Thanks for adding this! It seems fine to make this change, and I really like that we did it by only deleting code.

I'd like to have some tests to go with this, but that could be done in a follow-up.

Just like for any PR we want to merge, the first step is for me to pull it into Facebook's codebase. Since this presents a change in default behavior, I would like to do some manual testing and verify that we can maintain the restricted nesting for products where that is desirable. I don't expect problems there though.

juliankrispel:
@flarnie this still being merged I assume?

PR #766, opened 2016-11-05
Fix/richutils backspace

liqingfx:
**Summary**

This is a minor fix for #613. Handle backspace action when the cursor is in a atomic block.

**Test Plan**

use `convertFromHTML` api to render

```html
<figure><img src="1.jpg" /></figure>
```

Here is the result:

```html
<figure class="" data-block="true" data-editor="1pbbq" data-offset-key="7jm7f-0-0">
    <div data-offset-key="7jm7f-0-0" class="public-DraftStyleDefault-block public-DraftStyleDefault-ltr">
        <img src="1.jpg" />
    </div>
</figure>
```

If the cursor is after img tag but still in figure tag, backspace does not work properly. After applying this pr, the atomic block(figure in html) will be replaced with an empty unstyled block.

### DIFF
--- a/src/model/modifier/RichTextEditorUtil.js
+++ b/src/model/modifier/RichTextEditorUtil.js
@@ -13,6 +13,7 @@
 
 'use strict';
 
+const ContentBlock = require('ContentBlock');
 const DraftModifier = require('DraftModifier');
 const EditorState = require('EditorState');
 const SelectionState = require('SelectionState');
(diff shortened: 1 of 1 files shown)


flarnie:
Hi @leeching - thanks for making this PR. I think it makes sense, but I need more context on the problem which this is fixing. Can you give me more details, or even a jsfiddle to reproduce it? I understand the original fix in https://github.com/facebook/draft-js/pull/613/files but I'm not sure what you mean about "If the cursor is after img tag but still in figure tag", since the cursor doesn't literally go inside a figure tag.

I know this review is pretty late, so I can try to dig in more if I don't hear from you.

niveditc:
I'm closing this due to no response because we probably need more context + a rebase onto master. Please do comment/reopen if you're willing to fix this :)

PR #769, opened 2016-11-06
Unstyle empty list item on Enter key

ericbiewener:
**Summary**

This appears to be standard functionality in text editors. If the list item has no text and the user pressed `Enter`, rather than creating a new list item, the current one should simply be removed. This is how it works in Google Docs, TextEdit, etc.

**Test Plan**

1. Open the Rich Text editor example.
2. Create a list item. Don't add any text to it.
3. Press enter

The list item should be removed.


rylev:
@ericbiewener This change would be very useful. Any chance you can sign the CLA and we can get someone from Facebook to merge this?

ericbiewener:
@rylev I signed it a while ago. See the labels on the PR :)

rylev:
@ericbiewener sorry about that :-D Ok I guess we just need to sit back and wait for someone from Facebook to review this.

flarnie:
Very nice! 

As with https://github.com/facebook/draft-js/pull/762 I'd like to get a test for this, but that could be done as a follow-up. I can open an issue for the missing tests.

I'll see if we can get this merged. Great PRs! :D

ericbiewener:
@niveditc all set. Updated tests as well.

src/model/transaction/splitBlockInContentState.js
```diff
@@ -95,10 +96,22 @@ const splitBlockInContentState = (
...
 
   const key = selectionState.getAnchorKey();
-  const offset = selectionState.getAnchorOffset();
   const blockMap = contentState.getBlockMap();
   const blockToSplit = blockMap.get(key);
   const text = blockToSplit.getText();
+
+  if (!text) {
+    const blockType = blockToSplit.getType();
+    if (blockType && blockType.endsWith('-list-item')) {
```
niveditc: I think we should be specific about the two list item types:
`if (blockType === 'unordered-list-item' || blockType === 'ordered-list-item')`

ericbiewener: Agreed! All set.

niveditc:
It seems that package.json still shows up in this stack & didn't get deleted on d12d823

ericbiewener:
@niveditc Sorry, good to go for real this time :)

niveditc:
Next steps are:
* Provided the steps above are successful I'll merge it in.

PR #774, opened 2016-11-08
[documentation] Include reference to example for convertFromHTML

### LINKED ISSUE #236: convertFromHTML not documented

smtheard:
**Summary**
This doc page is the first place people find when looking for information on convertFromHTML but the original code sample is not very useful. This hopefully will give people a better idea of how to use the function and then a reference to where I found the code snippet.

Should close #236 as well.

### DIFF
--- a/docs/APIReference-Data-Conversion.md
+++ b/docs/APIReference-Data-Conversion.md
@@ -55,7 +55,11 @@ const sampleMarkup =
   '<img src="image.png" height="112" width="200" />';
 
 const blocksFromHTML = convertFromHTML(sampleMarkup);
-const state = ContentState.createFromBlockArray(blocksFromHTML);
+const state = ContentState.createFromBlockArray(
+  blocksFromHTML.contentBlocks,
+  blocksFromHTML.entityMap,
+);
+
 this.state = {
   editorState: EditorState.createWithContent(state),
 };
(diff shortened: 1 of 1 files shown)


docs/APIReference-Data-Conversion.md
```diff
@@ -49,11 +49,16 @@ other usage within an application.
...
-  DOMBuilder: Function = getSafeBodyFromHTML,
-  blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap
-): ?Array<ContentBlock>
+const sampleMarkup =
+  '<b>Bold text</b>, <i>Italic text</i><br/ ><br />' +
+  '<a href="http://www.facebook.com">Example link</a><br /><br/ >' +
+  '<img src="image.png" height="112" width="200" />';
+
+const blocksFromHTML = convertFromHTML(sampleMarkup);
+const state = ContentState.createFromBlockArray(blocksFromHTML);
```
davidchang: small nit, but since you're citing https://github.com/facebook/draft-js/blob/8ac72f723fb2d9102db833a9b060dfd66df65652/examples/convertFromHTML/convert.html#L61-L65 below, could you match it here to avoid confusion?

```
const blocksFromHTML = convertFromHTML(sampleMarkup);
  blocksFromHTML.contentBlocks,
  blocksFromHTML.entityMap,
);
```

smtheard:
@davidchang that's reasonable! I must have copy pasted from my own version instead of the example reference. Thank you, good catch.

davidchang:
wow, that was fast. thanks again for improving our docs - merging!

PR #783, opened 2016-11-14
Add default text direction

### LINKED ISSUE #634: Do not default to LTR for empty editors

Reporting a bug, I guess.

On an empty Draft.js field, the first block has a `public-DraftStyleDefault-ltr` class and the caret is on the left side. This is before the first character is typed. This default does not make sense for an RTL environment, where everything else is RTL except for the caret.

Not setting a direction when no content is available. It would default to the parent direction, which is a sensible default.

Draft 0.8.1, Chrome 52 and Safari 9.1.3, OS X 10.11.16. Didn’t check the older versions.

---

I’d be happy to work on a PR if you agree with the suggestion.

mostafah:
**Summary**

By default, Draft.js sets direction to LTR when text has no direction. In an RTL environment, it means that empty editors show the cursor on the wrong side, until the user starts typing some RTL text.

The new `defaultDirection` property of `EditorState` is used for that. It is sent to UnicodeBidiService as default direction.

**Test Plan**

A new tests is added for covering this change. For manual testing, you can set the `defaultDirection` of `EditorState` to `RTL` and expect to see the cursor on the right when the editor is empty.

Closes #634.

### DIFF
--- a/docs/APIReference-EditorState.md
+++ b/docs/APIReference-EditorState.md
@@ -136,6 +136,11 @@ The list below includes the most commonly used instance methods for `EditorState
     </a>
   </li>
   <li>
+    <a href="#defaultdirection">
+      <pre>defaultDirection</pre>
+    </a>
+  </li>
+  <li>
     <a href="#directionmap">
       <pre>directionMap</pre>
     </a>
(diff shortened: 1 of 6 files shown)


niveditc:
@mostafah - thanks for submitting this PR & apologies for the significant delay in reviewing this. Could you rebase on master? I think this is useful to have & we can merge it in :)

mostafah:
@niveditc Sorry for the delay in replying. I’ve not used Draft.js for a long time and I am currently occupied with a big project, so, unfortunately, I don’t have the time for this right now. Thanks for the interest in the change though.

PR #796, opened 2016-11-20
Add `editorKey` prop for SSR

### LINKED ISSUE #385: Rendering isomorphically

pofigizm:
**Summary**

We are getting the checksum errors while using Draft.js with SSR:
```
(client) "true" data-editor="2ec27" data-offset-k
(server) "true" data-editor="3vera" data-offset-k
```
This PR allows us to specify `editorKey` as a prop of  `<DraftEditor />`

**Test Plan**

It covered by unit tests

Closed #385


Pofigizm:
It looks like a magic! I can't understand it!
When I run locally `tests` and `flow` tasks I have got the same errors as in Travis.

But, if I change two devDependency, all tests and checks passed.

```
diff --git a/package.json b/package.json
index 5cd362c..5eee69b 100644
--- a/package.json
+++ b/package.json
@@ -64,8 +64,8 @@
     "gulp-uglify": "^1.2.0",
     "gulp-util": "^3.0.6",
     "jest": "^15.1.1",
-    "react": "^15.0.0-rc.1",
-    "react-dom": "^15.0.0-rc.1",
+    "react": "15.0.0-rc.1",
+    "react-dom": "15.0.0-rc.1",
     "run-sequence": "^1.1.2",
     "through2": "^2.0.1",
     "vinyl-buffer": "^1.0.0",
```

screen shot 2016-11-20 at 11 13 09

Why the `^` can change install process? (some parts of package will not be exist)
Can anyone explain it?

Pofigizm:
Now we have issue about the problem
#799

flarnie:
Hi @Pofigizm! Thanks for looking into this!

When Travis CI runs, it calls `npm update` which installs the latest allowed version of any dependencies listed in the `package.json`. The caret in the version number means it is allowed to grab any new version that is backwards compatible, so `15.*.*`. [Here is a longer explanation of semver versioning.](http://bytearcher.com/articles/semver-explained-why-theres-a-caret-in-my-package-json/)

It looks like something in [React version 15.4.0](https://github.com/facebook/react/blob/master/CHANGELOG.md#1540-november-16-2016) is causing this failure. For now we can specify a more specific dependency and I'll look into what is causing this version to fail.

I opened [an issue for this](https://github.com/facebook/draft-js/issues/799) - feel free to follow the progress there.

I'll comment on your PR separately - thanks for submitting it.

Pofigizm:
@flarnie thank for your explanation.
I knew that rule `^0.2.3 := >=0.2.3 <0.3.0` exist, but knew nothing about rule `^1.2.3 := >=1.2.3 <2.0.0`
It never happened with me to get same problem, because I never used carpet in my projects.
It's strange to use carpet, especially if we use private APIs.

gfx:
+1 I am about to create the same PR ;)

Pofigizm:
@flarnie
Are you going to merge the PR? May be another solution is required?

flarnie:
Extra thanks for the inclusion of a unit test, updating the props flowtype, and @nsfmc's new example! This makes sense and I plan on including it in an upcoming 0.10.1 release.

Since the example files were moved into subdirectories since this was made, we should probably move the 'universal' example into the same subdirectories as the other examples. We also need the copyright headers and 'use strict' in the new example.

I will make those tweaks quickly during the day today before merging this.

flarnie:
@nsfmc for some reason the ES6 polyfill isn't working in the universal example. I'm just removing destructuring for now and we can update it later, because that shouldn't block merging this PR.

Thanks again for making that example!

flarnie:
I'm trying to rebase the branch - seems like it didn't do what I expected. Will fix in a bit.

flarnie:
Thanks again to @nsfmc who talked through this PR with me in person!

cassln:
Great feature! When is planned release with that?

sponia-joker:
has it been released? I getting the same errors in version 0.10.1

EtaiG:
@pofigizm thank you for this :)

PR #798, opened 2016-11-21
Update failing example

sdelrio0:
**Summary**
This PR updates the ContentState.createFromBlockArray API example. 

**Description**
It seems that `convertFromHTML()` already returns the block array, which `ContentState.createFromBlockArray` is expecting.

**Changes**
Passing the following causes an error:
```javascript
  blocksFromHTML.contentBlocks,
  blocksFromHTML.entityMap
);
```

This works:
```javascript
const state = ContentState.createFromBlockArray(blocksFromHTML);
```

### DIFF
--- a/docs/APIReference-Data-Conversion.md
+++ b/docs/APIReference-Data-Conversion.md
@@ -55,10 +55,7 @@ const sampleMarkup =
   '<img src="image.png" height="112" width="200" />';
 
 const blocksFromHTML = convertFromHTML(sampleMarkup);
-const state = ContentState.createFromBlockArray(
-  blocksFromHTML.contentBlocks,
-  blocksFromHTML.entityMap,
-);
+const state = ContentState.createFromBlockArray(blocksFromHTML);
 
 this.state = {
   editorState: EditorState.createWithContent(state),


davidchang:
Sorry for the late response @sdelrio0 - examples were recently updated into an 0.9.1 directory and an 0.10.0 directory. Could you resolve merge conflicts if still applicable?

flarnie:
This looks good, and indeed does need rebased. I can try to rebase it and push an update, then this will be good to land. I'm doing a batch up documentation updates for 0.10.0 so I'll land it with those.

flarnie:
Aha - oddly enough, I glanced at this and thought it was making the reverse change. The documentation should have the following syntax:
```
  const state = ContentState.createFromBlockArray(blocksFromHTML);
   blocksFromHTML.contentBlocks,		
   blocksFromHTML.entityMap,		
 );
```

But instead it currently has
```
const state = ContentState.createFromBlockArray(blocksFromHTML);
```

Sorry for the confusion. In any case, there is a separate PR coming to update the docs for the 0.10.0 API changes, and in the future we'll respond more quickly to documentation updates like this.

PR #803, opened 2016-11-23
Fix advanced topics custom block render docs: Remove extra line in de…

StGeass:
**Summary**

Remove extra line from [default block render map](https://facebook.github.io/draft-js/docs/advanced-topics-custom-block-render-map.html#content) list (copy-past?)

screenshot


flarnie:
Thanks for fixing this!

PR #814, opened 2016-11-28
Part two deprecate but dont delete old entity api

flarnie:
We are deprecating the old DraftEntity API and this adds warnings for folks still using the outdated syntax.

Why is the API changing? Because
 - currently changing or adding entities won't trigger a re-render
 - the use of a non-Immutable storage for entities doesn't fit well with
   the rest of the Draft architecture
 - moves Entity storage out of global module; this just makes more sense
   as an API, rather than having it globally available.

This moves the logic for the old API into private methods, and adds
warnings and comments around the public-but-deprecated API methods.

We then use the private methods internally while we still support both
APIs.

A future version of Draft will actually remove the old API completely,
and these warnings and comments will help folks make updates to their
code to prepare.

We will release more documentation about this soon.

This PR also updates syntax in one example, since the warnings made it clear that the API had not been updated there yet.

### DIFF
--- a/examples/draft-0-10-0/tex/js/modifiers/insertTeXBlock.js
+++ b/examples/draft-0-10-0/tex/js/modifiers/insertTeXBlock.js
@@ -16,7 +16,7 @@
 
 import {
   AtomicBlockUtils,
-  Entity,
+  EditorState,
 } from 'draft-js';
 
 let count = 0;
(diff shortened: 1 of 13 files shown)


flarnie:
Tagging @johanneslumpe - and also thanks for your comments on the earlier PR!
Also tagging @stopachka, if you get a chance this week it would be great to get your thoughts on this direction.
Also @spicyj I'd be curious what you think.

johanneslumpe:
@flarnie LGTM! This will become 0.10 and then soon after 0.11 will be released with the updated api?

flarnie:
@johanneslumpe Yup! That's the plan. I think just two more issues should be fixed before releasing v0.10.0 - https://github.com/facebook/draft-js/issues/816 and https://github.com/facebook/draft-js/issues/815

flarnie:
Going to rebase and merge this after #838 lands. I think this will wrap up the changes we want to include in 0.10.0.

Issue #693, opened 2016-09-29
 How to detect unicode input event?


bowbowbow:
handleBeforeInput handler(https://facebook.github.io/draft-js/docs/api-reference-editor.html#handlebeforeinput) is not called when unicode like Korean is inserted.

davidchang:
Thanks for opening the bug, @bowbowbow - I'm going to close this as a duplicate of https://github.com/facebook/draft-js/issues/311

Issue #631, opened 2016-09-04
HandleBeforeInput not working when typing in Japanese.


matiosfree:
**HandleBeforeInput** not working when typing in Japanese. For other languages (En, Ru) it works fine.

Used DraftJS v.0.7.0, Firefox 48.0.2\Chrome 52.0.2743.116, Win7\Darwin

`
handleBeforeInput: function (chars) {
    console.log('Checking handleBeforeInput...', chars);
    return true;
},

render() {
        const {editorState} = this.state;
        return (
            <div>
                <Editor
                        editorState={editorState}
                        onChange={this.onChange}
                        handleBeforeInput={this.handleBeforeInput}
                        spellCheck
                        ref="editor"/>
            </div>
        );
    }
`

stopachka:
hey @matiosfree 

Could you tell us more about what you are trying to do? You have indeed come across a tricky issue -- exposing handleBeforeInput during composition could be tricky, because the DOM is in an uncertain state.

Based on what you're trying to do, maybe we can come up with some solutions.

matiosfree:
@stopachka When I type in Japanese I need to make some manipulations with the DOM.

I solved it like that:
        $('.DraftEditor-root').on({
            compositionupdate: function(evt) {
                self.handleBeforeInput(evt.originalEvent.data);
            }
        });

but, **onChange** method doesn't fire for each character. How can I get actual **editorState**?
this.state.editorState is not suitable for me, because in this state characters are missing.

I want to do like this:
       $('.DraftEditor-root').on({
            compositionupdate: function(evt) {
                let editorState; //?????
                self.handleBeforeInput(evt.originalEvent.data);
                self.onChange(editorState);
            }
        });

doortts:
This problem also happens with Korean. I think CJK inputs have same problems.

flarnie:
@matiosfree Thanks for reporting this! Because Japanese, Korean, and other languages use IME for input there are many differences in how Draft handles the input.

I commented on the PR that addresses this - I think it's on the right track and hope we can resolve this issue soon. IME entry is probably the main area where we'd like to improve Draft.js. :)

For what it's worth, you will be able to get the latest `editorState` as a parameter of `handleBeforeInput` once we do the next patch release. I don't think that will solve the problem here though. https://github.com/facebook/draft-js/blob/6f2bf66bacff3ffac234fc21ff3720344ff92776/src/component/handlers/edit/editOnBeforeInput.js#L101

[...]

PR #821, opened 2016-11-29
Fix handleBeforeInput not working problem when typing in IME

### LINKED ISSUE #819: handleBeforeInput is not trigger when input using an IME
*feature*

`handleBeforeInput` is not triggered when using an IME. I wonder if it's possible to add some code to trigger something act like `handleBeforeInput` in `resolveComposition`.

bowbowbow:
**Summary**

[handleBeforeInput](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handlebeforeinput), one of Editor's props, doesn't work when typing in [IME](https://en.wikipedia.org/wiki/Input_method) like Japanese and korean. 

`Related Issues`
1. handleBeforeInput is not trigger when input using an IME #819 
2. How to detect unicode input event? #693 
3. HandleBeforeInput doesn't work when typing in Japanese. #631

The cause of this problem is that `this.props.handleBeforeInput` is used in `edit` mode but it is not used in `composite` mode.

More specifically, `editOnBeforeInput`(onBeforeInput handler) in `/src/handler/edit/editOnBeforeInput.js` use `this.props.handleBeforeInput`, but `onBeforeInput` in `/src/handler/composition/DraftEditorCompositionHandler.js` doesn't use it.

I solved this problem by adding the `edit` mode logic to the `composite` mode.

close #819, #631 

**Test Plan**

- Add `handelBeforeInput` to Editor's props of example.
```javascript
_handleBeforeInput(chars) {
  console.log(':: handleBeforeInput :: ', chars);
  return false; // or return true;
}
```
- Typing Korean and Japanese.
- Check `handleBeforeInput` working correctly

### DIFF
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -18,6 +18,7 @@ const Keys = require('Keys');
 
 const getEntityKeyForSelection = require('getEntityKeyForSelection');
 const isSelectionAtLeafStart = require('isSelectionAtLeafStart');
+const isEventHandled = require('isEventHandled');
 
 /**
  * Millisecond delay to allow `compositionstart` to fire again upon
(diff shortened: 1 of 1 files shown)


flarnie:
Hi @bowbowbow! Thanks for submitting this. I think the general idea is good, and I'm wondering if we should have a different handler than `handleBeforeInput`. The intention is to let users hook into the firing of the `beforeInput` event when it fires, and the problem is that it does not fire when typing in composition mode.

This is just an idea off the top of my head - would it help to add a handler for the `compositionUpdate` event? It seems like folks want to hook into the moment that a new character is entered by the IME, and we only get the data about that from the `compositionUpdate` event.

This debugging tool is a good way to verify what events get fired - https://dvcs.w3.org/hg/d4e/raw-file/tip/key-event-test.html

screen shot 2017-04-04 at 7 10 08 am

PR #822, opened 2016-11-29
Defer adding editorKey until after initial render

nsfmc:
**Summary**

currently, the generation of `_editorKey` during construction causes serverside rendering to break. This PR changes the behavior so that the `_editorKey` is *either* passed in as a `prop` or generated dynamically during `componentDidMount` if no prop was provided.

**Test Plan**

Included is a minimal universal/iso example which breaks before adding the subsequent commit. Upon building draft-js with the second commit, the universal rendering works if an `editorKey` prop is passed, or not, or whatever.

A slight gotcha here: `EditorState.createEmpty()` will create mismatched content states between the server rendered editor and the browser rendered one. For this reason, the example uses a dummy empty rawDraftContentState rather than using the other examples' approach of using `createEmpty()`

I'm 100% ok if this is the wrong approach or not good to merge, i'm opening it as an rfc mostly to get feedback or to understand if this is not a tenable approach to getting universal rendering unbroken with draft.


Pofigizm:
It's similar to #796 ))

nsfmc:
@flarnie this can probably be closed. it's similar to #796 but keys blocks on mount via `setState` to maintain compat with the accessibility stuff. not sure what the status on the other pr is, but it seems to have greater mindshare and has probably been updated more frequently than this (as well as having inherited this pr's tests 😛 )

flarnie:
Thanks again - merged https://github.com/facebook/draft-js/pull/796 so closing this.

Issue #816, opened 2016-11-28
Show examples of deprecated *and* updated API for entities


flarnie:
**Problem:**

The examples included on the current master use the updated-but-undocumented API for Entity state.

We are in the process of releasing a version of Draft that supports both, which will be followed by a version which only supports the new API.

In the mean time, having 'examples' that don't match the current version or documentation is creating confusion. See issues #805, #794, and #778.

**Solution:**
1. Comment here and tell us you are working on this - otherwise I'll do it sometime soon.
2. Create two directories under 'examples' - 'draft-0-9-1' and 'draft-0-10-0'
3. Move the examples present on the latest 'master' branch into the 'draft-0-10-0' directory.
4. Copy [the examples that were present in version 0.9.0](https://github.com/facebook/draft-js/tree/4e4ad9805d3e112ae30df20becbd3f0ba0de5683/examples) into the 'draft-0-9-1' directory.

Note that some examples may currently be buggy - see issue #815.

tobiasandersen:
I've created a PR for this now, see #824.

flarnie:
Thanks again for fixing this @tobiasandersen

PR #824, opened 2016-11-30
Add deprecated examples

### LINKED ISSUE #815: 'color' and 'rich' examples failing on master
bug

On current master in Chrome and Firefox, for me, I get the following error in both 'color' and 'rich' examples:
```
Uncaught TypeError: Cannot read property 'purgeUnmountedComponents' of undefined(…)
```

1) Clone master
2) `npm update && npm run build`
3) Open either `examples/rich/rich.html` or `examples/color/color.html`

They should render the examples.

Draft un-released latest, in Chrome and Firefox and likely other browsers too.

tobiasandersen:
**Summary**

This is mainly a PR for #816. It moves the current examples into a new  'draft-0-10-0' directory and re-adds the deprecated examples into a new 'draft-0-9-1' directory. Since the examples are now located one level deeper than before, the paths to 'node_modules' and 'dist' have been updated. 

It also fixes #815 by not using the minified version of React. So all examples should be working properly now.


flarnie:
Thanks for doing this! It looks like the website actually uses the Rich Text Editor example css, so we need to update the path for that. That's why the CI build failed.
https://github.com/facebook/draft-js/blob/69890ed57b5f4256cc93ae8e9ea178296851a23e/website/server/convert.js#L113

Similarly, the documentation links to the examples, so those paths should be updated. 

Lastly - in the `/tex` examples the path in the local `package.json` needed to be updated.

I'll see if I can edit this PR and add those tweaks, and then this should be good to land. If that doesn't work then I'll ask you to revise it.

It's great to see this happening - glad to have your help Tobias.

tobiasandersen:
Oh I didn't think about that, thanks! Just let me know if there's anything else needed.

PR #826, opened 2016-11-30
Add DraftDecoratorStrategy type

nkt:
This change should help projects which uses draft-js with flow


flarnie:
Sweet - thanks! I'll land this concurrent with the 0.10.0 documentation updates, since the 'contentState' argument only gets passed there in v0.10.0.

PR #827, opened 2016-11-30
Overwrite correct block when handling spellcheck/autocorrect changes

Kofia5:
- dom selection and draft selection can be out of sync when the browser updates
- currently we translate dom selection to find the content block being changed but *don't* use that translation to figure out where to apply the change
- this PR forces the new editor selection to match the dom selection being used for the change (ensuring that the changes are targeted to the correct block)


gmertk:
@Kofia5 any chance you can rebase this onto the latest version? I think this solves the issue #2141 

Not sure why this wasn't merged back in time. I believe we can merge it now if it doesn't break anything else.

Let me know please if you can do the rebase. Then I'll make sure that this PR gets some love.

PR #828, opened 2016-12-01
Fix line breaks when pasting from word to draft

mitermayer:
**Summary**

When pasting adjacent paragraphs were converged into a single paragraph, by introducing the idea of alias for an element we can then keep there rendering elements at the same time as provide different element matches.

This issue is also referenced at: https://github.com/facebook/draft-js/issues/523 and https://github.com/facebook/draft-js/issues/395

**Test Plan**

We can test by running `npm test` or by pasting content from a word processing tool such as word into draft. Paragraphs should be now treated as a separate blocks and should no longer get converged into a single one.

### DIFF
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -97,6 +97,19 @@ class RichEditor extends React.Component {
 }
 ```
 
+There are cases where you want to define an alias for block elements.
+For example we use this internally on the `unstyled` block type
+in order to allow it to also match paragraphs
+
+*example of unstyled block type alias usage*
+
+```
+'unstyled': {
+  element: 'div',
+  aliasedElements: ['p'],
+}
+```
+
 ## Custom block wrappers
 
 By default the html element is used to wrap block types however a react component
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -97,11 +97,11 @@ class RichEditor extends React.Component {
 }
 ```
 
-There are cases where you want to define an alias for block elements.
-For example we use this internally on the `unstyled` block type
-in order to allow it to also match paragraphs
+When Draft parses pasted HTML, it maps from HTML elements back into
+Draft block types. If you want to specify other HTML elements that map to a
+particular block type, you can add an array `aliasedElements` to the block config.
 
-*example of unstyled block type alias usage*
+*example of unstyled block type alias usage:*
 
 ```
 'unstyled': {
--- a/src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
+++ b/src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @emails isaac, oncall+ui_infra
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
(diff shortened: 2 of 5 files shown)


src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @emails isaac, oncall+ui_infra
```
spicyj: you can remove isaac :'(

src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
+const convertFromHTMLToContentBlocks = require('convertFromHTMLToContentBlocks');
+const DefaultDraftBlockRenderMap = require('DefaultDraftBlockRenderMap');
```
spicyj: should be requires starting with a capital letter, then a blank line, then requires starting with a lowercase letter

src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -0,0 +1,51 @@
...
+      <${tag}>b</${tag}>
+    `;
+
+    const blocks = convertFromHTMLToContentBlocks(html_string);
+
+    expect(blocks.contentBlocks.length).toBe(2);
+  });
+}
+
+function performTestWithDefaultDraftBlockTags(
```
spicyj: This test is pretty hard to read. It would probably be clearer if you just hardcoded a couple HTML examples.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -155,9 +157,16 @@ function getBlockMapSupportedTags(
 ): Array<string> {
   const unstyledElement = blockRenderMap.get('unstyled').element;
   return blockRenderMap
-    .map((config) => config.element)
-    .valueSeq()
-    .toSet()
+    .reduce((accum, draftBlock, key) => {
+      if (draftBlock.aliasedElements) {
```
spicyj: This is really hard to read. Can you just do

```
var elements = [draftBlock.element];
if (draftBlock.aliasedElements) {
  elements.push(...draftBlock.aliasedElements);
}
```

in a `map` and then flatten?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -155,9 +157,16 @@ function getBlockMapSupportedTags(
...
+        draftBlock.aliasedElements.forEach(
+          (alias) => {
```
mitermayer: will inline this

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -97,6 +97,19 @@ class RichEditor extends React.Component {
 }
 ```
 
+There are cases where you want to define an alias for block elements.
+For example we use this internally on the `unstyled` block type
+in order to allow it to also match paragraphs
+
+*example of unstyled block type alias usage*
```
spicyj: add a colon after `usage`

docs/Advanced-Topics-Custom-Block-Render.md
```diff
@@ -97,6 +97,19 @@ class RichEditor extends React.Component {
 }
 ```
 
+There are cases where you want to define an alias for block elements.
```
spicyj: Let's replace this with:

> When Draft parses pasted HTML, it maps from HTML elements back into Draft block types. If you want to specify other HTML elements that map to a particular block type, you can an array `aliasedElements` to the block config.

mitermayer: > you can an array aliasedElements

or 
> you can add an array aliasedElements

Issue #809, opened 2016-11-24
convertFromHTML rejects error Cannot read property 'map' of undefined(…)


optimatex:
**Do you want to request a *feature* or report a *bug*?**
report a bug
**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
I get code from this docs https://facebook.github.io/draft-js/docs/api-reference-data-conversion.html#content and try to implement it but got `TypeError: Cannot read property 'map' of undefined(…)`

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
versions of Draft.js draft-js@0.9.1 
browser=chrome  OS=windows 8.

oscarmorrison:
+1

saymy:
+1

stagfoo:
+1

tobiasandersen:
That example is actually for a future release and won't work with version 0.9.1 (latest on npm). It's a bit misleading that this is on the docs already. I don't think that's intentional? (cc @flarnie)

I just created a PR with [an example for 0.9.1](https://github.com/tobiasandersen/draft-js/blob/1c7499baeef5866af95481c48aa5db177762e0aa/examples/draft-0-9-1/convertFromHTML/convert.html). The main difference is that you should do this:
```
const blocksFromHTML = convertFromHTML(sampleMarkup);
const state = ContentState.createFromBlockArray(blocksFromHTML);
```
rather than:
```
const blocksFromHTML = convertFromHTML(sampleMarkup);
const state = ContentState.createFromBlockArray(
  blocksFromHTML.contentBlocks,
  blocksFromHTML.entityMap,
);
```



Support for converting images was also added recently, so that won't work for 0.9.1, but the rest should.

flarnie:
Thanks @tobiasandersen for updating the example and documentation.

emorris:
https://draftjs.org/docs/api-reference-data-conversion/  4 years and the example is the same

PR #833, opened 2016-12-02
Add convertToHTML example for version 0.9.1

tobiasandersen:
**Summary**
This PR adds a convertToHTML example for version 0.9.1 (only present for 0.10.0 currently). In it, I've removed everything that has to do with images (since that's not supported in 0.9.1), and updated the link decorator to use the old API for entities. Note that this example won't work on master (because of the change to entities). Maybe we should add some information about this somewhere?

This should also fix the broken link in the bottom of this document: https://github.com/facebook/draft-js/blob/master/docs/APIReference-Data-Conversion.md#convertfromhtml. It's currently pointing to a file that doesn't exist (just noticed this now).

I've also updated that document to use the API for version 0.9.1, since it was creating confusion (see #809). Let me know if this isn't correct (or just remove it)!


flarnie:
Thanks for fixing this! 😄  
I'll push an update to that link you mentioned and then land this.

tobiasandersen:
Great! There seems to be a bit of a confusion/worry, both here on the issue tracker and in the slack channel, about these changes. I think it would be a good idea to clarify what's going on. Some questions I've seen come up are e.g:

- Is there any ETA for v0.10.0?
- What will happen when v0.10.0 is released? Will it be breaking?
- Why are we doing this? 
- What's going on with the examples? 
- What should I do in the meantime?

I can't answer all of these questions (e.g. I don't know when the next release will be). And I'm not really sure where this info would make sense either. In the docs somewhere? In the README?

flarnie:
Thanks @tobiasandersen - just made an issue for answering some of those questions. https://github.com/facebook/draft-js/issues/839

PR #851, opened 2016-12-07
Pass `block` to `customStyleFn` callback

DawnWright:
Recently, `customStyleFn` was added for more flexibility to specify styles: https://github.com/facebook/draft-js/issues/342 / https://github.com/facebook/draft-js/pull/354

This PR proposes an addition to that same API, to allow the block type to be taken into consideration, as well. Block type is important for cases where, for example, bold requires a different CSS object result for different font families:

```
const customStyleFn = (style, block) => {
  const isBold = style.has('BOLD')
  if (isBold) {
    if (block.getType() === 'code-block') {
      return {
        // fontFamily remains as normal (monospace)
        fontWeight: '700',
      }
    } else {
      return {
        // fontFamily changes from normal to demi
        fontFamily: 'MyFont-Demi',
        fontWeight: '700',
      }
    }
  }

  return {}
}
```

By adding `block` to the callback, it opens avenues for more flexible styling.

Before:
`customStyleFn?: (style: DraftInlineStyle) => ?Object`

After:
`customStyleFn?: (style: DraftInlineStyle, block: ContentBlock) => ?Object`

- [x] Updated the documentation for API.
- [x] The test suite passes (`npm test`)
- [x] Code lints (`npm run lint`) and passes Flow (`npm run flow`)
- [x] Completed the corporate [CLA](https://code.facebook.com/cla).

cc @hellendag, @davidbyttow

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -85,7 +85,7 @@ for details on usage.
 
 #### customStyleFn
 ```
-customStyleFn?: (style: DraftInlineStyle) => ?Object
+customStyleFn?: (style: DraftInlineStyle, block: ContentBlock) => ?Object
 ```
 Optionally define a function to transform inline styles to CSS objects that are applied
 to spans of text. See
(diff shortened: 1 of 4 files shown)


hellendag:
This seems totally reasonable to me. Thanks!

PR #856, opened 2016-12-08
Fix lints and flow errors for v0.10

flarnie:
This is a collection of flow and lint fixes that came up when testing the 0.10.0-stable branch internally.

I'm not completely sure why they don't show up when we run eslint and flow locally or in CI.

Hopefully nothing controversial, and in a couple of cases they are short-term fixes to accomodate supporting two APIs until we release 0.11.0. More info in each commit message.


flarnie:
Going to land this so we can keep moving - ideally will come back and deal with linting inconsistencies later.

PR #858, opened 2016-12-09
Change underscore method names to bypass lint about private methods

flarnie:
Private methods should not be called as public methods. A recent PR
(https://github.com/facebook/draft-js/pull/814) added warnings to the
public-but-deprecated API of DraftEntity, and then moved the logic for
those methods into private methods that can be called within the
Draft.js framework.

These methods are "private" in the sense that we are only calling them
within Draft. They should not be used outside of the Draft library,
because they will be removed in an upcoming version of the library.

We have a transform step that causes issues when an underscore-prefixed
method is called outside of it's module, so changing the method names
seems like the easiest way to deal with this issue and still send the
message that they are private.

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -408,7 +408,7 @@ function genFragment(
     node.textContent = imageURI; // Output src if no decorator
 
     // TODO: update this when we remove DraftEntity entirely
-    inEntity = DraftEntity._create(
+    inEntity = DraftEntity.__create(
       'IMAGE',
       'MUTABLE',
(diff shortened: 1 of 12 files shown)


spicyj:
Do you want to just add `@preventMunge` to DraftEntity instead?

flarnie:
Yes! I will just do that. Was not finding documentation about that transform and how to turn it off.

flarnie:
It ended up being easier to change the method names after all - seems like we would have to add `@preventMunge` to every file where we use Draft Entities.

spicyj:
Only the files that use ES6 classes and call private methods of DraftEntity from those classes, but yes.

PR #860, opened 2016-12-09
Fix 0.10 compatibility break in getDecorations

### LINKED ISSUE #859: 0.10 breaks existing decorators
Earlier in the development of 0.10 #376 broke backwards compatibility with 0.9 in a number of ways. #790 corrected many of those breaks but not a change to the `getDecorations` method on decorators. This appears to be an oversight and that no break was intended.

colinjeanne:
This change fixes #859 by reversing the order of the arguments to `getDecorations`. Existing decorators will continue to function as they had in 0.9 since they can safely ignore the new `ContentState` parameter. Without this change existing decorators like `MultiDecorator` in `draft-js-plugins` will need to be broken making it more difficult to migrate to 0.10 for existing users of `draft-js-plugins`.

The backwards compatibility break appears to be unintentional as the other breaks introduced by #376 were fixed by #790. This closes a break missed by #790.

### DIFF
--- a/src/model/decorators/CompositeDraftDecorator.js
+++ b/src/model/decorators/CompositeDraftDecorator.js
@@ -52,7 +52,7 @@ class CompositeDraftDecorator {
     this._decorators = decorators.slice();
   }
 
-  getDecorations(contentState: ContentState, block: ContentBlock): List<?string> {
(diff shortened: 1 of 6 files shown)


flarnie:
Thanks for catching this! lgtm - you are correct that the break was unintentional, and we want to maintain backwards compatibility for the next release.

PR #863, opened 2016-12-10
Update CSS Constants

flarnie:
Facebook has renamed these; just keeping things in sync.


flarnie:
This got out of sync because of an internal codemod - it would be nice if we could ping the oncall or maintainer of Draft when future codemods touch our version of Draft.

spicyj:
I set up a Herald rule for me a month or two ago because I saw this – you can too.

lgtm

PR #864, opened 2016-12-12
Fixes weird scroll position issues with Asian IMEs on. 

goodhyun:
Asian IME sometimes makes draft-js editor scrolls to an unintentional position.

The `restoreEditorDOM` function was called to force a re-render when the character composition is being done at the first position of a leaf, etc. However, not like with the case of cut action, `scrollPosition` is not passed, and `restoreEditorDOM` might need a fallback to figure out the scroll position.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -388,7 +388,11 @@ class DraftEditor extends React.Component {
    * state (cut command, IME) and we want to make sure that reconciliation
    * occurs on a version of the DOM that is synchronized with our EditorState.
    */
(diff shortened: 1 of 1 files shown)


hellendag:
Interesting! Can you provide some test plan details, especially OS/locale/browser permutations you tried out? This will need to be tested extensively for FB use.

goodhyun:
You can reproduce the symptom in a separately scrollable draft-js editor instance within a div. 
We built an app with an electron(chromium) on macOS/Windows and tested it using all CJKs, and you will see this anomaly when you are editing an obviously long articles. 

Happy new year.

grrowl:
👍  this is a great fix and is working for us. Any chance of some merge attention for this small fix ?

PR #867, opened 2016-12-13
Fix array lookup to .includes rather than .has

drsirmrpresidentfathercharles:
**Summary**

In the CharacterMeta file, to lookup if a certain CharacterMeta had a style, we were using .has, which looks up keys in an array, while something like 'ITALIC' is stored as a value, so Immutable's .includes function would be more appropriate here.

### DIFF
--- a/src/model/immutable/CharacterMetadata.js
+++ b/src/model/immutable/CharacterMetadata.js
@@ -49,7 +49,7 @@ class CharacterMetadata extends CharacterMetadataRecord {
   }
 
   hasStyle(style: string): boolean {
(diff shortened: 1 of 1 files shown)


hellendag:
Because `DraftInlineStyle` is an `OrderedSet<string>`, it doesn't really matter which one it uses -- both are supported. I would also not be surprised if they have the same performance, or even share the same underlying implementation, though it's unlikely to matter for the typical set size here anyway.

That said, `includes` does make a bit more sense semantically, so I'm cool with this. :)

PR #871, opened 2016-12-14
Improved version of "Fix modifying state during onChange ... #667"

flarnie:
This fixes some issues with the fix originally merged from PR #667

Credit to @spicyj  and @srmcconomy for this fix. 
For now I'm just getting 'www' and github back in sync.

We also had to tweak this compared to the current internal version to
accomodate a recent change that makes 'editOnBeforeInput' and the other
handlers take 'editor' as an explicit argument instead of assumming
'this' will be the editor;
https://github.com/facebook/draft-js/commit/e64c2c3f2a3654b925b18baff448100cfbbf3738

**Update:**

**Also thanks to @davidchang for pinging me about landing this, @colinjeane for reporting this bug in #917 and submitting a nearly identical fix: https://github.com/facebook/draft-js/pull/919**

The delay on landing this was until 'setImmediate' was available in fbjs.[1]
[1]: https://github.com/facebook/fbjs/pull/208

Note that we are updating the version of 'fbjs' required in order to get access to the 'setImmediate' polyfill.

### DIFF
--- a/src/component/handlers/edit/editOnSelect.js
+++ b/src/component/handlers/edit/editOnSelect.js
@@ -20,10 +20,8 @@ var getDraftEditorSelection = require('getDraftEditorSelection');
 import type DraftEditor from 'DraftEditor.react';
 
 function editOnSelect(editor: DraftEditor): void {
-  if (editor._blockSelectEvents) {
-    return;
-  }
-  if (editor._latestEditorState !== editor.props.editorState) {
+  if (editor._blockSelectEvents ||
+      editor._latestEditorState !== editor.props.editorState) {
     return;
   }
 
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-ci": "NODE_ENV=test npm run lint && npm run flow && npm run test"
   },
   "dependencies": {
-    "fbjs": "^0.8.3",
+    "fbjs": "^0.8.7",
     "immutable": "~3.7.4",
     "object-assign": "^4.1.0"
   },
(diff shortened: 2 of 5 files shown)


spicyj:
lgtm. In particular, onInput does not fire in IE so we need the setImmediate.

flarnie:
davidchang also submitted this fix - https://github.com/facebook/draft-js/pull/921
 - if he updates and lands I'll close this, otherwise will update this soon.

src/component/handlers/edit/editOnSelect.js
```diff
@@ -20,10 +20,8 @@ var getDraftEditorSelection = require('getDraftEditorSelection');
 import type DraftEditor from 'DraftEditor.react';
 
 function editOnSelect(editor: DraftEditor): void {
-  if (editor._blockSelectEvents) {
-    return;
-  }
-  if (editor._latestEditorState !== editor.props.editorState) {
+  if (editor._blockSelectEvents ||
```
colinjeanne: Checking these two conditions is insufficient; I observe this event handler called when pressing two different keys in quick succession which is why I added an extra check in #919.

colinjeanne:
I wish I had seen this before, I had written nearly the exact same change in #919 with the major difference being that I call `editOnInput` directly to ensure all browsers follow the same code path.

If there isn't a reason that IE should go through `editOnInput` and if you prefer I can close my PR in preference for this one.

src/component/handlers/edit/editOnSelect.js
flarnie: I'll look into this more before landing - thanks for flagging.

flarnie: Actually - going to land this first to get Github in sync with our internal version, then will make a separate PR to update the conditional.

Issue #570, opened 2016-07-28
[bug] Incorrect behavior of getRangeClientRects in webkit browsers


qrohlf:
video demo: https://www.youtube.com/watch?v=TpNzVH5jlcU
live demo: http://codepen.io/qrohlf/pen/YWjzOq?editors=1010#0

getVisibleSelectionRect does not fully work around webkit's bad behavior regarding selection range bounding boxes. Selecting the first word on a wrapped line results in an incorrect bounding box. Bug exists in Chrome 51.0.2704.106 and Safari 9.1.1 (11601.6.17). 

You can see the bad behavior this causes in the inline toolbar in Facebook Notes:

_normal behavior_
![screen shot 2016-07-28 at 4 19 37 pm](https://cloud.githubusercontent.com/assets/347189/17232759/388e2c00-54df-11e6-80f8-bcc05dfe609f.png)

_oh look, it's the webkit contenteditable selection range bounding box bug (WCESRBBB)_
![screen shot 2016-07-28 at 4 19 44 pm](https://cloud.githubusercontent.com/assets/347189/17232773/4f038f70-54df-11e6-8833-d09960eeedcc.png)

Hopefully once I finish the other functionality in the project I'm using draft for, I can circle back around and fix this bug, but if someone else wants to fix it for me I won't argue :).

ouchxp:
We have this problem as well. Here is my monkey patch.
```
function getVisibleSelectionRect(window) {
  let selection = draftVSR(window);

  // When selection the first work of the wrapped line, chrome will return two rects
  // And the first rect has 0 width
  if (selection) {
    const selectionRects = window.getSelection().getRangeAt(0).getClientRects();
    if (selectionRects.length === 2 && selectionRects[0].width === 0) {
      selection = selectionRects[1];
    }
  }
  return selection;
}
```

Issue #290, opened 2016-04-10
Incorrect values with getVisibleSelectionRect under certain selection circumstances?


deep-c:
Hi,

I am attempting to create a floating toolbar which shows on text selection. I read through a similar issue #45 and noted the use of getVisibleSelectionRect(). This works for the most part but as you can see in the video below there are a couple of strange issues I've encountered.

https://www.youtube.com/watch?v=eHTj7F62S7g

The toolbar is included as a seperate component to the `Editor`.
### TextEditor.js

```
      <div styleName="container">
          <Toolbar
            editorState={editorState}
            onChange={this.onEditorChange}
            boundingRect={boundingRect} />
          <Editor
            editorState={editorState}
            onChange={this.onEditorChange}
            customStyleMap={styleMap}
            placeholder="Enter your text.." />
      </div>
```

The condition ive used to hide and show the toolbar is this:
### Toolbar.js

```
const boundingRect = getVisibleSelectionRect(window); (passed from parent component)
const selectionState = editorState.getSelection();

if (boundingRect && !selectionState.isCollapsed() && selectionState.getHasFocus()) {
      var toolbar =
        <div
          styleName="container"
          style={{top: boundingRect.top, left: boundingRect.left + boundingRect.width/2}}>
            ...buttons
        </div>
}

//which is returned in render
    return (
      <span>
        {toolbar}
      </span>
    )
```

```
.container {
  position: fixed;
  transform: translate3d(-50%, calc(-100% - 5px), 0);
  max-width: 300px;
  z-index: 80;
}
```
### Environment

Ive tested on Chrome on both my window and mac machines which show identical issues. I'll test the same on firefox and safari and post my findings here later today.
### Issue 1 (can be seen in the vid from ~30s to end)

Selection of text from the start/end of any line except the first/last (i.e last word) results in the width value returned from the boundingRect being larger than the selection (as you can in the vid the toolbar seems to hover in the middle of the line which seems like the width being returned is that of the entire line instead of the just the selection? 
Ive noticed that this is sometimes intermittent on the last word of a line and might have something to do with a space character being included as part of double clicking a specific word. As for why it happens on the first word of a line (apart from first) I havent noticed anything concrete.
### Issue 2 (can be seen in the vid from 0 - ~ 30s)

[...]

PR #876, opened 2016-12-16
Ignores 0 width rects when calculating selection

vierno:
**Summary**

This patch fixes getVisibleSelectionRect for webkit based browsers. The bad behavior is described in the issues #570 and #290.

**Test Plan**

In Chrome, selecting a wrapped line will return top values higher than expected, after applying this patch, the value should match the visual selection.


hellendag:
Seems about right to me. Thanks!

Issue #607, opened 2016-08-21
console warning reported by immutable


kinglywork:
Because of the current version (3.7.4) of immutable in dependencies, there will be warning in console: "Iterable.length has been deprecated, use iterable.size or iterable.count()".

this warning is removed in immutable@3.8.0

could you please update the immutable version in your dependencies.

kinglywork:
The warning problem will cause serious performance issues in IE(tested in IE11), please consider to fix it.

alfonsoperez:
@kinglywork it does for me in Chrome as well, for now I have monkey patched `console.warn`, I wish I had the bandwidth to open up a PR.

hardywu:
Just a reminder to `yarn` users. 

In `yarn` configuration, updating immutable.js does not fix this syntax issue. Because `yarn` install a nested `immutable.js` copy under `node_modules/draft-js/node_modules`.  Have to remove this nested copy to walk around the current warning hell.

mitermayer:
we will be tracking this issue in here: https://github.com/facebook/draft-js/issues/950

PR #879, opened 2016-12-18
Immutable update

tylercraft:
**Summary**

Updates the version of Immutable draft is using to `3.8.1` in order to remove unnecessary console logs

**Test Plan**

Rebuilt and ran through all examples to verify they still worked.

#607

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "fbjs": "^0.8.3",
-    "immutable": "~3.7.4",
+    "immutable": "~3.8.1",
     "object-assign": "^4.1.0"
   },
   "peerDependencies": {


flarnie:
I'm guessing that there have been minor changes to syntax and strictness between versions, so we should update our type checking.

hellendag:
We need this: https://github.com/facebook/immutable-js/pull/850. (@mitermayer commented on there too.)

tylercraft:
Since the immutable project is blocking on this, I'll close out the PR and we can keep an eye on the issue instead.

Kronuz:
The pull request in facebook/immutable-js#850 was closed three days ago, they say they merged the facebook/immutable-js#1027 which is a superset of it, so it should be in their master now. Can this please be reopened to start trying to upgrade immutable in draft-js again?

Issue #455, opened 2016-06-09
Document how to create a targetRange.


kristoferjoseph:
In the Entities documentation there is a reference to `targetRange` but there is no documentation on what is required to create one.

https://facebook.github.io/draft-js/docs/advanced-topics-entities.html#content

```
const key = Entity.create('LINK', 'MUTABLE', {href: 'http://www.zombo.com'});
const contentStateWithLink = Modifier.applyEntity(
  contentState,
  targetRange,
  key
);
```

This has been brought up previously in #409

maxmatthews:
I'm also struggling with this right now. I've figured out how to create an entity if text is selected with the following code:
```javascript 
var entity = Entity.create('COPYBIT', 'IMMUTABLE', {someData: data});
this.setState({
    editorState: RichUtils.toggleLink(
        this.state.editorState,
        this.state.editorState.getSelection(),
        entity
    )
});
```
but I'm struggling on how to create an entity on a fragment I've created manually.
For example, this code successfully inserts the fragment but I can't understand how to add the entity to it.
```javascript
var entity = Entity.create('COPYBIT', 'IMMUTABLE', {data: someData});

const {editorState} = this.state;
const blockMap = ContentState.createFromText(text).blockMap;
const selection = editorState.getSelection();
let newState = Draft.Modifier.replaceWithFragment(editorState.getCurrentContent(), selection, blockMap);

this.onChange(EditorState.push(editorState, newState, 'insert-fragment'));
```

kristoferjoseph:
@maxmatthews Yup that is where I got stumped as well. Still not sure how create an entity from a programmatically created text selection. Would be great to have how to create a targetRange documented.

Use case: I want to identify user names while someone is typing and programmatically select and add styling to them.

maxmatthews:
Here's how I got it to work. Hopefully it helps @kristoferjoseph out! 
```javascript
var copyBitEntity = Entity.create('COPYBIT', 'IMMUTABLE', {
    copyBitID: newCopyBitID
});

const {
    editorState
} = this.state;
const tempContentState = ContentState.createFromText(text);
const blockMap = tempContentState.blockMap;
const selection = editorState.getSelection();
let newState = Draft.Modifier.replaceWithFragment(editorState.getCurrentContent(), selection, blockMap);

const selectionState = new Draft.SelectionState({
    anchorKey: selection.getAnchorKey(),
    anchorOffset: selection.getStartOffset(),
[...]

PR #880, opened 2016-12-18
Update applyEntity doc

tylercraft:
**Summary**

There has been some confusion over the params passed into `applyEntity`. Since targetRange is an instance of `SelectionState`, lets make the docs more clear. #455 

**Test Plan**

Doc change only.


flarnie:
I think this is helpful. Hope in the future we can add even more detailed examples, and maybe even make the API simpler too.

Issue #579, opened 2016-08-02
Missing .set() method for EditorState in the documentation


klimksh:
In the [documentation example](https://facebook.github.io/draft-js/docs/advanced-topics-decorators.html#setting-new-decorators) set() is used:
`return EditorState.set(editorState, {decorator: onlyHashtags});`

However, in EditorState documentation .set() method is missing.

mjomble:
The `set` method is inherited from `Immutable.Record`.
The documentation does mention this inheritance, although it stresses that the Immutable API should not be used directly, which kind of conflicts with the example.
And I also just noticed that the documentation for `Record` actually doesn't mention the `set` method.
So there's some room for improvement :)

flarnie:
Thanks for bringing this up! On the one hand we don't want to duplicate the Immutable.js documentation, but maybe erring on the side of giving more information rather than less is good here. Hopefully @tylercraft 's PR will clear this up - we will be continuing to improve the docs, so keep the feedback coming. :)

PR #882, opened 2016-12-18
EditorState.set()

tylercraft:
**Summary**

We reference using `Editor.set()` in the docs, as well as in multiple examples. I've updated docs to reference it (#579). However, the reason it wasn't there is because it's a method that is on Immutable's `Record` class.  I know we don't want to recommend these methods, but since we are using it in the examples it may not be a bad idea to at least include a reference to it. That said, feel free to suggest a better way to word this, or to just reject it.

### DIFF
--- a/docs/APIReference-EditorState.md
+++ b/docs/APIReference-EditorState.md
@@ -358,8 +358,8 @@ as the set of inline styles to be applied to the next inserted characters.
 ```
 static set(editorState: EditorState, options: EditorStateRecordType): EditorState
 ```
-Returns a new `EditorState` object with new options passed in. The method
-technically lives on the Immutable `record` API.
+Returns a new `EditorState` object with new options passed in. 'The method is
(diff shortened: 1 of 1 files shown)


docs/APIReference-EditorState.md
```diff
@@ -348,6 +353,14 @@ static setInlineStyleOverride(editorState: EditorState, inlineStyleOverride: Dra
 Returns a new `EditorState` object with the specified `DraftInlineStyle` applied
 as the set of inline styles to be applied to the next inserted characters.
 
+### set
+
+```
+static set(editorState: EditorState, options: EditorStateRecordType): EditorState
+```
+Returns a new `EditorState` object with new options passed in. The method
+technically lives on the Immutable `record` API.
```
flarnie: I would say 'The method is inherited from the Immutable `record` API.'

Issue #866, opened 2016-12-12
Add notice to `README` about ongoing API changes


flarnie:
Context:
We are changing the API of Entity storage in Draft. Right now the state of things on the master branch supports both the old and new API, and this will soon be released as v0.10.0. Then in v0.11.0 we will remove the old API and provide documentation on how to upgrade. See https://github.com/facebook/draft-js/issues/839

Your Mission:
We want to let people know in a really obvious way about this, so that it's not confusing. A notice about it in the `README` could help. Possibly also on the documentation site. We'd love some help for how to make this both concise and helpful.

dlutwuwei:
Great, Looking forward to it, we plan to import the editor to our PGC website.

flarnie:
Thanks @tylercraft !!!

PR #883, opened 2016-12-19
API Notice doc update

tylercraft:
**Summary**

Adds a notice to the readme and Overview Docs page to let users know the upcoming changes to the API.

#866


flarnie:
This is the best! Thanks!

Issue #412, opened 2016-05-24
After removing an inline style at the end of a string and then hitting return, the inline style remains


huangbong:
Using underline as an example:

![](http://g.recordit.co/OejE1iOcsm.gif)

When you hit return the Underline becomes highlighted again and the corresponding typed text on the next line is underlined.

hellendag:
Thanks for reporting!

I think the appropriate fix here would be to reapply any inline style override within https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/commands/keyCommandInsertNewline.js#L18.

PR #892, opened 2016-12-21
Prevent discarding style override on split-block

existentialism:
Another attempt at fixing #412 and #672 and added a test for not discarding override styles when adjusting depth.

mm


hellendag:
Sure, I think this seems reasonable. Thanks for the tests too!

PR #894, opened 2016-12-21
Fix Flow nits in DraftEditorProps

flarnie:
In one case it seems the return type was just incorrect.

In another case we were allowing 'string' instead of the more specific
'DraftEditorCommand'. To support the typing of 'getDefaultKeyBinding' we
should match the return type there.

### DIFF
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -48,12 +48,12 @@ export type DraftEditorProps = {
   blockRendererFn?: (block: ContentBlock) => ?Object,
 
   // Function that returns a cx map corresponding to block-level styles.
-  blockStyleFn?: (type: number) => string,
+  blockStyleFn?: (block: ContentBlock) => string,
 
   // A function that accepts a synthetic key event and returns
   // the matching DraftEditorCommand constant, or null if no command should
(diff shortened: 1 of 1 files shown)


src/component/base/DraftEditorProps.js
```diff
@@ -48,12 +48,12 @@ export type DraftEditorProps = {
...
 
   // Function that returns a cx map corresponding to block-level styles.
-  blockStyleFn?: (type: number) => string,
+  blockStyleFn?: (block: ContentBlock) => string,
 
   // A function that accepts a synthetic key event and returns
   // the matching DraftEditorCommand constant, or null if no command should
   // be invoked.
-  keyBindingFn: (e: SyntheticKeyboardEvent) => ?string,
+  keyBindingFn: (e: SyntheticKeyboardEvent) => ?DraftEditorCommand,
```
hellendag: I'm not sure this is entirely correct. `getDefaultKeyBinding` will return a `DraftEditorCommand`, but one might define a custom key binding function that returns a different arbitrary command string.

flarnie: Thanks - I will update the comment there and loosen types in other places to allow for that use case.

PR #897, opened 2016-12-22
Expose ContentBlock-> raw

AnilRedshift:
Due to perf reasons I need to expose the inner function of converting a content block to raw format.

This allows for caching outside of draftJS


AnilRedshift:
Wrong repo sorry!

PR #918, opened 2017-01-06
Fix Flow nits in DraftEditorProps

flarnie:
In one case it seems the return type was just incorrect. We also
clarified the comments for 'keyBindingFn' in order to avoid confusion -
previously due to some Flow issues we had almost changed this flow type
to be incorrect. See https://github.com/facebook/draft-js/pull/894

Also fixes a hanging comma lint.


flarnie:
Going to merge this in order to unblock releasing 0.10.0 soon.

PR #920, opened 2017-01-06
Make 'decorator' optional and allow null consistenly in EditorState

flarnie:
This fixes a minor flow error coming up in a FB use case, and makes the
overall typing more consistent in this file.


flarnie:
Merging this to keep moving and unblock releasing 0.10.0.

Issue #849, opened 2016-12-07
Deleting highlighted text deletes old selection


drewjenkins:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Highlighting text then deleting it deletes text at the previous selection state instead of the current selection.

![defect](https://cloud.githubusercontent.com/assets/5521743/20978037/a1a7a0bc-bc64-11e6-9da1-c9c0296c0827.gif)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: 
1.  Type out a sentence
  1a. Take note of the current cursor position
2. Highlight some text in the middle of the sentence
3. Press delete/backspace. 
Expected Results:
Text under the highlighting is deleted
Actual Results:
Text at the previous selection is deleted

This issue is reproducible on facebook.com's "What's on your mind?" editor.

**What is the expected behavior?**
Highlighted text should be deleted

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
Experiencing this on Mac OSX 10.11.6 Chrome Version 55.0.2883.75 beta (64-bit).  Have customers who have reported seeing it on Windows 7 Browser Control (pretty much a hacky version of IE) as well.

Unsure if this worked in previous versions of draft. We are using draft-js 0.9.1.

davidchang:
@drewjenkins - thanks for opening this. this is crazy, definitely a bug.

are you able to see it on the editor at this page? https://facebook.github.io/draft-js/ i tried there as well as on the "What's on your mind?" editor on Facebook and wasn't able to repro the behavior (though my Mac OSX was 10.12 and Chrome version 54). do you have any browser extensions?

drewjenkins:
Hi @davidchang 

I'm able to reproduce on that page as well.  I can also try on some other browsers as well. I'll let you know what I see.

![repro](https://cloud.githubusercontent.com/assets/5521743/21245074/ceb960fc-c2dd-11e6-80c5-eb9eb4d32658.gif)

drewjenkins:
Here are some preliminary testing results

Reproducible
Windows 10 - IE11 (Even Crazier issue, deletes all but the very first letter of whatever you typed, regardless of amount of text highlighted)
Windows 10 - WPF Web Browser Control (Same issue as IE11)
Mac OSX 10.11.6 - Firefox 49
[...]

PR #921, opened 2017-01-06
fix #917, #849 - IE input event bug missing editOnInput

### LINKED ISSUE #917: 0.10 does not accept more than one character of input in IE 11
Bug

Using the current master typing in IE appears to accept the characters as normal but only the first character is actually added to the editor state. As soon as the editor attempts to rerender itself all but the first character disappears.

* Visit https://facebook.github.io/draft-js using IE 11.
* Type 'Hello' into the sample editor
* Click away to move focus away from the editor
* Click back into the editor to return focus
* The text immediately changes to 'H'

All characters that are visible should be part of the editor's internal state and so should remain when rerendering the editor.

This behavior regressed as a result of https://github.com/facebook/draft-js/pull/667. If Draft decides that the browser should handle the insertion natively this change defers most of the work to update the editor state to when the next `input` event arrives. However, IE doesn't support this event in contenteditable elements and so `editOnInput` is never called and the editor state is never updated. The first character is successfully saved because the selection is at a leaf start and so Draft explicitly handles the input.

davidchang:
**Summary**

Fix #917 ; syncing from an internal fix to ensure Draft works properly on IE

**Test Plan**

Particularly on IE11, go to facebook.github.io/draft-js . Type in the editor, blur focus and then return; note that the text remains the same.

Select that text and hit backspace, note expected behavior (selected text is deleted, cursor position is where the deleted text had been).

### DIFF
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -75,6 +75,10 @@ function replaceText(
  * occurs on the relevant text nodes.
  */
 function editOnBeforeInput(editor: DraftEditor, e: SyntheticInputEvent): void {
+  if (editor._pendingStateFromBeforeInput !== undefined) {
+    editor.update(editor._pendingStateFromBeforeInput);
+    editor._pendingStateFromBeforeInput = undefined;
+  }
   var chars = e.data;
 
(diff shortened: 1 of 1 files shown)


davidchang:
actually, this is basically the same as #871

davidchang:
talked to flarnie, closing this in favor of #871

Issue #296, opened 2016-04-12
Refs must have owner


masylum:
Hi,

I just wanted to try Draft.js and tried to drop the example in my application and I've got the https://fb.me/react-refs-must-have-owner issue.

I simplified a little bit my code, but looks something like this:

``` js
// ModalComposer.jsx
import React from 'react'
import { Editor, EditorState } from 'draft-js'

const ModalComposer = React.createClass({
  getInitialState () {
    return {
      editorState: EditorState.createEmpty()
    }
  },

  onChange (editorState) {
    this.setState({editorState})
  },

  render () {
    const { editorState } = this.state

    return (
      <div>
        <Editor
          editorState={editorState}
          onChange={this.onChange}
        />
      </div>
    )
  }
})

export default ModalComposer
```

``` js
// Modal.jsx
import { connect } from 'react-redux'
import ModalComposer from './ModalComposer.jsx'
import * as modalActions from '../../ducks/modals'
import React from 'react'

const Modal = React.createClass({
  propTypes: {
    dispatch: React.PropTypes.func.isRequired
  },

  onClickClose () {
    this.props.dispatch(modalActions.pop())
  },

  render () {
    return (
      <div>
        <button onClick={this.onClickClose}></button>
        <ModalComposer />
      </div>
    )
  }
})

export default connect()(Modal)
```

I get the following:

> invariant.js:39 Uncaught Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner

When I comment out the component `ModalComposer` (which is basically the draft-js one) I don't get this warning, so its not related to npm and dependencies.

Thanks in advance!

hellendag:
I don't see any obvious problems with the code here.

At what point are you seeing this invariant? Immediately when the component renders?

masylum:
I will try to debug it and isolate better the issue to give you better information

apuntovanini:
+ 1 @masylum, same issue here, with super simple example

apuntovanini:
@hellendag I see it when the component renders, no matter whether it is included in a more complex component or rendered simply

hellendag:
@apuntovanini Can you provide a gist of your code? Are you seeing any issues with refs in the examples provided with the draft-js repo?

mattmarcello:
[...]

PR #925, opened 2017-01-09
using newer 'ref' syntax

mks11:

4. Ensure that:

-

**Summary**

[...]

**Test Plan**

[...]


examples/draft-0-10-0/plaintext/plaintext.html
```diff
@@ -49,7 +51,7 @@
                   editorState={this.state.editorState}
                   onChange={this.onChange}
                   placeholder="Enter some text..."
-                  ref="editor"
+                  ref={(d)=>{this.domEditor = d}}
```
Daniel15: This will create a new arrow function on every render, which is unnecessary. Instead, do:
```
ref={this.setDomEditorRef}
```

Then add a `setDomEditorRef` method. Looks like it does this in the constructor so it's bound correctly:
```
this.setDomEditorRef = ref => this.domEditor = ref;
```

flarnie:
This is close to ready - it would be great to see it updated based on @Daniel15's suggestion.

If we update the other 'ref' uses this would also fix an error that some users encounter when loading 2 versions of React; https://github.com/facebook/draft-js/issues/296

I can try and update this if/when there is time as well.

mks11:
I commited as per the feedback. I will also try to update the other old style refs.

flarnie:
This looks good to go - going to add it to my list of PRs to import and land.

> I commited as per the feedback. I will also try to update the other old style refs.

Feel free to open another similar PR updating the other refs. I think there are only two.

PR #927, opened 2017-01-09
Don't merge sequential unstyled tags

### LINKED ISSUE #738: `convertFromHTML` merged multiple `unstyled` tags into one

Report a bug

Based on `jsdom`, it is easy to reproduce this bug:
```js
const jsdom = require('jsdom')
jsdom.env('', function (err, window) {
  global.window = window
  global.document = window.document
  global.navigator = window.navigator
  global.HTMLElement = window.HTMLElement
  global.HTMLAnchorElement = window.HTMLAnchorElement
  const djs = require('draft-js')
  console.log(djs.convertToRaw(djs.ContentState.createFromBlockArray(djs.convertFromHTML('<h1>Title</h1><div>Hello</div><div>World</div>'))))
  process.exit(0)
})
```

Providing `<h1>Title</h1><div>Hello</div><div>World</div>`, `convertFromHTML` outputs two blocks:

```
{ entityMap: {},
  blocks:
   [ { key: 'bv0iv',
       text: 'Title',
       type: 'header-one',
       depth: 0,
       inlineStyleRanges: [],
       entityRanges: [],
       data: {} },
     { key: '1j1ko',
       text: 'HelloWorld',
       type: 'unstyled',
       depth: 0,
       inlineStyleRanges: [],
       entityRanges: [],
       data: {} } ] }
```

instead of three:

```
{ entityMap: {},
  blocks:
   [ { key: '8ahq8',
       text: 'Title',
       type: 'header-one',
       depth: 0,
       inlineStyleRanges: [],
       entityRanges: [],
       data: {} },
     { key: '18k8c',
       text: 'Hello',
       type: 'unstyled',
       depth: 0,
       inlineStyleRanges: [],
       entityRanges: [],
       data: {} },
     { key: '9p850',
       text: 'World',
       type: 'unstyled',
       depth: 0,
       inlineStyleRanges: [],
       entityRanges: [],
       data: {} } ] }
```

https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L504
might be the cause of this. Should we add 'div' to `workingBlocks` to make it be [ 'blockquote', 'figure', 'h1', 'h2', 'h3','h4','h5','h6','li','pre','div'] ? 

Draft-js 0.9.1

colinjeanne:
**Summary**

This change fixes #738 by ensuring that `convertFromHTMLToContentBlocks` counts unstyled blocks as blocks and by moving the logic to suppress a block's type into that function.

Previously the existence of any tag other than the unstyled tag would cause unstyled blocks not to be parsed as blocks (but, curiously, the aliases of the unstyled tag would be parsed as blocks). This is mean to handle a case like

```
<div><h2>Hello</h2></div>
```

ensuring that the final block is a `header-two` but breaks with examples like

```
<h2>Hello</h2>
<div>world!</div>
<div>Goodbye!</div>
```

which parses as two blocks with the two divs being combined into a single unstyled block.

The core issue is that not counting the unstyled tag when an interesting tag appears anywhere in the document is a global problem to a local issue: the fix is to instead consider each tree under a block individually with unstyled blocks deferring their final type to whatever the next lowest block type is.

**Test Plan**

Several tests were added including uncommenting a test that had existed before.

The ordering of tags in the uncommented test had to be revised because the HTML was invalid and so being parsed with three children under `body`. The corrected ordering keeps the goal of the test.

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -156,7 +156,6 @@ function getListBlockType(
 function getBlockMapSupportedTags(
   blockRenderMap: DraftBlockRenderMap
 ): Array<string> {
-  const unstyledElement = blockRenderMap.get('unstyled').element;
   let tags = new Set([]);
 
   blockRenderMap.forEach((draftBlock: DraftBlockRenderConfig) => {
@@ -170,7 +169,6 @@ function getBlockMapSupportedTags(
   });
 
   return tags
-    .filter((tag) => tag && tag !== unstyledElement)
     .toArray()
     .sort();
 }
@@ -430,7 +428,13 @@ function genFragment(
   }
 
   // Block Tags
-  if (!inBlock && blockTags.indexOf(nodeName) !== -1) {
+  if (
+      (
+        !inBlock ||
+        getBlockTypeForTag(inBlock, lastList, blockRenderMap) === 'unstyled'
+      ) &&
+      blockTags.indexOf(nodeName) !== -1
+    ) {
     chunk = getBlockDividerChunk(
       getBlockTypeForTag(nodeName, lastList, blockRenderMap),
       depth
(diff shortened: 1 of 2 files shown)


DawnWright:
I logged the same bug here: https://github.com/facebook/draft-js/issues/1153
Can we get this merged into master?

Issue #958, opened 2017-01-19
draft-js fails flow 0.38.0


villesau:
new flow gives following errors:

node_modules/draft-js/lib/expandRangeToStartOfLine.js.flow:34
 34:   document.body.appendChild(div);
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `appendChild`. Method cannot be called on possibly null value
 34:   document.body.appendChild(div);
       ^^^^^^^^^^^^^ null

node_modules/draft-js/lib/expandRangeToStartOfLine.js.flow:36
 36:   document.body.removeChild(div);
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `removeChild`. Method cannot be called on possibly null value
 36:   document.body.removeChild(div);
       ^^^^^^^^^^^^^ null

node_modules/draft-js/lib/getSafeBodyFromHTML.js.flow:29
 29:     doc.documentElement.innerHTML = html;
                             ^^^^^^^^^ property `innerHTML`. Property cannot be assigned on possibly null value
 29:     doc.documentElement.innerHTML = html;
         ^^^^^^^^^^^^^^^^^^^ null

flarnie:
Thanks for reporting this - I'm having trouble reproducing it locally. I've created a PR to update the flow-bin version to 0.38.0 to verify that things work on master.

flarnie:
Glad that this reproduces in CI - not sure why I am not seeing it locally. Will look into it more when I get time, but if someone else wants to tackle it go ahead.

existentialism:
`document.body` and `document. documentElement` are nullable after https://github.com/facebook/flow/pull/3127

thetalecrafter:
It would be great to have a patch release with this fix. It's blocking upgrading flow.

riseremi:
I agree with you. These are annoying warnings.

PR #963, opened 2017-01-27
Update flow-bin version to 0.38.0

flarnie:
To further investigate issue #958 and to keep our dependencies fresh, we are updating the flow package version.


flarnie:
outdated since we landed https://github.com/facebook/draft-js/pull/996

PR #966, opened 2017-01-27
add the code to show how obtain myContentBlock

Kiwka:
add code that make it more easier to understand how the myContentBlock can be obtained for the SelectionState

### DIFF
--- a/docs/APIReference-SelectionState.md
+++ b/docs/APIReference-SelectionState.md
@@ -47,13 +47,13 @@ For instance, when extracting a slice of text from a block based on a
 `SelectionState`, it is irrelevant whether the selection is backward:
 
 ```
-var selection = editorState.getSelection();
-var anchorKey = selection.getAnchorKey();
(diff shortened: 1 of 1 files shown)


docs/APIReference-SelectionState.md
```diff
@@ -47,6 +47,10 @@ For instance, when extracting a slice of text from a block based on a
 `SelectionState`, it is irrelevant whether the selection is backward:
 
 ```
+var selection = editorState.getSelection();
```
flarnie: I would call this 'selectionState', and then we can assume that it's the same variable being used below.

I'm also wondering if it would be more clear to rename 'myContentBlock' to 'currentContentBlock'?

Kiwka: @flarnie I have made appropriative changes: change the name for the selection variable, and use currentContentBlock in spite of the myContentBlock, as for me also it makes more sense

Issue #839, opened 2016-12-04
Draft.js API improvements in 0.10.0 and 0.11.0


flarnie:
**What:** The API for managing entities will be moved from the global `Entity` module into `contentState`, with a deprecation warning in v0.10.0 and a full breaking change in v0.11.0.

**Why:** 
 - currently changing or adding entities won't trigger a re-render. This is a big pain!
 - the use of a non-Immutable storage for entities doesn't fit well with the rest of the Draft architecture
 - moves Entity storage out of global module; this just makes more sense as an API, rather than having it globally available.
Other maintainers - feel free to add detail and say this more eloquently.

**Will 0.10.0 include breaking changes?**
Nope! When you update to v0.10.0, you should start seeing warnings in the console when the old `Entity` module API is used. The new API, using `contentState` will be supported, and you can migrate to the new API incrementally.

We will put out documentation about how to update all use cases when releasing 0.10.0.

Once you are fully migrated to the new API, you can update to 0.11.0, which will include the fully breaking change and will also include the benefits of the new API.

**What's taking so long???**
 - Since we are improving the API in a breaking way, we are doing extra testing and preparing documentation to make the 0.10.0 and 0.11.0 releases run smoothly.
 - We are a tiny team with other projects, and so thanks for everyone's patience! We are actively working on catching up, and there are discussions ongoing for how to improve the state of this project. See #835.

flarnie:
Just an update - we released 0.10.0, including the deprecation of `DraftEntity` and the new API for managing `DraftEntity` records.

Next steps:
 - Publish v0.11.0@next for beta testing
 - Publish any intermediate bug fixes at v0.10.1
 - Internal testing of v0.11.0 on Facebook's wide range of use cases
 - Release v0.11.0 and continue doing regular releases in the future

tobiasandersen:
Awesome! Great job @flarnie :1st_place_medal:

sdtsui:
Nice!  Looking forward to seeing the changes. 🎉

asprouse:
@flarnie Any update on publishing 0.11@next? 0.10's pseudo-immutable behavior is causing bugs/confusion (#1047)

SmilinBrian:
It is unclear to me how import from any source other than "raw" `ContentState` or plain-vanilla HTML is best accomplished once Entities cannot be created without a `ContentState` instance.

[...]

Issue #203, opened 2016-03-14
Selection and Keyboard Navigation of Media Blocks


thesunny:
This is a problem I've been thinking about for a week and I thought I'd start discussion on it.

The problem is how selection and keyboard navigation works with Media blocks.

I'm also focusing on the keyboard not the mouse.

There are two ways media types like images work in  other editors: (1) as inline elements (e.g. in Google Drive) and (2) as blocks (e.g. in Medium). I'm specifically targeting their use as blocks.

## Inline Media Behavior

For completeness of understanding though I'll go over Inline media first. The typical way to select inline media elements is that the caret moves around them. If there are two images, a caret moving from left to right would look like the following where X is the image:

|XX
X|X
XX|

In order to select an image, one could hold down the shift key. For example, hold shift then key right one to select the first image:

shift + right arrow results in

|**X**|X

## Current Media Block Navigation Behavior

When the media types are implemented as blocks in DraftJS, we typically make the blocks contentEditable={false}. When we mouse down from above 2 media blocks, the selection skips the media blocks.

```
|
X
X
```

Key down arrow results in:

```
X
X
|
```

## Media Block Navigation Behavior I'd Like to Support

What I'd like to have the DraftJS editor be able to support, is to have the key down arrow to instead select the first media block. This behavior is that adopted by the Medium editor.

**|X|**
X

Another key down arrow results in

X
**|X|**

Something like this is important. Consider 10 media blocks stacked on top of each other. A key down arrow would skip more than an entire screenful. This is jarring.

## Alternate Block Navigation Behavior (Not Recommended Personally)

Some other options would allow the cursor to be in position around the media block but it feels a little less straightforward and I don't see any utility above the implementation above. For completeness though, a user could key right to get all these positions:

|
X
X

key right

|X
X

key right

X|
X

key right

X
|X

And so forth.

## Discussion of Solution

Note: I'm not advocating for selection of media blocks to necessarily be baked into DraftJS although this may not be a bad idea. I'm hoping for a solution that could work for this common use case.

The biggest issue is the skipping of all contentEditable={false} blocks.

One thing I've been thinking about is to leave the root media block as contentEditable={true} and only have a contentEditable={false} `div` inside. When the cursor moves into that contentEditable position, we recognize the caret is within the media block, and we can restyle the block with an outline or with handles or whatever we choose. In this scenario, we would also have to figure out a way to make the blinking caret be invisible, perhaps with some CSS. Not sure if this is even possible though.

[...]

PR #971, opened 2017-01-28
Add onLeftArrow and onRightArrow props

kedromelon:
**Summary**

While implementing some custom selection logic, I found it somewhat unintuitive that `onUpArrow` and `onDownArrow` are exposed as props, but `onLeftArrow` and `onRightArrow` are not.

Given the added complexity due to [the workaround needed by firefox for handling CMD+LEFT/RIGHT](https://github.com/facebook/draft-js/blob/master/src/component/utils/getDefaultKeyBinding.js#L113-L124), it seems to me that it may make sense to have these handlers return `DraftHandleValue` (like `handleReturn`, etc) so that they can still fall through to `getDefaultKeyBinding`. 

Is this appropriate? If so, would these props be better named `handleLeftArrow` and `handleRightArrow`? Should any other `on*` props be refactored to use this logic?

**Test Plan**

As far as I can tell there are no existing tests for the modified files. Did I miss any? Should any be added?

⚠️  This is my first OSS PR, so I apologize in advance if I've missed any PR etiquette and such. Love draft-js and hope I'm able to help contribute to it 😄

Thanks for taking the time to check out this PR!

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -226,7 +226,6 @@ onUpArrow?: (e: SyntheticKeyboardEvent) => void
 onRightArrow?: (e: SyntheticKeyboardEvent) => DraftHandleValue
 ```
 
-
 #### onDownArrow
 ```
 onDownArrow?: (e: SyntheticKeyboardEvent) => void
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -211,10 +211,21 @@ onEscape?: (e: SyntheticKeyboardEvent) => void
 onTab?: (e: SyntheticKeyboardEvent) => void
 ```
 
+#### onLeftArrow
+```
+onLeftArrow?: (e: SyntheticKeyboardEvent) => DraftHandleValue
+```
+
 #### onUpArrow
 ```
 onUpArrow?: (e: SyntheticKeyboardEvent) => void
 ```
+
+#### onRightArrow
+```
+onRightArrow?: (e: SyntheticKeyboardEvent) => DraftHandleValue
+```
+
 
 #### onDownArrow
 ```
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -123,7 +123,9 @@ export type DraftEditorProps = {
    */
   onEscape?: (e: SyntheticKeyboardEvent) => void,
   onTab?: (e: SyntheticKeyboardEvent) => void,
+  onLeftArrow?: (e: SyntheticKeyboardEvent) => DraftHandleValue,
   onUpArrow?: (e: SyntheticKeyboardEvent) => void,
+  onRightArrow?: (e: SyntheticKeyboardEvent) => DraftHandleValue,
   onDownArrow?: (e: SyntheticKeyboardEvent) => void,
 
   onBlur?: (e: SyntheticEvent) => void,
--- a/src/component/handlers/edit/editOnKeyDown.js
+++ b/src/component/handlers/edit/editOnKeyDown.js
@@ -106,9 +106,25 @@ function editOnKeyDown(editor: DraftEditor, e: SyntheticKeyboardEvent): void {
     case Keys.TAB:
       editor.props.onTab && editor.props.onTab(e);
       return;
+    case Keys.LEFT:
+      if (
+        editor.props.onLeftArrow && 
+        isEventHandled(editor.props.onLeftArrow(e))
+      ) {
+        return;
+      }
+      break;
     case Keys.UP:
       editor.props.onUpArrow && editor.props.onUpArrow(e);
       return;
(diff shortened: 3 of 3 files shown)


flarnie:
:) Thanks for the great commit message and for contributing @kedromelon! I'm curious what your Draft use case is, and how you would use `onLeftArrow` and `onRightArrow`.

You are right that it seems silly to expose `onUpArrow` and `onDownArrow` but not `onLeftArrow` and `onRightArrow`. You're also right that, following the pattern of `handleReturn` we would want to return a `DraftHandledValue` from the callback in order to allow falling back to the default behavior in `getDefaultKeyBinding` that fixes that FF bug.

I am still thinking about the question of how to be consistent once we have made that change - it looks like `handleReturn` has been treated as a special case and the other key handlers are not treated as cancellable.

Some options we may consider;
 -  Update all `on<Key>` props to be cancellable, meaning they must return a 'DraftHandledValue' or true in order to actually override the default behavior of the editor. But for now we don't have any special behavior for any of those.
    - +++ This leaves room for adding more custom behaviors within `getDefaultKeyBinding.js` in the future, if more browser specific bugs were to arise, like in the case of left and right in older FF versions.
    - --- We would need to consider how to do this in a way that doesn't break behavior for folks using `on<Key>` props currently.
 - Deprecate the `on<Key>` props and replace with `handle<Key>` props that return a 'DraftHandledValue'.
    - +++ The naming of props would be consistent between `handleReturn` and the other key handler props.
    - +++ This _also_ leaves room for adding more custom behaviors within `getDefaultKeyBinding.js` in the future, if more browser specific bugs were to arise, like in the case of left and right in older FF versions.
    - --- This does break the API and would require folks to migrate their changes. We could probably publish a codemod to make it easier though, since it's just a rename and possibly wrapping the callback in something that returns 'handled'. I would delay releasing that type of change until after v0.11.0 is released though, due to [other API changes happening](https://github.com/facebook/draft-js/issues/839).

Either way it would be nice to add an example, or at least know some possible use cases for these props.

_Sidenote:_
I found some notes about the FF bug that was being fixed by adding the left and right key handlers for Firefox, and will probably add a bit more detail to the inline comments about that, for posterity. Mainly [a link to the bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=289384), and more detail in `getDefaultKeyBinding.js`.

tobiasandersen:
Just thought I'd chime in hear. We've created our own version of atomic blocks, that are selectable. In order for them to be that, and still play nicely with the "normal selection", we need to tell Draft what to do with the different arrow key events. We also have to see if `shift` is held down, and in that case expand the selection rather than moving it. We're actually intercepting a lot of different key events to make it work well, and it's a little bit annoying that they're handled using different naming conventions and with different return types.

Some more thoughts about the arrow events: the recommendation seems to be to use the `keyBindingFn` to create commands for `handleKeyCommand` to take care of. Using the current API, this is not possible to do with the up and down arrow keys, but it is for left and right. While this isn't preventing us from creating what we want, the solutions for seemingly similar operations look quite different.

Having said that, I think @flarnie's second option sounds best, since all key events would be handled, and named, in a similar way. For someone unfamiliar with the code base, I don't think it's clear how and why `handleReturn` is different from `onArrowUp` (except for the actual keys).

kedromelon:
@tobiasandersen Yep! I have a similar use case. Essentially trying to create Medium-like selection of Atomic Blocks, not unlike what was discussed in https://github.com/facebook/draft-js/issues/203.

I would agree that @flarnie's second option sounds better. As mentioned above, having all of these "handlers" (for lack of a better term) named `handle<Key/Action>` and returning `DraftHandleValue` would be much clearer. It certainly took me a bit of research to understand the difference between `on<Key>` and `handle<Key>`.

I'll try to see if I can get an isolated example put together soon.

Thanks for the input, @flarnie and @tobiasandersen!

kedromelon:
Okay, got a chance to put together an [example](https://jsfiddle.net/ue7krjyu/1/).

Essentially, what I'm trying to illustrate is the inconsistent API between up/down and left/right arrows. In the example, I want to handle key commands myself when the cursor is adjacent to an atomic block.

flarnie:
Thanks for your patience on this PR - I have been following the conversation, and need to fix some other issues with Draft before giving this my full attention. It is still on my radar.

kedromelon:
👌 no rush! was able to work around it just fine for my use case, but i do think this API is worth improving.

octatone:
> Deprecate the on<Key> props and replace with handle<Key> props that return a 'DraftHandledValue'.

Just wanted to put a vote in favor of this solution.  I have been using draft-js for the past months for a project and have found the API for custom key handling inconsistent.  I would prefer a one-way-to-rule-them all API.

kedromelon:
🎉 saw this functionality was merged in https://github.com/facebook/draft-js/pull/1384! gonna close this issue.

Issue #965, opened 2017-01-27
Documentation page titles don't help content navigation


everyonesdesign:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
All the titles of documentation pages say "Draft.js | Rich Text Editor Framework for React", which is really annoying when you open multiple documentation tabs.
![image](https://cloud.githubusercontent.com/assets/3520704/22364252/2c3718ae-e481-11e6-8092-b46d92fc7883.png)
It seems that there's also no any favicon for the docs.

**What is the expected behavior?**
Informative beginning of the titles, e.g. "Custom Block Components | Draft.js | Rich Text Editor Framework for React" would be great in my opinion.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
N/A

Thank you.

flarnie:
Great idea! Hope someone picks it up - I'm pretty focused on releasing 0.10.0 and 0.11.0 but after that is done I will pick this up if nobody has done it.

Thanks @everyonesdesign

kedromelon:
Opened https://github.com/facebook/draft-js/pull/972 for this! 

Let me know if you've got any suggestions! 😄

kedromelon:
Just realized that the [og:title](https://github.com/facebook/draft-js/blob/master/website/core/Site.js#L26) meta should probably be updated as well. I'll try to get around to making a PR for this later!

kedromelon:
Is there any appropriate favicon that should be used? That's the only aspect of this issue that's still unresolved.

flarnie:
Good question @kedromelon - ideally we will get a logo soon but we don't have one yet. Until then, my first guess would be to use a 'D' with the same colors as the documentation, like the screenshot below. 
![screen shot 2017-02-10 at 7 39 26 am](https://cloud.githubusercontent.com/assets/1114467/22832787/41b2d306-ef64-11e6-8cce-22aa46208743.png)
We could also close this issue and open a separate one for a logo to use in the favicon, docs, twitter, etc.

kedromelon:
eek tough to get this to look right given the visual weight of the letter -- but something like this?

![draftjs-draft](https://cloud.githubusercontent.com/assets/5393571/22868055/39807024-f15f-11e6-93f5-634291866daa.png)

baldwmic:
@flarnie @kedromelon I added the above favicon to the website in #1526. Let me know what you think!

[...]

PR #972, opened 2017-01-28
Add documentation page title to header data

kedromelon:
Adds documentation page title as a prefix to `<title>` tag in site header as proposed in https://github.com/facebook/draft-js/issues/965. For example, "Custom Block Components | Draft.js | Rich Text Editor Framework for React" would be the title for the "Custom Block Components" page.

Implements `pageTitle` prop on the `Site` component to accomplish this. If `titlePrefix` is not available, it will fall back to "Draft.js | Rich Text Editor Framework for React".

### DIFF
--- a/website/core/Site.js
+++ b/website/core/Site.js
@@ -15,12 +15,13 @@ var HeaderLinks = require('HeaderLinks');
 
 var Site = React.createClass({
   render: function() {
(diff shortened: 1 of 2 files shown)


flarnie:
So awesome! ^_^ Thanks for doing this.

PR #979, opened 2017-01-30
documentation: fix the arguments in the decorators

Kiwka:
Fix the arguments order and quantity in the decorators strategy accordingly to the code expectations https://github.com/facebook/draft-js/blob/7390e2972145494146202c3d3480eb9509dc90d4/src/model/decorators/CompositeDraftDecorator.js#L71


flarnie:
Nice catch - we had some back and forth about the ordering there when changing the API, I was probably looking at an older version of the code when updating that.

PR #984, opened 2017-01-31
documentation: share prefixed Title string with og:title metatag

kedromelon:
Following https://github.com/facebook/draft-js/pull/972, reuse prefixed page title strings for `og:title` metatag on the documentation pages.


flarnie:
Double-thanks! :)

Issue #873, opened 2016-12-15
Tests for `getDecorations`


flarnie:
The API of `getDecorations` got broken, and it wasn't clear what the desired behavior there was. Thanks to @colinjeanne for fixing it in https://github.com/facebook/draft-js/pull/860.

We should protect that logic with tests, so that it can't silently break in the future.

PR #995, opened 2017-02-04
Test that `getDecorations` is called with correct arguments

kedromelon:
For issue https://github.com/facebook/draft-js/issues/873, adds test to `EditorState` to check that `getDecorations` is called with desired argument types and in the proper order.

### DIFF
--- a/src/model/immutable/__tests__/EditorState-test.js
+++ b/src/model/immutable/__tests__/EditorState-test.js
@@ -299,6 +299,15 @@ describe('EditorState', () => {
       );
     });
 
+    it('must call decorator with correct argument types and order', () => {
+      var decorator = new Decorator();
+      var editorState = getDecoratedEditorState(decorator);
+      decorator.getDecorations.mock.calls.forEach((call) => {
+        expect(call[0] instanceof ContentBlock).toBe(true);
+        expect(call[1] instanceof ContentState).toBe(true);
+      });
+    });
+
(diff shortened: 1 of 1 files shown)


src/model/immutable/__tests__/EditorState-test.js
```diff
@@ -299,6 +299,15 @@ describe('EditorState', () => {
       );
     });
 
+    it('must call decorator with correct argument types and order', () => {
+      var decorator = new Decorator();
+      getDecoratedEditorState(decorator);
+      decorator.getDecorations.mock.calls.forEach((call) => {
+        expect(call[0] instanceof ContentBlock).toBe(true);
```
kedromelon: Is this sufficient? Was the best way I could think to test the API, but it feels a bit scant.

flarnie:
This looks good! It will verify that the call to `getDecorations` in  `src/model/immutable/BlockTree.js` is correct.

Looking at this now, it seems like `CompositeDraftDecorator-test.js` would normally also catch a problem with the API of 'getDecorations' breaking, but the test was also using the wrong API because of the weird state that things got into.

So I think this is good - some would say it's testing too much of the internal implementation, but for now I'd rather safeguard the API and change the test if we need to later.

flarnie:
Also  not sure why the website build failed - could be a fluke so I restarted the build.

PR #996, opened 2017-02-05
Update flow to 0.38

nkt:
@flarnie review, please


flarnie:
Thanks for working on this - lgtm. I restarted the build to see if that failure was a fluke. It seems like the website should build fine since you updated both `package.json` and `src/.flowconfig`.

PR #1004, opened 2017-02-10
Add default value for type in convertFromRawToDraftState

GordyD:
**Summary**

In instances where type is undefined or null, this would be used as type due to the way type is destructured from each `block` in `contentBlocks`. This commit adds to a default value which ensure any falsey is changed to `unstyled` ensure the returned ContentBlock always has a valid DraftBlockType as its type. A similar approach in this file exists for all other values that are extracted from each block.

**Test Plan**

- As this change is within `model` the test plan is to use a unit test to validate the behavior that is expected after the change.
- Running the test without the change shows that this behavior was not present prior.


flarnie:
Super awesome of you to add the test! :)

PR #907, opened 2017-01-01
Fix editor losing focus after first word on Android.

tonygentilcore:
This patch narrows the scope of `restoreEditorDOM` to only blow away the
`DraftEditorContents` instead of the entire editor. The key difference is
that the `contentEditable` DOM element is not replaced, which means that
the soft keyboard isn't dismissed on Android.

Manually tested against "Cut" and "IME" tests:
https://github.com/facebook/draft-js/wiki/Manual-Testing#cutcopypaste
https://github.com/facebook/draft-js/wiki/Manual-Testing#ime

Fixes #888, Ref #626, Ref #102

PR #1013, opened 2017-02-16
Fix autocomplete behavior for Android chrome.

AndrewSouthpaw:
This is essentially the same as https://github.com/facebook/draft-js/pull/175, which was closed due to inactivity. A couple minor adjustments were made to the original PR in order to have it fit with the latest version of Draft. 

This commit fixes some issues with autocomplete behavior on Android Chrome.

## Problem 1: Keyboard loses focus

Refers to #907 and repro steps are listed there. This PR addresses the problem listed there, albeit in a different way.

## Problem 2: Replacing a word with autocomplete is broken

Steps to repro:
1. Type a few words
1. Place cursor inside a word
1. Choose a different autocomplete word
1. The resulting composition is mangled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /website/src/draft-js/lib/
 /website/src/draft-js/docs/
 npm-debug.log
+.idea/
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -42,6 +42,7 @@ const RESOLVE_DELAY = 20;
 let resolved = false;
 let stillComposing = false;
 let textInputData = '';
+let formerTextInputData = '';
 
 var DraftEditorCompositionHandler = {
   onBeforeInput: function(editor: DraftEditor, e: SyntheticInputEvent): void {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@
 /website/src/draft-js/lib/
 /website/src/draft-js/docs/
 npm-debug.log
+.idea/
```
flarnie: Is it typical to add this to your global .gitignore?

AndrewSouthpaw: Oops, sorry, that's an artifact of IntelliJ editors, I added it so config files wouldn't be thrown in with the changes. I'll remove it...

AndrewSouthpaw:
@flarnie Ping?

mathieumg:
For anyone wondering, this doesn't seem to fix #1224 (I thought it might) Tested with this PR merged in https://www.npmjs.com/package/@mathieumg/draft-js

flarnie:
Thanks for reopening this @AndrewSouthpaw and for verifying that it doesn't seem to fix #1224 @mathieumg. I'm working on setting up syncing between FB and Github today, and once that is done we will be unblocked to merge some PRs and test them internally at Facebook. Thanks to everyone for your patience.

mathieumg:
Thanks for the update @flarnie , looking forward to that sync! =) If you look in #1224, I also found out how to fix it properly since.

spirift:
How comes this PR never got merged? Seems like it would be useful to have.

PR #1023, opened 2017-02-24
Small update keybinding docs.

myyellowshoe:
**Summary**
Updated docs to show proper way to bind functions when using them inside render. By binding them in the constructor it prevents bind from getting called on each render.

Some more thoughts on proper binding:
https://daveceddia.com/avoid-bind-when-passing-props/

**Test Plan**
No Test Plan needed. Just updated docs.

### DIFF
--- a/docs/Advanced-Topics-Key-Bindings.md
+++ b/docs/Advanced-Topics-Key-Bindings.md
@@ -70,9 +70,8 @@ In our editor component, we can then make use of the command via the
 import {Editor} from 'draft-js';
 class MyEditor extends React.Component {
 
-    constructor(props) {
+  constructor(props) {
     super(props);
-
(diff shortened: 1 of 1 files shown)


Daniel15:
This is great, thanks. I definitely agree with you that this is bad, and I actually wrote the lint rule against this (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) 😃 

At Facebook we use class instance fields (way number 4 in the document you linked), binding in the constructor is fine too.

myyellowshoe:
Hah nice :)
Funny enough that's where I learned about this from :D 

I actually prefer to use class instance fields as well, as it seems more straightforward. Though I figured that binding in the constructor would have wider support, being es6.

Thanks again for contribution to the eslint, and your time working on draft!

PR #1028, opened 2017-02-25
Add .gitkeep to /website/src/lib

flarnie:
When we build the website [`fs.writeFileSync` expects the 'lib' directory to exist](https://github.com/facebook/draft-js/blob/master/website/server/convert.js#L112-L114). We previously had a `.gitkeep` in this file, even though it's also in the `.gitignore`. We could probably find a more elegant solution but for now this will fix the build.


flarnie:
Looks like there is a new error that this does not fix. Looking into that more later today.

flarnie:
Also tagging @Daniel15 in case you have ideas about the cause of the new error.

Daniel15:
Oops, I don't know why the `.gitkeep` file went missing. Sorry about that! Since I broke it, I can take a look and see what's happening :)

TravisCI was running far far behind (several hours) yesterday, so I had to build the site manually on my computer. Unfortunately that meant that I didn't have a chance to see whether the TravisCI build actually worked before going to sleep.

flarnie:
Thanks for merging this - I wish I had figured out what is causing that other error. I would expect to see `Error: EISDIR: illegal operation on a directory, read` if we were calling `fs.readFileSync` on a directory path instead of a file path, but that doesn't seem possible in any of the places we call `fs.readFileSync`.

I have some other stuff to catch up on but will look again at this too when I have time.

Issue #1011, opened 2017-02-15
Deleting composition and starting new one results in extra character appearing


flarnie:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
In Android 5.1 (not 5.0 or other versions afaik);
When you write a word, delete it, and then write another word, the first character of the previously deleted word appears after the focus/cursor.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
1) Visit https://facebook.github.io/draft-js/ on an Android 5.1 device (not 5.0 or 6.* or 7.*)
2) Enter non-whitespace characters, then press 'space'.
3) Delete all the characters entered.
4) Again enter non-whitespace characters, and press 'space'.
5) Notice that the first character of the characters you deleted earlier will appear right after the focus/cursor.

![draftjs_extra_char_bug_android_5point1](https://cloud.githubusercontent.com/assets/1114467/22955143/6b093506-f2cf-11e6-9616-09c32c5b0d81.gif)

**What is the expected behavior?**
There should be no random extra character appearing after the cursor.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

Only Android 5.1 as far as I have been able to test. I don't have access to every version of Android yet, but have tested on Android 5.0, 6.0.1, and 7.1.1. None of those reproduce this bug, but 5.0 does have a similar bug related to autocomplete - see #1010.

Guardiannw:
Is this one tied to #1010?  It seems like the same problem

flarnie:
Perhaps - looking into it. If the same fix works for both that will be most excellent. :)

flarnie:
Merging this into a general Android bug issue - see https://github.com/facebook/draft-js/issues/1895

PR #1033, opened 2017-02-27
Initial fix for text insertion bugs in Android 5.1

flarnie:
Related to issue #1011

There is a bug in Android 5.1 (not 5.0 oddly enough) that makes text
entry nearly unusable. It's documented in detail here -
https://github.com/facebook/draft-js/issues/1011

Logging everything shows that;
 - when we have a block with text and spaces, and start deleting the
   characters one by one, there is an 'onInput' for each deletion that
   updates the text.
 - when you delete the very last character, the 'anchorNode' of the
   domSelection has a type that is not a text node. So we skip updating
   at all, and that means the last character doesn't actually get
   properly deleted from editorState.
 - Then the next time something fully updates the editorState, that last
   character shows up again.

The root issue is that Android 5.1, when we tested it, was in
"composition" mode and the current handlers are not really built to
handle text deletion properly in composition mode.

We're looking forward to finding a more complete solution to the root
cause - for now this change improves the usability on mobile and doesn't
break anything.

4. Ensure that:

-

**Summary**

[...]

**Test Plan**

[...]

### DIFF
--- a/src/component/handlers/edit/editOnInput.js
+++ b/src/component/handlers/edit/editOnInput.js
@@ -47,7 +47,10 @@ function editOnInput(editor: DraftEditor): void {
   var domSelection = global.getSelection();
 
   var {anchorNode, isCollapsed} = domSelection;
-  if (anchorNode.nodeType !== Node.TEXT_NODE) {
+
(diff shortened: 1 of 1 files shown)


flarnie:
We have reviewed and tested this change internally at FB. So I think it's fine to merge.

PR #1034, opened 2017-02-27
Add text directionality override prop to DraftEditor

flarnie:
Credit for this change goes to @mmmoussa

 - Added an optional textDirectionality prop of type
   DraftTextDirectionality to the DraftEditor component which allows
   overriding the directionality of all blocks in the editor.
 - Fixed lint errors in the modified Draft.js files.

Note that we will be updating the DraftTextDirectionality type
definition to move it to the `fbjs` shared utils, possibly along with
related text directionality utilities.


flarnie:
Because this updates the docs, I might not land it until we are actually including it in a release. Or just split the documentation update into a separate PR.

flarnie:
Going to split the docs into a separate PR.

src/component/base/DraftEditorProps.js
```diff
@@ -17,6 +17,7 @@ import type {DraftBlockRenderMap} from 'DraftBlockRenderMap';
 import type {DraftDragType} from 'DraftDragType';
 import type {DraftEditorCommand} from 'DraftEditorCommand';
 import type {DraftTextAlignment} from 'DraftTextAlignment';
+import type {DraftTextDirectionality} from 'DraftTextDirectionality';
```
zpao: fbjs already has `UnicodeBidiDirection`. I just gave it a quick test in your branch and it seems to work. You should be good to go without `DraftTextDirectionality`.

```js
import type {BidiDirection} from 'UnicodeBidiDirection';
```

flarnie:
I had planned to hold off merging this until releasing v0.11.0, because folks might depend on these new props and patch versions should be backwards compatible. Going to merge and ship this anyway because -

- it's important for some users to get this out sooner
- wrapping props in a feature flag is tricky and I don't think it's worth the extra time

Issue #1010, opened 2017-02-14
auto-correct creates bad state in Android 5.0


flarnie:
**Do you want to request a *feature* or report a *bug*?**
a bug

**What is the current behavior?**
Entering text that triggers auto-correct and then deleting it leaves things in a weird state.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
You can reproduce this on Android 5.0 on the 'try draft' demo page.
1) enter 'aaa' and then hit 'space'. Notice autocorrect will change it to something.
2) delete all those characters. Notice it deletes it all at once and unfocuses the editor.
3) enter 'bbb' and hit 'space'. Notice the previously deleted characters reappear.

Disabling 'autocorrect' and 'spellcheck' in android 5.0 chrome either doesn't work or doesn't fix this bug.

**What is the expected behavior?**
Characters should be gone forever if we delete them, unless we 'undo'.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

Draft 0.9.1
Draft 0.10.0

Android 5.0 chrome
(thankfullly not Androit 6.0 or 7.0 I don't think)

flarnie:
![android5_autocorrect_bug](https://cloud.githubusercontent.com/assets/1114467/22945663/d340e4fa-f2a9-11e6-8ae6-a39cbff367ed.gif)

Guardiannw:
I have seen this as well

flarnie:
Note - I'm only seeing this in Android 5.0.2. Android 5.1 shows issue #888 instead.

Guardiannw:
I am on android 5.1.0 and chrome 55 and am seeing this issue, not #888.

will-hart:
I have this issue in Chrome 55 on Android 7.1.1, Nexus 6P. Are this issue and #1011 duplicates of #911?

flarnie:
Basically we don't have resources to tackle Android support right now - still open to PRs for this, but not actively working on mobile web support.

flarnie:
Merging this into a general Android bug issue - see https://github.com/facebook/draft-js/issues/1895

PR #1035, opened 2017-02-27
Expose props for disabling auto-correct, auto-complete, auto-capitalize

flarnie:
Wanted: manual testing of this PR.

There is some really weird behavior in Android 5.0 when auto-correct happens.
More details here;
https://github.com/facebook/draft-js/issues/1010

Setting 'spellCheck' to 'false' didn't seem to be enough to stop this.

My quick fix for now is to expose more properties on Draft for
controlling autocorrect. This could also be useful to people in general,
since there are often complaints about autocorrect on mobile web.

Long term we will be addressing the root causes of the bugs with mobile
web, but this seems like a useful change regardless.

We did some testing with mobile web using these props.

Ideally I'd like to get testing done of each prop in each supported
browser, and document the effects. If anyone can help verify different
behaviors please comment here, otherwise I'll get to it soon.

Note that we'll update the documentation once this change is actually
included in a released version of Draft.js.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -243,6 +243,9 @@ class DraftEditor extends React.Component {
             aria-haspopup={readOnly ? null : this.props.ariaHasPopup}
             aria-label={this.props.ariaLabel}
             aria-owns={readOnly ? null : this.props.ariaOwneeID}
+            autoCapitalize={this.props.autoCapitalize}
+            autoComplete={this.props.autoComplete}
+            autoCorrect={this.props.autoCorrect}
             className={cx('public/DraftEditor/content')}
             contentEditable={!readOnly}
             data-testid={this.props.webDriverTestID}
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -71,6 +71,11 @@ export type DraftEditorProps = {
 
   tabIndex?: number,
 
+  // exposed especially to help improve mobile web behaviors
+  autoCapitalize?: string,
+  autoComplete?: string,
+  autoCorrect?: string,
+
   ariaActiveDescendantID?: string,
   ariaAutoComplete?: string,
   ariaDescribedBy?: string,


flarnie:
I had planned to hold off merging this until releasing v0.11.0, because folks might depend on these new props and patch versions should be backwards compatible. Going to merge and ship this anyway because - 
- it's important for some users to get this out sooner
- wrapping props in a feature flag is tricky and I don't think it's worth the extra time

leanazulyoro:
@flarnie I'm beggining to test this PR. I have a question:
setting these 3 attributes to "off" should be enough for this behavior to get fixed?
This is what I did after merging this PR:

```jsx
<Editor 
   onChange={this.onChange}
   plugins={plugins}
   placeholder={this.props.placeholder}
   stripPastedStyles={true}
   spellCheck="false"
   autoCapitalize="off"
   autoComplete="off"
   autoCorrect="off"
   ref={(element) => { this.editor = element; }}
/>
```
I tested it in my android phone and it didn't get fixed.

CharlesKuncheria:
@flarnie using GBoard with 'Show Suggestion Strip' setting ON, doesn't fires 'onchange' event, but the editor state doesn't seem to change. 

In the below code, the div that mirrors the editor's new state, is updated only when we add a space (not while typing). 
But, it behaves as expected when the 'Show Suggestion Strip' is turned OFF.

Device: Pixel 3a
Keyboard: GBoard

```
class TestEditor extends React.Component {
  constructor(props) {
    super(props);
    this.state = { editorState: EditorState.createEmpty() };
    this.onChange = (editorState) => {
      this.setState({ editorState });
    }
    this.keyBinding = (e) => {
      return getDefaultKeyBinding(e);
    }
    this.setEditor = (editor) => {
      this.editor = editor;
    };
    this.focusEditor = () => {
      if (this.editor) {
        this.editor.focus();
      }
    };
  }

  componentDidMount() {
    this.focusEditor();
  }

  render() {
    return (
      <div>
        <div>
          Editor State: {this.state.editorState.getCurrentContent().getPlainText()}
        </div>
        <div style={styles.editor} onClick={this.focusEditor}>
          <Editor
            ref={this.setEditor}
            onChange={this.onChange}
            keyBindingFn={this.keyBinding}
            tabIndex={0}
            textAlignment={'left'}
            **autoCapitalize="off"
            autoComplete="off"
            autoCorrect="off"**
          />
        </div>
      </div>
    );
  }
}
```

Issue #1018, opened 2017-02-21
Pressing Enter Immediately Reverts IE


colinjeanne:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
On Windows systems with performance problems or when performing expensive operations in Draft (such as with a decorator) pressing the enter key will sometimes immediately revert itself. In the following `hithere <enter>` is quickly typed but after pressing `<enter>` the new line is reverted and the `e` is added:

![jan-30-2017 22-19-13](https://cloud.githubusercontent.com/assets/7006910/22454738/ff855230-e73c-11e6-923c-eaf9b9b631b6.gif)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**

While the image is not from core Draft, it is possible to reproduce on https://facebook.github.io/draft-js but significantly more difficulty. The image repros it easily because there is a good amount of processing happening on change which slows IE perf down enough to hit this issue.

**What is the expected behavior?**

The `<enter>` will not revert itself.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

IE 11, Draft 0.10

@max-winderbaum did the investigation on our side and discovered the root cause:

1. Type `a`.
2. It's a simple character. Native insertion is allowed `onBeforeInput`
3. Native insertion does a `setImmediate()` call to update the state because IE won't fire the `input` event (see below)
-  Normal browsers fire the `input` event which clears the `_pendingStateFromBeforeInput` and when the `setImmediate()` fires it noops
-  IE relies on the `setImmediate()` completing to rectify the DOM change with its internal state
4. Before the `setImmediate()` fires, the user presses `<enter>`. This action is not a "native insertion" event and the state is updated immediately.
5. The `setImmediate()` fires and the content is replaced with the state before the `<enter>` was pressed

This is fixed by https://github.com/textioHQ/draft-js/pull/24 in our branch. My plan is to port this change to facebook/draft-js to make our solution to https://github.com/facebook/draft-js/issues/989 easier to merge.

ssivagnanam:
i am also facing same issue.  any update on this issue?

Issue #989, opened 2017-01-31
Lost position of the cursor when editing


cestrensem:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
User already has some text in the text area. He puts cursor with his mouse somewhere and quickly types something, then again quickly puts cursor to other place in the text area and again quickly types something. After these several "puts and types" the place where the cursor is and the place where the typed text appears are not in sync anymore. The cursor is in one place, the text appears in another place of the text area.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: 
https://youtu.be/RZUzLyvLgnY

**What is the expected behavior?**
Text appears on the same position where the cursor is.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
Found on 0.10.0 & 0.9.1. 

Browsers:  
- Chrome 55.0.2883.95 (64-bit) Mac OS Sierra 11.02.12
- Safari 10.0.2 Mac OS Sierra 11.02.12
- Firefox 49.0.0 Mac OS Sierra 11.02.12
- IE 11 WIndows 10
- EDGE 13 Windows 10

flarnie:
Hi @cestrensem - thanks for reporting this and making the recording of the problem! This definitely does not look good, and I will look into it as soon as I get a chance. I'd also be happy to review any PR that has a solution.

It looks like that is a large area with many blocks, and since we have seen performance issues with rendering a large number of text blocks I wonder if that is contributing to this problem also.

colinjeanne:
Having a large number of blocks seems to make this easier to repro but I find it pretty easy to repro with just two blocks with a few words each (and a decorator that does a small amount of work). I'm trying to investigate this and it seems - so far - that the issue is that a call to `getUpdatedSelectionState` appears to be returning the wrong selection when typing quickly causing Draft and the browser to disagree where the caret actually is.

This repros only when native insertion is allowed, which makes sense. The issue is fundamentally that Draft and the browser are disagreeing about where selection is. The browser inserts text in one location and Draft inserts it in another.

This is not, however, the result of the `setImmediate` call in `editOnBeforeInput`. We have a local branch which removes this call due to another issue which we have yet to file on Facebook's Draft repo and this issue still occurs in that branch.

colinjeanne:
[...]

PR #1042, opened 2017-02-28
Ensure selection is always valid

max-winderbaum:
**Summary**

Fixes https://github.com/facebook/draft-js/issues/1018
Fixes https://github.com/facebook/draft-js/issues/989

There are a couple fundamental problems with the typing loop and selection in Draft:

**React doesn't fire a `handleSelect` event until `mouseUp`**
This creates the following bad scenario:

1. Create two blocks with text
2. Place selection in the first block
3. Click on the second block to change selection, but hold the mouse down
4. Type a letter
5. Draft still thinks the selection is on the first block and performs its `onInput` block-matching logic, which makes wacky things happen like the blocks duplicating. Things quickly get out of sync.

**Internet Explorer has a timing window around using `setImmediate` to fake input events**
See https://github.com/facebook/draft-js/issues/1018

The core of this PR is based on two things:
1. Sync draft's internal selection with `window.getSelection()` more often.
2. Update the editorState optimistically `onBeforeInput` so as to not let the internal state get out of sync with the DOM. If something happens between `onBeforeInput` and `onInput` to change the optimistically-updated editorState, we fix that up `onInput`

**Test Plan**

I'd love to hear from the folks at Facebook (@flarnie?) about how to best test these scenarios. This code has been in production at Textio for about a week without issue, and I'm not seeing many existing tests in this area of code.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -21,6 +21,7 @@ const DraftEditorContents = require('DraftEditorContents.react');
 const DraftEditorDragHandler = require('DraftEditorDragHandler');
 const DraftEditorEditHandler = require('DraftEditorEditHandler');
 const DraftEditorPlaceholder = require('DraftEditorPlaceholder.react');
+const ContentBlock = require('ContentBlock');
 const EditorState = require('EditorState');
 const React = require('React');
 const ReactDOM = require('ReactDOM');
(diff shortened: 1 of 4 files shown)


max-winderbaum:
One other thing to note about this fix is that we didn't do the (probably complicated) work to fix up the selection post-spellcheck because we're in the process of implementing our own spellcheck. The selection is mostly ok, but sometimes can be wonky depending on stray entities.

max-winderbaum:
The last commit includes a solution that retains post-spellcheck selection fixup capability :)

michaelcox:
Why was this PR closed? It appears to be the only existing attempt at a solution to #989.

againksy:
@max-winderbaum Hello, dear what about issue with jumpg cursor ? https://github.com/facebook/draft-js/issues/989 how to fix it ?  the cursor jumps when beginning to enter text, jumps before first letter, and if we press backspace it jumps to beginning of line , so what is the solution, maybe anyone knows?

PR #1057, opened 2017-03-07
Ability to select which inline styles, block types, links should be retained on paste

### LINKED ISSUE #166: Ability to select what to strip in paste
Right now, you can choose whether your editor accepts rich text or not in paste using `stripPastedStyles`. It's not always all-or-nothing, though: in case your editor only supports a subset of the supported styles, pasting can insert things the user can't remove anymore. For example, the Facebook Notes editor doesn't support underline or strikethrough, and yet you can paste underlined/striked text, and you can't get rid of that style anymore without deleting the entire piece of text.

It would be handy to have a prop on the editor component to specify which inline styles should be retained on paste.

mkhazov:
fixes #166

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -119,11 +119,20 @@ needed to observe spellcheck events are not fired in IE.
 
 Default is `false`.
 
-#### stripPastedStyles
+#### pasteSupport
 ```
-stripPastedStyles?: boolean
+stripPastedStyles?: Object
 ```
-Set whether to remove all information except plaintext from pasted content.
+Optionally define which inline styles, block types and links should be retained on paste.
+
+If your use case should not have any blocks, inline styles and links, set this to
+```
+{
+  inlineStyles: []
+  blockTypes: [],
+  links: false,
+}
+```
 
 This should be used if your editor does not support rich styles.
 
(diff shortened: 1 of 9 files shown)


docs/APIReference-Editor.md
```diff
@@ -119,11 +119,20 @@ needed to observe spellcheck events are not fired in IE.
 
 Default is `false`.
 
-#### stripPastedStyles
+#### pasteSupport
 ```
-stripPastedStyles?: boolean
+stripPastedStyles?: Object
```
colinjeanne: Minor: this is still using the old new

src/model/immutable/DraftPasteSupport.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+'use strict';
+
+import type {List} from 'immutable';
+
+export type DraftPasteSupport = {
+  inlineStyles: List<string>,
+  blockTypes: List<string>,
+  links: boolean,
```
colinjeanne: If we're going to go this far, why not also allow images to be blocked?

docs/APIReference-Editor.md
```diff
@@ -119,11 +119,20 @@ needed to observe spellcheck events are not fired in IE.
...
+```
+Optionally define which inline styles, block types and links should be retained on paste.
```
colinjeanne: We should go into more depth here to make the documentation easier to use. For example, we should clarify that the supported block types is the intersection between those defined in the block render map and those defined in this structure. We should provide an example, non-empty, value for this field as well so that users don't need to spelunk through the source to find how each field is being interpreted and used.

mkhazov:
@colinjeanne thanks, I've added ability to strip images and updated the docs.

PR #1065, opened 2017-03-11
Correctly delete immutable and segmented entity content when at the edge of a selection

### LINKED ISSUE #1063: Immutable and segmented entity content is not deleted correctly when at the edge of a selection
Bug

When deleting after having made a selection that contains part of an immutable or segmented entity, the entity status of the entity is correctly removed but the text within it is not deleted.

Visit the draft.js entity example here: http://codepen.io/Kiwka/pen/wgpOoZ. Using the cursor, make a selection that includes part of the immutable "Superman" entity and some of the text that follows it (eg. "erman. Dele"). Click backspace.

The text within the entity should be deleted in accordance with the entity's deletion rules. In the case of the immutable entity example, all text within the entity should be deleted.

All versions of Draft.js are affected on all browser / OS combinations. This bug did not work previously.

mmmoussa:
There's a bug with deletion of segmented and immutable entities when they are not fully selected and only part of a selection. The bug is documented in #1063.

This fixes #1063 so that when deleting (such as through backspace) after having made a selection that contains part of an immutable or segmented entity, the entity text deletion is now correctly handled.

### DIFF
--- a/src/model/modifier/DraftModifier.js
+++ b/src/model/modifier/DraftModifier.js
@@ -138,29 +138,49 @@ var DraftModifier = {
     rangeToRemove: SelectionState,
     removalDirection: DraftRemovalDirection
   ): ContentState {
+    let startKey, endKey, startBlock, endBlock;
+    startKey = removalDirection === 'forward'
+      ? rangeToRemove.getAnchorKey()
+      : rangeToRemove.getFocusKey();
+    endKey = removalDirection === 'forward'
+      ? rangeToRemove.getFocusKey()
+      : rangeToRemove.getAnchorKey();
+    startBlock = contentState.getBlockForKey(startKey);
+    endBlock = contentState.getBlockForKey(endKey);
+    const startOffset = rangeToRemove.getStartOffset();
+    const endOffset = rangeToRemove.getEndOffset();
+
+    const startEntityKey = startBlock.getEntityAt(startOffset);
+    const endEntityKey = endBlock.getEntityAt(endOffset - 1);
+
     // Check whether the selection state overlaps with a single entity.
     // If so, try to remove the appropriate substring of the entity text.
-    if (rangeToRemove.getAnchorKey() === rangeToRemove.getFocusKey()) {
-      var key = rangeToRemove.getAnchorKey();
-      var startOffset = rangeToRemove.getStartOffset();
-      var endOffset = rangeToRemove.getEndOffset();
-      var block = contentState.getBlockForKey(key);
-
-      var startEntity = block.getEntityAt(startOffset);
-      var endEntity = block.getEntityAt(endOffset - 1);
-      if (startEntity && startEntity === endEntity) {
-        var adjustedRemovalRange = getCharacterRemovalRange(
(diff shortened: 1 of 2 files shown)


src/model/modifier/DraftModifier.js
```diff
@@ -138,29 +138,49 @@ var DraftModifier = {
...
-      var endEntity = block.getEntityAt(endOffset - 1);
-      if (startEntity && startEntity === endEntity) {
-        var adjustedRemovalRange = getCharacterRemovalRange(
+    if (startKey === endKey) {
+      if (startEntityKey && startEntityKey === endEntityKey) {
+        const adjustedRemovalRange = getCharacterRemovalRange(
           contentState.getEntityMap(),
-          block,
+          startBlock,
+          endBlock,
```
flarnie: Interesting that we pass the 'startBlock' and 'endBlock' now, when before we only passed one block into this method. ?

flarnie: nvm - I see now you updated the method parameters. Should have finished reading the changes.

flarnie:
This makes sense!
My only hesitation is that this "bug" has been the accepted behavior for entity deletion since the start of Draft.js, and this change might be perceived as a bug in some use case.
I'm going to ask around and try to verify that it won't break anything at FB - it seems more correct than the current behavior.

Also I still need to manually test this change. Will do that today.

flarnie:
Sorry to come back with a different answer - after talking to product folks, it seems that this change would contradict the desired behavior they have for entities. And folks outside of FB may have cases where this change would be perceived as a bug.

For now let's look for other ways to address this.

flarnie:
And updating this again - we talked more and determined that this only affects a few cases, so we'll try it out. Glad to be able to merge this after all. :)

sarmstrong:
Is this functionality still under evaluation? And if so, what's the best way to enable the feature flag? We're having issues w/ this functionality. Thanks!

flarnie:
Hi @sarmstrong - 

This PR was merged but then reverted due to bugs it introduced (https://github.com/facebook/draft-js/pull/1095), and then re-introduced under a feature flag in https://github.com/facebook/draft-js/pull/1104. The feature flag is 'draft_segmented_entities_behavior' and you can enable it by changing this value from 'false' to 'true' in your local copy of 'draft-js'; https://github.com/facebook/draft-js/blob/master/src/component/utils/DraftFeatureFlags-core.js#L18

Also @mmmoussa kindly made some improvements to the behavior (https://github.com/facebook/draft-js/pull/1108), so I assume the main issues with the original PR are resolved.

We have a system that shims the `DraftFeatureFlags` file and reads from our own internal configs, allowing us to experiment with having those feature enabled or disabled at Facebook. I don't think we have enabled the 'draft_segmented_entities_behavior' widely internally yet, and we would want to do that before making the default 'true' and removing the flag. Will get a status update on this when I can.

sarmstrong:
Ah, got it, thanks for the update @flarnie! We've got a hack to handle this currently but it's not full proof, this fix would most likely solve it. We'll be on the look out.

Issue #1066, opened 2017-03-11
Backspacing and deletion of immutable/segmented entities handled incorrectly with some Android keyboards


mmmoussa:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When backspacing using some Android keyboards such as the Google keyboard, no key events are fired and no key codes are provided by the input event data when editOnInput gets called. This is an issue with the keyboard and is not likely to be fixed in the near future, if at all. As a result of this issue, backspacing cannot currently be detected before the editOnInput event fires and is being handled incorrectly. Among other problems, the selection state being used in editOnInput is incorrect and there is no support for deleting entities.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
This bug is reproducible on Android phones using the Google keyboard. It can be reproduced by inserting an immutable entity into the editor and then trying to delete it by backspacing into it.

**What is the expected behavior?**
The expected behaviour is that backspacing is handled the same on Android devices as it is with other devices, including entity deletion.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
This issue affects all versions of Draft.js on Android.

flarnie:
Merging this into general android bug issue - see https://github.com/facebook/draft-js/issues/1895

PR #1067, opened 2017-03-11
Add partial backspace support to editOnInput for Android devices

mmmoussa:
**Summary**

This pr addresses two parts of the problem raised in #1066. 

The first part is that it adds support for entity deletion within editOnInput. It does this by detecting whether the start of the selection is a mutable entity, and if it's not, falling back to the backspace method normally used as part of the keyDown event handler.

The second part is that it identifies cases where the selection is incorrect and then corrects it before falling back on the backspace method. The correction is performed by using the end of the DOM's selection as the start of the selection to evaluate, and the end of the last editor state's selection as the end of the selection to evaluate.

**Known issues**

The backspace selection range cannot be correctly determined when the starting selection edge is in the middle of a span. In such cases, the span containing the starting selection edge does not have any of its contents deleted, and temporarily the DOM selection understanding is incorrect until the cursor is moved.

### DIFF
--- a/src/component/handlers/edit/editOnInput.js
+++ b/src/component/handlers/edit/editOnInput.js
@@ -12,20 +12,22 @@
 
 'use strict';
 
+import type DraftEditor from 'DraftEditor.react';
+
 var DraftModifier = require('DraftModifier');
 var DraftOffsetKey = require('DraftOffsetKey');
 var EditorState = require('EditorState');
 var UserAgent = require('UserAgent');
 
 var findAncestorOffsetKey = require('findAncestorOffsetKey');
+const keyCommandPlainBackspace = require('keyCommandPlainBackspace');
 var nullthrows = require('nullthrows');
 
-import type DraftEditor from 'DraftEditor.react';
+const isGecko = UserAgent.isEngine('Gecko');
+const isAndroid = UserAgent.isPlatform('Android');
 
-var isGecko = UserAgent.isEngine('Gecko');
 
 var DOUBLE_NEWLINE = '\n\n';
-
 /**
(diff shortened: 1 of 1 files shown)


flarnie:
"It does this by detecting whether the start of the selection is a mutable entity, and if it's not, falling back to the backspace method normally used as part of the keyDown event handler."
I'd like to understand that part of the change more.
It's true that in android, when the user hits 'delete', that triggers the 'editOnInput' callback, and in that case we could just call the 'keyCommandPlainBackspace' method.
- What about the other cases where 'editOnInput' gets called? It gets used for a bunch of things. Maybe we should split it into some separate handlers - it's getting convoluted I think.
- Why would we not call 'keyCommandPlainBackspace' when there is a mutable entity present?

flarnie:
So some ideas from our meeting:
- It seems like the core issue is that Draft tries to listen to keyDown and it's not there on delete in Android with that keyboard. We could file a bug and/or file a bug and push to have it fixed.
 - we could try to write unit tests around this; that might make the issues more clear and help us find other changes that will fix the problem
    - for inspiration look at the DOM selection unit tests
 - let's be careful and understand this behavior before merging this or any similar change
 - let's be intentional about which conditional blocks we gate to Android; that might warrant more discussion
     - we might look at the way we handle 'cut' where we let the DOM get mutated and then force a re-render to get it back into a good state

Hope that is helpful, and sorry that we're not prepared to merge this change.

flarnie:
Since we never resolved the questions and problems with this PR, and nobody is currently working on fixing the known issues with Android web, I'm closing this.

PR #1069, opened 2017-03-13
Change website main font-weight to normal.

Lyonsclay:
This pull request changes the base font-weight in the website docs from 300 to normal. These old eyes find it easier to read and it is the same font-weight as the React website docs.


tylercraft:
Thanks!

Issue #473, opened 2016-06-16
Updating dual editorStates not working.


jussch:
My scenario: I have two draftjs editors. One is the core input, and the other is a helper input which wraps its input in an entity and adds it to the core input. The core input and the helper input work as intended when the helper only adds to the core input on blur. The problem arrises when I tried to get the helper input to update the core input live, (updating both `editorStates` on `onChange` of the helper input) I receive this error:

```
getUpdatedSelectionState.js:32 Uncaught TypeError: Cannot read property 'getIn' of undefined
  getUpdatedSelectionState @ getUpdatedSelectionState.js:32
  getDraftEditorSelectionWithNodes @ getDraftEditorSelectionWithNodes.js:35
  getDraftEditorSelection @ getDraftEditorSelection.js:33
  editOnSelect @ editOnSelect.js:26
  (anonymous function) @ DraftEditor.react.js:157
  ReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:70
  executeDispatch @ EventPluginUtils.js:87
  executeDispatchesInOrder @ EventPluginUtils.js:110
  executeDispatchesAndRelease @ EventPluginHub.js:42
  executeDispatchesAndReleaseTopLevel @ EventPluginHub.js:53
  forEachAccumulated @ forEachAccumulated.js:24
  processEventQueue @ EventPluginHub.js:215
  runEventQueueInBatch @ ReactEventEmitterMixin.js:18
  handleTopLevel @ ReactEventEmitterMixin.js:29
  handleTopLevelImpl @ ReactEventListener.js:73
  perform @ Transaction.js:136
  batchedUpdates @ ReactDefaultBatchingStrategy.js:63
  batchedUpdates @ ReactUpdates.js:98
  dispatchEvent @ ReactEventListener.js:150
```

I've boiled down my code as much as possible to find the perpetrator, and I've believe I've found the issue. Updating the `editorState` of the core Editor through the helper's `editorState` while the core Editor has `readOnly={true}` triggers the error. (And without `readOnly={true}`, the core editor constantly takes focus from the helper input.)

Here is the code to re-produce the error.

``` javascript
export default class DualInput extends Component {
  constructor(props) {
    super(props);
    this.setHelperState = this.setHelperState.bind(this);
    this.state = {
      readOnly: false,
      editorState: EditorState.createEmpty(),
      helperState: EditorState.createEmpty(),
    };
  }

  setHelperState(helperState) {
    this.setState({
      helperState,
[...]

PR #1084, opened 2017-03-17
Updating selection between rendering can crash

colinjeanne:
There is a timing window that exists from when a new editorState is passed into draftjs as a prop until that new state is rendered. If an onSelect event fires in this time, then the code will throw because the editorState doesn't yet match the DOM state. getUpdatedSelectionState should detect this case and just return editorState.getSelection() in this scenario.

This is a port of https://github.com/textioHQ/draft-js/pull/2 and fixes https://github.com/facebook/draft-js/issues/473. #473 is closed because the original poster no longer repros the issue but others in Slack have said they are hitting the issue and that this fixes their repro.

### DIFF
--- a/src/component/selection/getUpdatedSelectionState.js
+++ b/src/component/selection/getUpdatedSelectionState.js
@@ -43,8 +43,11 @@ function getUpdatedSelectionState(
 
   var anchorPath = DraftOffsetKey.decode(anchorKey);
   var anchorBlockKey = anchorPath.blockKey;
-  var anchorLeaf = editorState
-    .getBlockTree(anchorBlockKey)
+  var anchorBlockTree = editorState.getBlockTree(anchorBlockKey);
+  if (!anchorBlockTree) {
+    return selection;
+  }
+  var anchorLeaf = anchorBlockTree
     .getIn([
       anchorPath.decoratorKey,
       'leaves',
@@ -53,8 +56,11 @@ function getUpdatedSelectionState(
 
   var focusPath = DraftOffsetKey.decode(focusKey);
   var focusBlockKey = focusPath.blockKey;
-  var focusLeaf = editorState
-    .getBlockTree(focusBlockKey)
+  var focusBlockTree = editorState.getBlockTree(focusBlockKey);
+  if (!focusBlockTree) {
+    return selection;
+  }
+  var focusLeaf = focusBlockTree
     .getIn([
       focusPath.decoratorKey,
       'leaves',


cameackeSEL:
+1 this is a major problem for us, and the fix is simple. Please merge in soon! :)

aligon:
I've also been getting this error. Any word on this pull request?

thickman:
+1 this bug is very annoying. Can you merge the fix please?

adamskihood:
+1 this bug is very annoying. Can you merge the fix, please?

juliankrispel:
Please merge this @facebook-github-bot - do your magic

flarnie:
thanks for everyone's patience, I'll try to look at this over the weekend, and pinging other maintainers to see if anyone can verify it sooner

juliankrispel:
@flarnie thank you so much <3

juliankrispel:
@flarnie here take this delicious slice of cake 🍰  and a beer :beer:

juliankrispel:
@flarnie while you're at it would you mind looking at this as well https://github.com/facebook/draft-js/pull/1192

juliankrispel:
@flarnie any luck? :)

flarnie:
@juliankrispel thanks for your patience - want to get Github and facebook's Draft copies in sync before making any further changes to either, and found they had diverged more than I realized. Will be talking to folks today about how to get them to auto-sync, so that we can merge things here without creating extra work.

Topperfalkon:
@flarnie Don't suppose there's been any progress with this?

adamskihood:
Any update on that @flarnie?

birdwell:
Are there any new developments on this?

Topperfalkon:
@flarnie any progress with this?

saharj:
Any update here?

flarnie:
Thanks everyone who has pinged this - it seems like a problem worth solving.

**tldr:** [race conditions are a known pitfall](https://draftjs.org/docs/advanced-topics-editorstate-race-conditions.html#content) with Draft.js and I'd like to understand the problem more before taking a change like this.

I'd *really* like to get master back in sync with the [textioHQ fork of Draft.js](https://github.com/textioHQ/draft-js), which seems to have been ahead of the curve with bug fixes in the past. So @AnilRedshift @max-winderbaum @colinjeanne maybe you can help me connect the dots here.

In [the example repro case from that issue](https://github.com/jussch/DraftJs-DualEditor) with two editors, it looks like there is an 'onSelect' event which fires and somehow gets an `editorState` which has a ref to the DOMNode for the top editor, but the BlockTree from the second editor.

I'm trying to reconcile that with the description of this PR:

> "There is a timing window ..."

-> in the repro example, it's not a window of time. It just happens automatically once you get the `editorState` into this broken state with the mismatch between the `ref` and the BlockTree.

> "the code will throw because the editorState doesn't yet match the DOM state." 

-> The DOM state always looks correct to me when I inspect it - the `data-offset-key` is unchanged in editor1 and editor2, and gets identified correctly in the event handler. It seems like somehow we are getting mixing between the two editorStates though, which could be a problem with Draft.js, or a pitfall that is easy to fall into perhaps.

If the `editorState` gets mixed up between two editors, we need to fix the root cause. I would like to get more info about how that happens and a PR to fix the root cause, if possible.

AnilRedshift:
Hi Flarnie,

With the caveat that I haven't worked at Textio since March, I'll add my recollections below. Hopefully Colin or Max can help out with the latest of @textio/draft-js

The problem here is not that you have two editors, but one editor with diverging editorStates.

At its core, the issue is this: The editorState changes at a different time than when the DOM changes. At some point after `render()` only then does the editorState again reflect what is in the DOM. If you end up with any events (such as selection, or otherwise) that come in before this synchronization, you end up trying to relate DOM nodes to a different editorState.

Having the time & space to think about this more, I think a more robust solution than the fix I proposed would be to tag the editorState with an incrementing key on each update, and then saving that key to the DOM. Then, when you get a new event, you can compare and see if that event happened while the DOM was out of sync. If so, you can drop the event (or take other actions)

This issue isn't seen as much when just typing normally, because draft-js allows the browser to update the DOM when possible. Even otherwise, draft-js reuses most of the nodes so there is only one area where this timing window is possible.

On the other hand, if you replace the entire editorState with a different editorState, then there's a better chance to get an event in the middle that doesn't map.

Does this help?

colinjeanne:
@flarnie: I would love to bring https://github.com/textioHQ/draft-js back in sync with master and have a few of our changes in PR here. The largest is that our fix for #989 has diverged quite a bit from what exists in the code (in particular, we don't have https://github.com/facebook/draft-js/commit/1c6a49b8801183fe0c29458626c0b5dbe1238e59#diff-638c3e5663e5bd194aefe231001f51f2 or https://github.com/facebook/draft-js/commit/5863399a3a1bcbbe9b090249504a70496a7af7cc#diff-638c3e5663e5bd194aefe231001f51f2 and we haven't had the time to investigate how to merge this into our branch).

Otherwise @AnilRedshift's recollection of the issue matches mine. I don't recall the repro because we unfortunately didn't describe it in our bug and my memory of it had already been lost when I originally opened this.

flarnie:
Thanks both @AnilRedshift and @colinjeanne for responding! That is helpful - I still would like to get a repro, and ideally even a test, for this bug. It seems like what you are describing is related to [the "DualEditor" example](https://github.com/jussch/DraftJs-DualEditor) but not quite the same.

Hoping I'll have time to look at this more on the weekend.

colinjeanne:
@flarnie: as I recall this was very timing related. Our editor has a rather complex decorator which likely increased the timing window for us but I wonder if a simple repro might be to simply move the DOM selection manually in a call to render.

kaylee42:
Just wanted to leave a note here as I have been following issue #473 and this PR was brought up - I'm dealing with this issue and tried out this fix on my codebase. However, it had the impact of deleting image blocks when they were clicked and the editor did not previously have focus (though, I didn't get the `getIn` error anymore 😝 ).

shpakkdv:
Faced with this issue. Crash in Safari (only) when Draft.js is updating the editor state.
My solution / hack:
```js
  const nativeSelectionExtend = Selection.prototype.extend;

  Selection.prototype.extend = function (...args) {
    try {
      return nativeSelectionExtend.apply(this, args);
    } catch (error) {
      console.log('Selection error.', error);
    }
  };
```

It works properly for me. Maybe will be useful for somebody as well.

PR #1103, opened 2017-03-29
Add clarifying comments to 'getRemovalRange'

flarnie:
I found some of this code confusing and added comments. Will probably
add more comments while looking at this part of the code.

For now pushing this before I lose track of things. Not in any hurry to merge it.


flarnie:
Probably need to rebase this. Will get around to it in a bit.

PR #1104, opened 2017-03-29
Add Feature Flags in Draft, and add Entity Deletion change under Flag

flarnie:
We use feature flags in the FB version of Draft to test new and
potentially unstable changes. This mirrors out the flags we currently
have and sets the stage for keeping things in sync going forward.

Now internal use of feature flags will not block syncing and releasing
Draft.

This is part of an effort to make releases simpler and more frequent,
and keep the internal and external versions of Draft in sync.

As part of matching up the internal and external Draft flags we added
the changes from PR #1065 by @mmmoussa, with the main change under a
flag that is set to false.

---Test Plan
Manually tested the examples in Chrome, and we will do the full test
plan before releasing v0.10.1.

### DIFF
--- a/src/component/handlers/edit/editOnInput.js
+++ b/src/component/handlers/edit/editOnInput.js
@@ -12,6 +12,7 @@
 
 'use strict';
 
+const DraftFeatureFlags = require('DraftFeatureFlags');
 var DraftModifier = require('DraftModifier');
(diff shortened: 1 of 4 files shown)


flarnie:
Merging this in order to unblock doing a sync and (hopefully) a release.

Issue #862, opened 2016-12-10
Enable dangling comma lint


flarnie:
It doesn't seem that ESLint has been catching missing dangling commas, and we would like to be consistent with the Facebook ESLint configurations.
Edit: It would be also nice to warn for lines over 80 characters long.

davidchang:
looks like comma-dangle is OFF and maxLen is set to 120 here: https://github.com/facebook/fbjs/blob/master/packages/eslint-config-fbjs/index.js#L310

we could specifically change these in draft's own eslintrc, but not sure if we would want to do that. quick glance around other FB JS projects shows everyone's kinda doing their own thing, not just using eslint-config-fbjs

@zpao - can you advise?

mmmoussa:
Once #1126 gets merged I think this issue can be closed.

PR #1107, opened 2017-03-30
Fix missing commas

mmmoussa:
Updated lint rules to catch missing dangling commas in multiline function syntax. This brings the open-source project closer to being in line with Facebook's internal copy, making syncing them easier. This partially addresses #862.


flarnie:
Thanks again for doing this update!

PR #1109, opened 2017-03-31
Add bundles from 'universal' example to eslintignore

flarnie:
After running the build step for the 'universal' example there is a bundle file which we don't want to lint. This fixes the issue.

Steps to reproduce the problem:
 - install dependencies and run `npm run demo` in the universal example directory.
 - then run `npm run lint` and see the errors from linting the `bundle.js` file

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,5 @@ node_modules/
 website/
 gulpfile.js
 scripts/module-map.js
+examples/draft-0-10-0/universal/static/bundle.js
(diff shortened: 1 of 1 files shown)


flarnie:
I really am not seeing the reason the tests failed. I even updated to the exact node version that travis is using, and still they all pass for me locally.

flarnie:
Will do a follow-up that:
 - merges the `universal/.gitignore` with the top level .gitignore
 - adds the `yarn.lock` file in `universal` to the top level .gitignore

Issue #1120, opened 2017-04-05
Issue with undo and custom block components


mshibl:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**steps to reproduce?**
1- Add new custom component
2- Undo last step (using CMD+Z or CTRL+Z) 
--> cursor will jump to the far right of the screen
3- Try to type anything 
--> Cursor will type the first character normally
--> next letter will jump to next line, while cursor is still on the first line
--> next letter, all text on the first line is copied to next line

**Demo**
![Issue](http://g.recordit.co/70qiSzEsj2.gif)

**Which versions of Draft.js, and which browser / OS are affected by this issue?**
"draft-js": "0.10.0" - Chrome

mshibl:
This same issue can be seen when running the [TeX Editor](https://github.com/facebook/draft-js/tree/master/examples/draft-0-10-0/tex) example in the examples/draft-0-10-0 folder

flarnie:
Thanks for documenting this - it does indeed reproduce in the TeX Editor and is still an issue in Draft 0.10.1. I saw this error thrown:
```
app.js:26460 Uncaught TypeError: Cannot read property 'replaceChild' of null
    at Function.replaceChildWithTree (app.js:26460)
    at Object.dangerouslyReplaceNodeWithMarkup (app.js:26897)
    at Object.dangerouslyReplaceNodeWithMarkup [as replaceNodeWithMarkup] (app.js:26284)
    at ReactCompositeComponentWrapper._replaceNodeWithMarkup (app.js:32313)
    at ReactCompositeComponentWrapper._updateRenderedComponent (app.js:32303)
    at ReactCompositeComponentWrapper._performComponentUpdate (app.js:32253)
    at ReactCompositeComponentWrapper.updateComponent (app.js:32174)
    at ReactCompositeComponentWrapper.receiveComponent (app.js:32076)
    at Object.receiveComponent (app.js:24245)
    at Object.updateChildren (app.js:31347)
```

Would love to see a PR addressing this.

PR #1122, opened 2017-04-06
Ensures that after an undo is done, the next selected block has focus

mshibl:
This addresses issue #1120 

This issue could be reproduced by adding any custom component and then doing an **undo**.  This will cause a glitch in the cursor position and will affect all text written after that. 

This issue occurs because if the last added element is a custom block, and then the user clicks undo. The next block doesn't automatically get focus. This commit ensures that the new block after the undo gets focus.

_Issue reproduced on facebook notes_:
Issue

### DIFF
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -440,7 +440,7 @@ class EditorState {
       inlineStyleOverride: null,
       lastChangeType: 'undo',
       nativelyRenderedContent: null,
-      selection: currentContent.getSelectionBefore(),
+      selection: currentContent.getSelectionBefore().set('hasFocus', true),
     });
   }
 


src/model/immutable/EditorState.js
```diff
@@ -440,7 +440,7 @@ class EditorState {
       inlineStyleOverride: null,
       lastChangeType: 'undo',
       nativelyRenderedContent: null,
-      selection: currentContent.getSelectionBefore(),
+      selection: currentContent.getSelectionBefore().set('hasFocus', true),
```
octatone: Should [EditorState.forceSelection](https://draftjs.org/docs/api-reference-editor-state.html#forceselection) be used here instead?

PR #1126, opened 2017-04-07
Update lint rules, flow version, and inline documentation

mmmoussa:
**Summary**

This pr is cleanup to improve code quality. It includes the following improvements:
- updated the lint rule for max-len to better match the Fb configuration (addresses part of #862)
- updated the flow version
- updated some inline documentation

**Test Plan**

Made sure lint, flow, and test all pass with no errors.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,5 +5,6 @@ node_modules/
 website/
 gulpfile.js
 scripts/module-map.js
+scripts/jest/preprocessor.js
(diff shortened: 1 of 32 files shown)


mmmoussa:
While everything seems to be configured correctly, running flow does not work as expected. Some of the flow errors addressed in this pr (identified internally) are not being flagged by flow when I left them unfixed. I'm still looking into this.

PR #1135, opened 2017-04-13
Clarify comment above `tryToRemoveBlockStyle`

kedromelon:
I found the comment above `RichUtils.tryToRemoveBlockStyle` a bit confusing while trying to understand the function's logic.

I would argue that the current comment's inclusion of information about where it is used is unnecessary, and makes it seem as if its interaction with those key commands was implemented inside the function itself.


kedromelon:
great -- thank you @flarnie!

PR #1139, opened 2017-04-14
Docs tweaks

spicyj:
- Fix reference to older contentState
- Change `href` to `url` for consistency with "link" example in case people copy-paste between both (which I did today)
- Fix indentation in README


docs/Advanced-Topics-Entities.md
```diff
@@ -67,11 +67,11 @@ const contentState = editorState.getCurrentContent();
 const contentStateWithEntity = contentState.createEntity(
   'LINK',
   'MUTABLE',
-  {href: 'http://www.zombo.com'}
+  {url: 'http://www.zombo.com'}
```
mmmoussa: You should also change line 89 to use `url` instead of `href` for the sake of consistency.

spicyj: done

PR #1141, opened 2017-04-14
Update Entities docs in Advanced Topics section

hnordt:
Fixed a wrong sentence in **Creating and Retrieving Entities**.

Issue #1142, opened 2017-04-14
ContentState.createEntity Documentation Error


hnordt:
Hello.

In `ContentState.createEntity` docs we have:

![image](https://cloud.githubusercontent.com/assets/1625399/25055714/e37d1932-213a-11e7-9f0d-e1ec0c7289d2.png)

It says "Returns new `ContentState` record **_updated to include the newly created DraftEntity record in it's EntityMap_**.".

Now in `ContentState.createEntity` source code, we have:

```typescript
  createEntity(
    type: DraftEntityType,
    mutability: DraftEntityMutability,
    data?: Object,
  ): ContentState {
    // TODO: update this when we fully remove DraftEntity
    DraftEntity.__create(
      type,
      mutability,
      data,
    );
    return this;
  }
```

Either the docs needs to be fixed or `ContentState.createEntity` should apply the created entity (atm it's just creating).

Actually `ContentState.createEntity` behavior is a bit strange, e.g.:

```js
      const contentState = editorState.getCurrentContent();
      const selectionState = editorState.getSelection();
      const contentStateWithEntity = contentState.createEntity(
        "LINK",
        "MUTABLE",
        { url }
      );
      const entityKey = contentStateWithEntity.getLastCreatedEntityKey();
      const contentStateWithLink = Modifier.applyEntity(
        // contentState doesn't has the created entity, but still applied the created
        // entity successfully
        // I think it happens because the entity is owned by the editorState currently
        // Reading the docs, I expected that only contentStateWithEntity
        // would be able to apply the created entity
        contentState,
        selectionState,
        entityKey
      );
      const editorStateWithLink = EditorState.push(
        editorState,
        contentStateWithLink,
        "apply-entity"
      );
```

mmmoussa:
An instance of `ContentState` can contain an entity in its entity map without the entity being applied to any text. When using a new entity you have to create it and then apply it. The docs are describing this correctly. `ContentState.createEntity` should not apply the created entity.

With regard to your example about the strange behaviour, the reason you can use `contentState` is because your comment that it doesn't have the created entity is incorrect. `contentState.createEntity` returns the same instance of the `ContentState`, which means that both `contentState` and `contentStateWithEntity` are the same.

hnordt:
@mmmoussa thank you, I understand now. The source of my misunderstanding was the "Returns new `ContentState`". I fixed it: https://github.com/facebook/draft-js/pull/1150

[...]

PR #1149, opened 2017-04-17
Fixed incorrect entity documentation

mmmoussa:
I made a mistake in merging in #1141. The `ContentState` returned by `ContentState.createEntity` is not a new `ContentState`. Instead of reverting #1141 I'm just opening a new PR with the small fix since it's documentation.

### DIFF
--- a/docs/Advanced-Topics-Entities.md
+++ b/docs/Advanced-Topics-Entities.md
@@ -56,7 +56,7 @@ add rich rendering to your editor based on entity metadata.
 ## Creating and Retrieving Entities
 
 Entities should be created using `contentState.createEntity`, which accepts the
-three properties above as arguments. This method returns a new `ContentState` record updated to include the newly created entity, then you can call `contentState.getLastCreatedEntityKey` to get the key of the newly created entity record.
+three properties above as arguments. This method returns a `ContentState` record updated to include the newly created entity, then you can call `contentState.getLastCreatedEntityKey` to get the key of the newly created entity record.
 
 This key is the value that should be used when applying entities to your
 content. For instance, the `Modifier` module contains an `applyEntity` method:


hnordt:
@mmmoussa I believe we should update `ContentState.createEntity` docs too, please check #1150

flarnie:
So I might be misunderstanding but I think that it does return a new instance of `ContentState` because `ContentState` is Immutable.

hnordt:
@flarnie please check my comment on https://github.com/facebook/draft-js/issues/1142#issuecomment-294568933.

Atm `ContentState.createEntity()` looks like a mutable API, because it creates a new entity and returns the same `ContentState` instance, instead of a new one. https://github.com/facebook/draft-js/blob/47c26cc778da654231df0808fcb9bb1d2052662d/src/model/immutable/ContentState.js#L136-L148

flarnie:
Thanks for the link and the context - that helps.

This change does make the docs more clear @mmmoussa - thanks for updating this.

This will work differently once we fully deprecate the old Entity API. But for now I don't think the docs should get into that, so this is good.

If you're interested in more context:
In v0.11.0 the `contentState.createEntity` method will actually return a new instance of contentState. Because:
`contentState.createEntity` will call `createEntityInContentState` - https://github.com/facebook/draft-js/blob/0.11.0-alpha/src/model/immutable/ContentState.js#L138-L144
Which calls `addEntityToContentState` https://github.com/facebook/draft-js/blob/47c26cc778da654231df0808fcb9bb1d2052662d/src/model/transaction/addEntityToContentState.js#L25-L28
which calls `set` on the Immutable set, returning a new instance. https://facebook.github.io/immutable-js/docs/#/Map/set

hnordt:
Thank you @flarnie.

PR #1150, opened 2017-04-17
Fixed ContentState.createEntity docs 

hnordt:
Related to #1149


flarnie:
This is true - in v0.10.* it's true that it is not a new instance of `ContentState` that gets returned. Thanks for updating. :)

hnordt:
Worth noting it needs to be reverted after v0.11. Context: https://github.com/facebook/draft-js/pull/1149#issuecomment-294659681

PR #1151, opened 2017-04-18
Prepare Entities docs for v0.11 changes

hnordt:
As explained by @flarnie on https://github.com/facebook/draft-js/pull/1149#issuecomment-294659681, `ContentState.createEntity` will return a new `contentState` on `v0.11` instead of the same instance.

To prevent additional migration work, I've updated Entities docs to behave as if `ContentState.createEntity` already returns a new instance instead of the same instance.


flarnie:
Nice! Thanks! We will merge this once v0.11.0 is released. I expect we will do a v0.10.1 release first, in the near future. Have run into some blockers on doing releases, but hope to do it soon.

BornaP:
@flarnie Will v0.10.1 include those fixes merged into master, regarding Android browsers? And when can we expect release to be up and running, this week maybe? (I'm talking about v0.10.1)

Issue #1128, opened 2017-04-10
handleBeforeInput does not support composition input


goddyZhao:
**Do you want to request a *feature* or report a *bug*?**
feature

We are working on a feature like automatically inserting space between english character and chinese character, but the `handleBeforeInput` is not triggered when user types chinese characters.

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**

**What is the expected behavior?**

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

Draft.js `v0.9.1`
Browser `Chrome 57`

Issue #1079, opened 2017-03-16
Reapiting user insert  android


Nikcrysis:
Using Firefox on Android say you type: Hello!
Comes out: Hello!Hello!Hello!

flarnie:
Merging this into an issue that tracks all Android specific Draft.js bugs - see https://github.com/facebook/draft-js/issues/1895

PR #1152, opened 2017-04-20
Modified composition mode event handlers to provide better support fo…

njjewers:
Most browsers follow the spec for `compositionend` in a reasonable way that conforms to the spec, that is, the `data` field of the `compositionend` event contains the result of the composition. Firefox on Android, however, fires four events in sequence if you select a suggestion for a partially-typed word: `compositionend` (`data` is the partial the user typed), `beforeinput` (`data` is the partial the user typed), `compositionend` (`data` is the remaining letters in the complete word), and `beforeinput` (`data` is the remaining letters in the complete word). As a workaround for nonconforming behaviour like this, this patch changes the handling of events so that a variable textInputData is built from the concatenation of the `data` members of the `beforeinput` events fired during composition. During resolution, if this string has been built, it is used as the definitive source for the characters that have been typed. Otherwise, the `data` member from the first `compositionend` event fired during composition is used as the definitive source for the characters that have been typed.

I've tested this on Firefox and Chrome under Android 5.x, Safari on the iOS simulator, Chrome and Firefox under iOS, and IE11 under Windows 10. On desktop, I've tested it with both English and Japanese keyboard layouts. I'm reasonably confident that this change will improve support for mobile browser IMEs without breaking compatibility for any other existing supported IME, but I'm very interested in hearing any input that you have wrt these changes.

I acknowledge that I have received permission from Zugata Inc, the company for which I developed these changes under company time, to submit this pull request.

### DIFF
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -41,11 +41,23 @@ const RESOLVE_DELAY = 20;
  */
 let resolved = false;
 let stillComposing = false;
-let textInputData = '';
+let beforeInputData = null;
+let compositionEndData = null;
 
 var DraftEditorCompositionHandler = {
-  onBeforeInput: function(editor: DraftEditor, e: SyntheticInputEvent): void {
-    textInputData = (textInputData || '') + e.data;
+  /**
+   * Some IMEs (Firefox Mobile, notably) fire multiple `beforeinput` events
+   * which include the text so far typed, in addition to (broken) 
+   * `compositionend` events. In these cases, we construct beforeInputData from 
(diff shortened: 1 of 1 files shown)


njjewers:
Specifically, these changes should fix the issues described in #1107, #1079, and possibly #1128, as all of these are likely issues stemming from the differing (and arguably, non-spec compliant) sets of events generated by various combinations of OS, browser, and IME. The included test cases simulate a few different input combinations I've noticed during testing, as well as cases listed in the comments of draftEditorCompositionHandler.

AdrianMachado:
@flarnie Review this pls

nmn:
@njjewers I tried to pull in your PR. But since it's been sitting here for so long it no longer works consistently. Any chance you could rebase it?

PR #1162, opened 2017-04-22
Mark editor element as notranslate

spicyj:
This should help a lot with "Failed to execute 'setStart' on 'Range'" exceptions we've logged in production and seen in bug reports.

Test Plan:
Type in editor, right-click in Chrome then "Translate to English", editing no longer breaks.


spicyj:
(accepted internally)

Issue #1160, opened 2017-04-22
Update `createClass` use in Draft website code


flarnie:
In React v15.5 a warning was added for the impending deprecation of the `createClass` syntax. We have several calls to the deprecated `createClass` in the website code, and it should be updated.

[Check out the notes about how to update the syntax](https://facebook.github.io/react/blog/#migrating-from-react.createclass), and then please update each place we call it:

https://github.com/facebook/draft-js/search?utf8=%E2%9C%93&q=createClass&type=

morrme:
I'm on it!

niveditc:
I'm going to claim this one as a first fix to figure out the process of contributing to open source :)

niveditc:
This is no longer needed after the Docusaurus migration in #1759

PR #1170, opened 2017-04-24
Migrated from using 'createClass' per issue 1160

morrme:
**Summary**
Fixes: #1160
Replaced usage of createClass with createReactClass.


morrme:
@flarnie Seems like one test on Travis failed, saying to "arc build"? Please advise.

mmmoussa:
I think you need to update the version of React being required in `package.json`.

morrme:
@mmmoussa I tried that and have gotten the same result.

jomasti:
`website/package.json` is missing the dependency.

morrme:
@jomasti thanks for the suggestion. it is still failing :(   it definitely seems related to the versions, so i need to see if react-dom has a more recent version now, i think

flarnie:
Thanks for working on this! I'll take a look in a bit and see if we can get CI passing.

flarnie:
I checked this out, rebased it onto master, and was able to build the website with no issue. I'm hoping that whatever problem was happening in CI is solved by rebasing.

flarnie:
We are thinking the issue is an incompatibility between the new React build system and the unmaintained `react-page-middleware` which we use for the Draft.js docs site. https://github.com/facebookarchive/react-page-middleware

Probably should migrate off of `react-page-middleware` but for now we'll just stick with an earlier version of React.

niveditc:
These files were deleted when we migrated the website to Docusaurus in #1759 so this is no longer needed.

PR #1179, opened 2017-04-26
Update APIReference-Editor.md

ambewas:
update API reference Editor to include onFocus and onBlur props


flarnie:
Thanks for doing this!

PR #1190, opened 2017-05-03
selection.extend is called on selection without ranges

sarmstrong:
**Summary**

In our application we have multiple draft-js instances. When users try to drag entities from one editor to the other, the selection state is temporarily lost and we end up with the following error:

*Failed to execute 'extend' on 'Selection': This Selection object doesn't have any Ranges.*

We were able to recreate the issue and the following PR fixes the issue by verifying that the selection has ranges.

**Test Plan**

This particular file does not appear to have any active spec files, but I did confirm that the solution works for both when there is a selection and when there is not an active selection.

### DIFF
--- a/src/component/selection/setDraftEditorSelection.js
+++ b/src/component/selection/setDraftEditorSelection.js
@@ -235,9 +235,7 @@ function addFocusToSelection(
       });
     }
     // Check for active selection before extending
-    if (selection.rangeCount !== 0) {
-      selection.extend(node, offset);
-    }
+    selection.extend(node, offset);
   } else {
     // IE doesn't support extend. This will mean no backward selection.
     // Extract the existing selection range and add focus to it.
@@ -268,6 +266,9 @@ function addPointToSelection(
   }
   range.setStart(node, offset);
   selection.addRange(range);
+  if (selection.rangeCount === 0) {
+    selection.addRange(range);
+  }
 }
 
 module.exports = setDraftEditorSelection;
--- a/src/component/selection/setDraftEditorSelection.js
+++ b/src/component/selection/setDraftEditorSelection.js
@@ -234,7 +234,10 @@ function addFocusToSelection(
         selectionState: JSON.stringify(selectionState.toJS()),
       });
     }
-    selection.extend(node, offset);
+
+    if (selection.rangeCount > 0) {
+      selection.extend(node, offset);
+    }
   } else {
     // IE doesn't support extend. This will mean no backward selection.
     // Extract the existing selection range and add focus to it.
--- a/src/component/selection/setDraftEditorSelection.js
+++ b/src/component/selection/setDraftEditorSelection.js
@@ -234,8 +234,8 @@ function addFocusToSelection(
         selectionState: JSON.stringify(selectionState.toJS()),
       });
     }
-
-    if (selection.rangeCount > 0) {
+    // Check for active selection before extending
+    if (selection.rangeCount !== 0) {
       selection.extend(node, offset);
     }
   } else {


ryanswanson:
Our application includes multiple draft-js instances as well that are encountering this exception upon startup of the application for all users after upgrading to Chrome 58, leaving React in a broken state (i.e. app is unresponsive/hangs).  We've verified that this pull request addresses the issue.

sarmstrong:
@flarnie Any change you could merge in either:

https://github.com/facebook/draft-js/pull/1190
https://github.com/facebook/draft-js/pull/1192
https://github.com/facebook/draft-js/pull/1203

They are all addressing the same issue. I don't care if it's mine you choose, as long as the bug is squashed!!!

Much appreciated!

flarnie:
This is one of many great fixes we want to pull into Draft - I'm working on https://github.com/facebook/draft-js/issues/1244 in order to unblock landing things on Github again, but @spicyj might want to look at those fixes. We have also been looking into this bug internally.

Hoping to get some of these things landed and released soon, thanks for your work on this!

sarmstrong:
Awesome, thanks @flarnie, much appreciated!

spicyj:
So, I have also seen this but am pretty unconvinced that the fix is correct. Based on my reading of the code, we always add a range immediately before calling .extend here – so .rangeCount should never be zero. One time when I was able to reproduce this, it also seemed like calling .addRange *again* (even though it had already been called) fixed the selection's state so that .rangeCount was then 1.

If you have a consistent repro of this, could you try checking if you also see that? I'm thinking when we call `addRange` it could be something like:

```
selection.addRange(range);
if (selection.rangeCount === 0) {
  selection.addRange(range);
}
```

even though that doesn't make too much sense. It sounds like a Chrome bug to me. We can add a workaround for browser bugs but I like to understand them first – and finding a way to fix the selection instead of silently failing (as your proposal does) seems better.

sarmstrong:
Thanks @spicyj, I was able to reproduce the issue and your suggestion worked!!! I updated the PR.

spicyj:
Would anyone here be interested in digging through to figure out why Chrome has this issue? If you're feeling adventurous it might make sense to build Chromium from source and add logging to the Selection.prototype.addRange method to understand why it fails the first time but works the second time.

flarnie:
@sarmstrong - Since this looks like it's a Chrome bug, we'd like to get a non-Draft repro case and file a bug with Chrome. Then we can add a comment next to this work-around that references the bug report.

With that addition we would be happy to merge this. Thanks for your patience and let me know if you have questions. Thanks! 🙏

spicyj:
To be clear: it would be really helpful if anyone here could come up with a minimal repro of this issue without using Draft that demonstrates the issue. I'm not sure on the exact circumstances that make the bug reproduce, but if someone has time to try making something that would be helpful. I'd try:

* make a page with one button and two contenteditable divs
* when the button is clicked, set the selection into the first contenteditable div (using the same addRange+extend strategy as here), then set the selection into the second
* try to figure out how to make it throw?

If anyone here has time to look into it that would help us out a lot. Otherwise, we'll hopefully have time to try ourselves sometime in the next few weeks.

flarnie:
Just an update - none of the core Draft.js folks have time this week to debug the Chrome bug causing this, and we'd still like to get more info there and file a bug with Chrome before adding this work-around to Draft.js. This PR is still on our radar, and we're happy to help with reviews or discussion if someone can continue looking into this.

sarmstrong:
From my understanding you still need a consistent reproduction? I can try and carve out some time this week and see if I can throw something together, would that help getting it out the door?

sarmstrong:
Sorry, one of my colleagues just informed me that I misread the request, I think I might be able to reproduce within the context of draft, but I'm not quite sure where to start w/o it and probably won't be able to find the carve out.

sarmstrong:
Sorry accidentally closed this, after our scrum planning for the week, it doesn't look like I'll be able to get to this, alternately, we were able to get ourselves unblocked for the time being by refactoring some of the surrounding code so the request has become less urgent.

flarnie:
Just an update - we are starting to investigate filing a Chrome bug, and so if anyone picks this up please post here. ccing @paopaoF who may be picking this up

spicyj:
The most obvious way to maybe repro doesn't throw for me:

```
  <button id="button">boom</button>
  <div contenteditable id="ce_a">first contenteditable</div>
  <div contenteditable id="ce_b">second contenteditable</div>
  <script>
    function setSelection(node) {
      var selection = window.getSelection();
      selection.removeAllRanges();
      var range = document.createRange();
      range.setStart(node.firstChild, /* offset */ 0);
      selection.addRange(range);
      selection.extend(node.firstChild, 1);
    }
    button.onclick = function() {
      setSelection(ce_a);
      setSelection(ce_b);
    };
  </script>
```

kgoggin:
I've been lurking on this issue because I had experienced this error myself around the time it was first reported. But, I just noticed that it's not popping up for me any longer. There's a chance its due to a code change on my end, but I'm wondering if Chrome may have fixed this already, especially given @spicyj's inability to repro just now?

spicyj:
@kgoggin Unfortunately I can still repro in Chrome 62.0.3175.0 canary with

```
function getFreshEditorState(text) {
  return EditorState.createWithContent(ContentState.createFromText(text));
}

class Demo extends React.Component {
  state = {
    editorStateFoo: getFreshEditorState('Foo'),
    editorStateBar: getFreshEditorState('Bar'),
  };
  
  handleClick = () => {
    this.setState(state => ({
      ...state,
      editorStateFoo: EditorState.moveFocusToEnd(getFreshEditorState('Foo')),
      editorStateBar: EditorState.moveFocusToEnd(getFreshEditorState('Bar')),
    }));
  }
  
  render() {
    return (
      <div>
        <DraftEditor editorState={this.state.editorStateFoo} />
        <DraftEditor editorState={this.state.editorStateBar} />
        <button onClick={this.handleClick}>Ruin everything</button>
      </div>
    );
  }
}

ReactDOM.render(
  <Demo />,
  container
);
```

and clicking the button.

spicyj:
THIS IS RELEVANT https://github.com/facebook/draft-js/blob/744e9b4eb4810797a93c66591fea6f9cac533b4b/src/component/handlers/edit/editOnBlur.js#L25-L33

That happens (synchronously) when we add the range causing the second contenteditable to be focused. Which is why there are no ranges after we call it.

This might not actually be a Chrome bug!

spicyj:
Changing editOnBlur to read:

```
  if (isWebKit && getActiveElement() === document.body) {
    const selection = global.getSelection();
    const editorNode = editor.refs.editor;
    if (
      selection.rangeCount === 1 &&
      containsNode(editorNode, selection.anchorNode) &&
      containsNode(editorNode, selection.focusNode)
    ) {
      selection.removeAllRanges();
    }
  }
```

fixes this I think. Will verify at Facebook and commit a permanent fix if so.

martinbroos:
Is it possible to add this fix to 0.10 release as well? I'm experiencing an indexSize Error on selection.extend in IE11 when applying a entity on the editor state.

```
    // the reference needs to be added after the current selection, so we first need to create a new selection
    var newSelectionThatStartsAfterTheCurrentSelection = currentSelection.merge({
        anchorOffset: currentSelection.focusOffset,
        focusOffset: currentSelection.focusOffset
    });

    // insert empty text as a placeholder at that selection, to contain the reference icon that is displayed in the editor
    // after the word the user selected for the reference
    var newContentStateWithExtraSpaceAndReference = Modifier.insertText(
        newContentStateWithEntity,
        newSelectionThatStartsAfterTheCurrentSelection,
        ' ',
        undefined,
        lastCreatedEntityKey
    );

    return EditorState.push(newEditorStateWithEntity, newContentStateWithExtraSpaceAndReference, 'apply-entity');
```

Or does anyone know a workaround? I tried using forceSelection, this works in IE but causes problems later on when getting the active selection.

lostpebble:
Has this fix been implemented yet in an available version? This is causing major headaches for people using an editor I built which has been functioning just fine for the past year. Basically can't do anything from within the editor that leads to focus being removed from the editor without an error being thrown. Buttons within the editor for initiating various types of edits are all throwing this error, and some url links which contain hash links within the page are also causing problems.

I've also moved on to React v16 now, and the issue is persisting.

sophiebits:
@flarnie is hoping to do a 0.10.1 release soon which will include this fix.

flarnie:
It will be 0.10.2, and thanks @sophiebits - I'm waiting for https://github.com/facebook/draft-js/pull/1378 to land.

michelson:
hello, this fix is not released yet ?

flarnie:
Hi @michelson - 
Based on the logs and reports I have access to, this bug was fixed in v0.10.2 of Draft.js.

See the changelog for 0.10.2 - https://github.com/facebook/draft-js/releases/tag/v0.10.2

> Fix pesky "This Selection object doesn't have any Ranges" bug (@sophiebits in 96688e10 and 036e3a84)

The fix was added by @sophiebits in https://github.com/facebook/draft-js/commit/036e3a848e3593c97f4c3011e1ddc045e128a7af under a feature flag, and the flag was opened/removed in https://github.com/facebook/draft-js/commit/96688e10b22a778c76e03009da4b9f3d05eba5f7

If you are still seeing this error in Draft v0.10.2 then please open a new issue and provide a repro case plus any other info you have.
Thanks!

michelson:
HI @flarnie, thanks for your reply. I think I've hitting a related issue then. 
not sure if it's my implementation. I've a custom component which has an input text field to edit the data from the component itself. The problem arises when I type outside the component, in other words, when I type over the editor text. For some reason when I type back (blur editor) in the input I get a chrome warning: `[Deprecation] The behavior that Selection.addRange() merges existing Range and the specified Range was removed. See https://www.chromestatus.com/features/6680566019653632 for more details.` and the typing on the input field gets clumsy

how could I clean the ranges properly in order to make this work in Chrome?
in firefox works fine

michelson:
you can see video example at http://recordit.co/9690tkLIzh

shpakkdv:
Similar issue still exists. 
Faced with this in Safari (only) when Draft.js is updating the editor state.
The error which I see: `IndexSizeError: The index is not in the allowed range.`

My solution / hack:
```js
  const nativeSelectionExtend = Selection.prototype.extend;

  Selection.prototype.extend = function (...args) {
    try {
      return nativeSelectionExtend.apply(this, args);
    } catch (error) {
      console.log('Selection error.', error);
    }
  };
```

It works properly for me. Maybe will be useful for somebody as well.

lvdang:
> Similar issue still exists.
> Faced with this in Safari (only) when Draft.js is updating the editor state.
> The error which I see: `IndexSizeError: The index is not in the allowed range.`
> 
> My solution / hack:
> 
> ```js
>   const nativeSelectionExtend = Selection.prototype.extend;
> 
>   Selection.prototype.extend = function (...args) {
>     try {
>       return nativeSelectionExtend.apply(this, args);
>     } catch (error) {
>       console.log('Selection error.', error);
>     }
>   };
> ```
> 
> It works properly for me. Maybe will be useful for somebody as well.

Which file did you add this code?

shpakkdv:
@lvdang 
Actually, in any file. But better to add in any root file, which will be initialized only once to avoid incorrect behavior during hot reload.
Or to make sure you can add additional check:
```js
if (Selection.prototype.extend.toString().includes('[native code]')) {
  const nativeSelectionExtend = Selection.prototype.extend;

  Selection.prototype.extend = function (...args) {
    try {
      return nativeSelectionExtend.apply(this, args);
    } catch (error) {
      console.log('Selection error.', error);
    }
  };
}
```

lvdang:
@shpakkdv  - thanks

Issue #1193, opened 2017-05-08
Is this project fading away?


crowdwave:
I'm looking for a rich editor for reactjs but with 350 open issues and seemingly a trickle of commits I'm wondering if this project is still one to choose if I want one that is very actively developed and supported.

If not, can anyone suggest a more actively developed rich text solution for reactjs?

robbertbrak:
That's a fair concern and one that I'm worried about myself. It seems that the original author is not around anymore and that there is very limited activity from the people at Facebook that have since taken over. Just take a look at the (lack of) 'weekly' meeting notes to get an idea: https://github.com/facebook/draft-js/tree/master/meta/meeting-notes.

If you are looking for a more actively developed Reactjs project, you could take a look at http://slatejs.org/. It has much better support for nested structures such as tables, so it may be more suitable for rich text editors that need complex functionality. Be aware, though, that Slatejs is not tested against IE/Edge and mobile devices, according to its FAQ (https://docs.slatejs.org/general/faq.html). Also, IME support seems terrible.

flarnie:
This is absolutely a fair concern. There are still folks at Facebook actively maintaining this library, including myself, and we appreciate your patience since we also work on other open source projects. Draft is one of many options for a rich text editor, and it's worth looking at all the options and choosing what works best in your case.

Regarding the meeting notes - we didn't have much to report for the last few weeks so I have not been updating them on github, but in the future I will add some record of each meeting.

I'm about to do a minor release of the latest bug fixes and improvements - hopefully that is helpful to folks.

robbertbrak:
Thanks for the reply, @flarnie! It's good to see in the meeting notes that you're also looking for ways to involve more maintainers both from Facebook and externally.

crowdwave:
Of course no-one is obliged to work on this because its open source.

I ask because the community seems to say this is the flagship rich text editor for reactjs but my measures of a healthy active open source project - a low number of open issues plus a fairly constant flow of commits - suggest that this project is largely (but not entirely) abandoned.

ironosity:
Over a year later, 480 open issues and 80 open PRs.  It's frustrating for my team because many of our pain points are open issues AND have open PRs...would be great to see Facebook put more resources at this project so at the very least, community PRs can be merged and more releases.

flarnie:
Thanks for your comments - I set up a meeting with @sophiebits to come up with an update on the state of this project. You can expect that in the next few weeks.

[...]

PR #1195, opened 2017-05-09
Update weekly meeting notes from March-April 2017

flarnie:
**what is the change?:**
We had few things to note each week and were deferring updating the
meeting notes.

**why make this change?:**
Based on [issue #1193](https://github.com/facebook/draft-js/issues/1193)
it sounds like the meeting notes will provide signal that the project is
still maintained.


spicyj:
nice, thanks

Issue #1201, opened 2017-05-10
Mention including Draft CSS in 'getting started'


flarnie:
Since this was not clear, (see https://github.com/facebook/draft-js/issues/1197), I think we could update the docs to mention this in the 'Quick Start', 'Overview' section. https://draftjs.org/docs/overview.html#content

jackyho112:
@flarnie Hello, I have given this one a go. Please let me know if there is any improvement required. We use `draft.js` a lot. I look forward to doing more contribution and `0.11.0`!

PR #1208, opened 2017-05-14
Add mention of draft css in overview doc

jackyho112:
**Summary**

Mention the need to include `Draft.css` when rendering the editor in the overview doc and reference the link, https://draftjs.org/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css, to learn more about why it is needed. 

This is related to https://github.com/facebook/draft-js/issues/1201.

**Test Plan**

This is a doc update. There is probably no need for additional tests.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -75,4 +75,10 @@ Because Draft.js supports unicode, you must have the following meta tag in the `
 <meta charset="utf-8" />
 ```
 
+`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css).
+
+```js
+import from 'draft-js/dist/Draft.css';
+```
+
 Next, let's go into the basics of the API and learn what else you can do with Draft.js.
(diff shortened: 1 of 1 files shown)


docs/Overview.md
```diff
@@ -75,4 +75,10 @@ Because Draft.js supports unicode, you must have the following meta tag in the `
 <meta charset="utf-8" />
 ```
 
+`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css).
+
+```js
```
flarnie: There are many ways to include CSS, and some folks will use `require` vs. `import` or might even compile their CSS and include the bundle in a `link` tag in their HTML header. What do you think of leaving out the sample of how to import it, since that part can vary so much?

jackyho112: Yeah, I think your suggestion makes sense. What I'm doing here could be giving the impression that this is the standard way to include the css file, which, as you pointed out, is not true. I will just remove the code example.

jackyho112:
@flarnie I made the change. Please let me know if there is any more improvement you would like on this branch. Thanks!

flarnie:
Thanks!!! :)

lesreaper:
Yes, how about directions about how to actually include the CSS? There doesn't seem to be an explanation about how to do so properly.

sjelfull:
@jackyho112 I realise this is an old pull request, but as @lesreaper mentions, it does not actually tell you how to do it.

Issue #1244, opened 2017-06-11
Set up automated syncing between Github and Facebook Draft.js versions


flarnie:
Right now changes are manually copied between the Facebook and Github copies of Draft.

Automating this process will save time, freeing me and other maintainers up to do more issue/PR triage and merge things faster.

flarnie:
Looks like this is working! Making a list now of PRs that could be merged or are close to ready for merge.

PR #1247, opened 2017-06-15
Tweak syntax of accessing '_instance' in DraftEditor test

flarnie:
**what is the change?:**
We either check for the 'getEditorKey' method on the
'renderedComponent._instance' or the
'renderedComponent._instance._instance'

**why make this change?:**
Internally we are using a different variation of the 'shallowRenderer'
and this test needs to run internally at Facebook as well as on Github.

This is not an ideal solution, but we can continue improving this test
once cleaning up the inconsistencies between Facebook and Github.

**test plan:**
`jest src/component/base/__tests__/DraftEditor.react-test.js`

Since I have limited time to work on this I will probably merge if/when CI passes.

**issue:**
https://github.com/facebook/draft-js/issues/1244

### DIFF
--- a/src/component/base/__tests__/DraftEditor.react-test.js
+++ b/src/component/base/__tests__/DraftEditor.react-test.js
@@ -34,8 +34,15 @@ describe('DraftEditor.react', () => {
       shallow.render(
         <DraftEditor />,
       );
-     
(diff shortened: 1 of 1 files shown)


spicyj:
Does .getRenderedInstance() work? This shouldn't be using internals anyway.

PR #1254, opened 2017-06-19
Allow Option+Space to be handled on OSX Chrome

colinjeanne:
**Summary**

Currently pressing Option+Space in Chrome on OSX will cause a nonbreaking space to always be inserted into the editor. This behavior ignores any key bindings that the editor has set.

This change moves the logic which adds the non breaking space to after the point that the key binding function has been run in order to give editors a chance to handle this behavior on their own.

**Test Plan**

This has been tested manually since there are no current tests for the event handlers and no current framework for testing the handlers. We have been running [with this change in our fork of Draft](https://github.com/textioHQ/draft-js/commit/f4c3aebe6593e3ccdb1de4c15363e72faae0f236) for about seven months without issue.

### DIFF
--- a/src/component/handlers/edit/editOnKeyDown.js
+++ b/src/component/handlers/edit/editOnKeyDown.js
@@ -148,19 +148,9 @@ function editOnKeyDown(editor: DraftEditor, e: SyntheticKeyboardEvent<>): void {
       }
       break;
     case Keys.SPACE:
-      // Handling for OSX where option + space scrolls.
+      // Prevent Chrome on OSX behavior where option + space scrolls.
       if (isChrome && isOptionKeyCommand(e)) {
         e.preventDefault();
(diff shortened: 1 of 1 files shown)


niveditc:
@colinjeanne - apologies for the delay in reviewing this.

Regarding the bundle size files, I think you can ignore them & remove them from the commit. Could you do that and rebase on master? (side note: I saw your PM on Slack and am looking into the possibility of adding this to `.gitignore` and/or automatically generating those files)

niveditc:
Next steps are:
* Provided the steps above are successful I'll merge it in. I'm limiting riskier merges to Draft.js to working hours & staggering them by a couple of hours each to be able to quickly bisect & revert in the slim chance that something breaks. So I estimate that this one will be merged towards EOD today or sometime tomorrow.

PR #1261, opened 2017-06-26
Support moving atomics

tedmx:
**Summary**

While testing `Editor`, my team has discovered that drag'n'drop doesn't work well with atomics (blocks of type `atomic`) selected. Delving deeper into the issue, it came out that core `removeRangeFromContentState` and `insertFragmentIntoContentState` functions don't fully support atomics.

`removeRangeFromContentState`, if given a fragment with a first block as atomic, only cuts text from that atomic without removing it, which effectively leaves the block unchanged. 

`insertFragmentIntoContentState`, if given a fragment with a first block as atomic, takes that atomic and merges its text (whitespace character) to a target block, leaving target's block type as it was (for example, `unstyled`). That effectively does not insert the atomic block.

This solution is generally to remove atomics in first case and to append them, not merge, in second case.

**Test Plan**

```
src/model/transaction/__tests__/removeRangeFromContentState-test.js
src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
```

### DIFF
--- a/src/model/modifier/AtomicBlockUtils.js
+++ b/src/model/modifier/AtomicBlockUtils.js
@@ -13,6 +13,8 @@
 
 'use strict';
 
+const invariant = require('invariant');
+
 const BlockMapBuilder = require('BlockMapBuilder');
 const CharacterMetadata = require('CharacterMetadata');
 const ContentBlock = require('ContentBlock');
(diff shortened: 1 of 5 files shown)


tedmx:
Bug found; reverting.

tedmx:
Bug fixed. It was a duplication of `unstyled` text range on drag'n'drop.

delijah:
What about using AtomicBlockUtils.moveAtomicBlock()?

niveditc:
Hey @tedmx - apologies for the late review! Could you help us understand why `AtomicBlockUtils.moveAtomicBlock` doesn't work for this?

PR #1266, opened 2017-06-28
Allow DraftJS to work in iframes

bloodyowl:
**Summary**

This allows DraftJS to render within an iframe that is controlled by a script in an ancestor window.

The `instanceof` checks on `HTMLElement`, `Element` and friends triggered an invariant error because they don't necessarily inherit from the same one that is available as a global variable to the `draft-js` script.

**Test Plan**

Tests currently fail, and it seems to fail because of a jsdom or jest bug.


charlieclark:
Trying to use DraftJS in an iframe and getting the same errors. Would love to see this merged in.

Issue #1036, opened 2017-02-27
Updating docs when releasing v0.11.0


flarnie:
This issue is for tracking updates that we will need to make and/or merge when releasing v0.11.0, because that version will include API changes and new features. For starters:

 - Update all docs *and* examples to only include v0.11.0 APIs for entity management. See http://draftjs.org/docs/v0-10-api-migration.html#content for context
 - Add docs on new props: `[textDirectionality](https://github.com/facebook/draft-js/pull/1034)`, [`autoCapitalize`, `autoComplete`, and `autoCorrect`](https://github.com/facebook/draft-js/pull/1035). That assumes those PRs are merged. 😄 

We can comment on this issue with any other new features or API changes that get added to v0.11.0

juliankrispel:
@flarnie I opened a pr for adding docs for autoCapitalize, autoComplete and autoCorrect, as far as I can see entity docs are updated. Could you tell me what else you had in mind in terms of docs?

Shoukld there be a v0.11 migration guide?

flarnie:
Hi again @juliankrispel - thanks for working on this! I forgot that we did update the Entity docs when releasing `v0.10.0`, so all that is left is to have a PR ready that moves the contents of [`examples/draft-0-10-0`](https://github.com/facebook/draft-js/tree/master/examples/draft-0-10-0) into `examples/` and deletes the [`examples/draft-0-9-1`](https://github.com/facebook/draft-js/tree/master/examples/draft-0-9-1) directory. It would also need to update links to examples in the docs.

That would get merged when we do the `v0.11.0` release. I expect to post more info about the `v0.11.0` release soon.

juliankrispel:
@flarnie I made those [changes here](https://github.com/facebook/draft-js/pull/1304) lmk if all looks good and what else I can do to move the release along! Would be happy to help in any way!

PR #1269, opened 2017-06-29
docs for autoComplete, autoCorrect and autoCapitalize

juliankrispel:
**Summary**

Documentation for `autoComplete`, `autoCorrect` and `autoCapitalize`

Addresses #1036


docs/APIReference-Editor.md
```diff
@@ -106,6 +106,31 @@ for details on usage.
 
 ### Behavior (Optional)
 
+### autoCapitalize?: string
+
+```
+autoCapitalize?: string
+```
+
+Set if auto capitalization is turned on and how it behaves. More about platform availability and usage can [be found on mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#attr-autocapitalize).
```
flarnie: I like the linking to MDN docs!

The part "and how it behaves" doesn't totally make sense to me in these three. Could it be something similar to the `spellcheck` description?; "Set whether spellcheck is turned on for your editor."

It would also be nice to mention the default value, like we do with `spellcheck` and `stripPastedStyles`.

juliankrispel: @flarnie the thing is - with `spellCheck` it is just a boolean so you can set it on or off.

But with `autocapitalize` you have several options:

`none`
`sentences`
`words`
`characters`
`on`
`off`

So just saying you can turn it off and on doesn't cut it for me. I would include the options you can give it, but I'm afraid that the docs would go out of date fairly quickly since it's such a new nonstandard feature...

flarnie: I stand 'autocorrected' ha ha :)

You're right, since these all take strings and not booleans, the wording you had makes sense. And linking to the MDN docs will hopefully keep this up to date with no changes. Nice!

flarnie:
We ended up releasing these three props in v0.10.1, so I'll add this to my queue for merging.

Thanks again for your help with the docs!

thanks_dog

PR #1270, opened 2017-06-29
use `event.target` in FileReader load handler

bjohn465:
**Summary**

According to the spec (https://w3c.github.io/FileAPI/#events):

> The FileReader object must be the event target for all events in this specification.

Using `event.target` also avoids a Flow error when using Flow 0.49.0 or greater, since 0.49.0 changed FileReader's `result` property to be `string | ArrayBuffer`.

**Test Plan**

* Open `examples/draft-0-10-0/rich/rich.html` and drag and drop one or more text files from your operating system's file browser into the editor
* Contents of dropped file(s) should be added to the editor

### DIFF
--- a/src/component/utils/getTextContentFromFiles.js
+++ b/src/component/utils/getTextContentFromFiles.js
@@ -69,8 +69,8 @@ function readFile(
   }
 
   var reader = new FileReader();
-  reader.onload = function() {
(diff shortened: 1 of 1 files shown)


niveditc:
Looks like this was fixed in #1286 & please reopen if this is still a necessary fix.

PR #1277, opened 2017-07-06
Add support for mark (highlight) inline tag when converting from HTML.

deep-c:
**Summary**
Support for mark (highlight) HTML inline tag style when converting from HTML.


deep-c:
Would it be a better idea to allow for users to pass in a custom inlineTags object which gets merged with the current inlineTags object to convertFromHTMLtoContentBlocks so that we could override or customise this for our particular needs?

niveditc:
This was added in https://github.com/facebook/draft-js/commit/37f2f2ac16e4edf29c7c7e8230dec544faeafdfe

PR #1285, opened 2017-07-12
Added support for the `compositionupdate` event

mathieumg:
**Summary**

@flarnie As we discussed, this adds support for the [`compositionupdate`](https://developer.mozilla.org/en-US/docs/Web/Events/compositionupdate) event and ignores the `beforeinput` event if it were to be fired in that scenario too.  I imagine a slightly cleaner way of doing this would be to remove the listener on `beforeinput` altogether when we realize `compositionupdate` fires. Should I squash my changes?

**Test Plan**

I didn't add new tests, should I add tests that fail without this fix?

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -100,6 +100,7 @@ class DraftEditor extends React.Component {
   _onCharacterData: Function;
   _onCompositionEnd: Function;
   _onCompositionStart: Function;
+  _onCompositionUpdate: Function;
   _onCopy: Function;
   _onCut: Function;
   _onDragEnd: Function;
@@ -144,6 +145,7 @@ class DraftEditor extends React.Component {
     this._onCharacterData = this._buildHandler('onCharacterData');
     this._onCompositionEnd = this._buildHandler('onCompositionEnd');
     this._onCompositionStart = this._buildHandler('onCompositionStart');
+    this._onCompositionUpdate = this._buildHandler('onCompositionUpdate');
     this._onCopy = this._buildHandler('onCopy');
     this._onCut = this._buildHandler('onCut');
     this._onDragEnd = this._buildHandler('onDragEnd');
@@ -263,6 +265,7 @@ class DraftEditor extends React.Component {
             onBlur={this._onBlur}
             onCompositionEnd={this._onCompositionEnd}
             onCompositionStart={this._onCompositionStart}
+            onCompositionUpdate={this._onCompositionUpdate}
             onCopy={this._onCopy}
             onCut={this._onCut}
             onDragEnd={this._onDragEnd}
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -68,7 +68,10 @@ var DraftEditorCompositionHandler = {
    * A `compositionupdate` event has fired. Update the current composition
    * session.
    */
-  onCompositionUpdate: function(editor: DraftEditor, e: SyntheticInputEvent): void {
+  onCompositionUpdate: function(
+    editor: DraftEditor,
+    e: SyntheticInputEvent,
+  ): void {
     compositionUpdates = true;
     textInputData = e.data;
   },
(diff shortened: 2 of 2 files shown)


flarnie:
Thanks for making this into a PR! 

> Should I squash my changes?

Nah, we always 'Squash and merge' when merging, so you can leave the commits separate if that's easier for you.

mathieumg:
> we always 'Squash and merge' when merging

I totally forgot GitHub added that as a feature a while ago, all good! 👍

spicyj:
Also, one other thing: onBeforeInput is not actually the browser's native event. We polyfill it in many cases inside of React here: https://github.com/facebook/react/blob/3b5e3b5a23069dd03feaa6979bccb64de4d76d09/src/renderers/dom/shared/eventPlugins/BeforeInputEventPlugin.js. It might make sense to switch that logic so that onBeforeInput events are properly synthesized based on the composition events that Android provides, and then we don't need to change Draft proper at all.

mathieumg:
@flarnie Thanks for the return! I just got home right now, will try to update the PR ASAP!

flarnie:
Hi again - 
I checked this out and ran into a couple of issues. I'd still like to see this fix or a similar fix added, but I'm not sure how to make it work. Or we may go in the direction of adding this to the existing 'onBeforeInput' polyfill in React. I'm not sure.

*Issue #1*
To make this a smaller change to the current behavior, we suggested defaulting to 'onBeforeInput' instead of 'onCompositionUpdate'. But because 'onCompositionUpdate' fires first, it's going to be more complex to fall back to it. Just seems like it will add complexity and maybe there is a different way.

*Issue #2*
This doesn't quite make sense with the pattern of events fired for Korean input, and other IME use cases.

I manually tested Korean, and also recorded the input events using https://dvcs.w3.org/hg/d4e/raw-file/tip/key-event-test.html in Chrome on mac. You can see that the data doesn't contain the whole composition, and so it ends up clearing the composition at the end and replacing it with the last character:

bug_with_korean_input

This is inconsistent with Japanese I think, where the data on the 'onCompositionUpdate' contains the entire composition so far. Japanese seemed to work ok with this fix, but I still would be concerned about subtle changes there.

I'm attaching pdfs of the keypress events for Korean and Japanese examples, for reference. Basically I think someone (maybe me eventually) needs to catalogue all the event patterns for different IME and composition use cases, and come up with handlers or modes for the different cases.

If you don't need to support Korean or other IME input then this fix may still be good enough for your use case. If you're open to addressing the issues with Korean input I can help look into it, or we can close this pending further investigation.

octatone:
@flarnie This PR, from Apr 20, seems to tackle a number of IME/OS composition event combinations in a simplified manner here: https://github.com/facebook/draft-js/pull/1152

What are your thoughts on this solution?

mathieumg:
@octatone I'm pretty sure I gave a shot to that PR first, but it also ignores the `compositionupdate` event, which is the only thing working on some browsers/devices.

@flarnie I've been meaning to get to this but have been rather busy as of late, will try my best to find some time soon! 😅

octatone:
@mathieumg @njjewers @flarnie 

I merged and resolved conflicts with the work done so far by both @njjewers and @mathieumg which hopefully fixes a bunch of issues with IME and Android keyboard weirdness.

You can see my result here: https://github.com/facebook/draft-js/compare/master...octatone:raymond/input-bugs

I manually tested Korean, Japanese and Chinese (simplified) with these changes and added unit tests to ensure that further improvements don't break expect normal IME cases. My unit tests are based on captured IME events on Chrome + Windows.

Please review and let me know if combining these two bodies of work into one is okay., and I will submit a proper PR with feature flags, etc.

flarnie:
@octatone - this looks interesting, thanks for working on it!

I'm working on some React stuff at the moment but will review your result as soon as I can.

tomkelsey:
@octatone Thanks to you (and everyone else) for your work on this. It's been driving our Android users a bit nutty!

I've just tested out your changes with my own Android device and it looks like it clears things up with GBoard but still odd behaviour with SwiftKey - typing any word causes the keyboard to immediately lose focus. But, either way, significantly better than it was.

matthewshirley:
@tomkelsey Same problem with SwiftKey. If I enter @ or $ before typing it works as intended though. Same for you?

octatone:
I have opened a PR that continues from this work https://github.com/facebook/draft-js/pull/1419

PR #1295, opened 2017-07-19
Add aria-multiline attr to DraftEditor

jessebeach:
`aria-multiline` is used to indicate to assistive technology agents that a textbox is a multiline textbox. When `aria-multiline` is true, an AT should treat return key presses as carriage returns, not submit actions.

https://www.w3.org/TR/wai-aria-1.1/#aria-multiline

teeeeeeeested: `npm run test && npm run lint && npm run flow`

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -245,6 +245,7 @@ class DraftEditor extends React.Component {
             aria-expanded={readOnly ? null : this.props.ariaExpanded}
             aria-haspopup={readOnly ? null : this.props.ariaHasPopup}
             aria-label={this.props.ariaLabel}
+            aria-multiline={this.props.ariaMultiline}
             aria-owns={readOnly ? null : this.props.ariaOwneeID}
             autoCapitalize={this.props.autoCapitalize}
             autoComplete={this.props.autoComplete}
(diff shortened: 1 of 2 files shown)


spicyj:
Cool! Do screen readers support this property yet?

jessebeach:
@spicyj no, not yet. @mcking65 is working with them as part of the ARIA 1.1 specification process. 

https://codepen.io/jessebeach/pen/gRVKeY

flarnie:
I'm excited to land this, the build keeps failing. Going to try a couple more times, I'm not getting much info on why it failed.

jessebeach:
💯 💯 💯 💯 💯 💯

flarnie:
@jessebeach can you try making this an internal diff and then we can export it to github? I don't know why the build won't pass when I import it from github.

cc @zpao

jessebeach:
Will do.

spicyj:
@flarnie See the error here if you click through the internal build logs: https://fburl.com/hhnvslws. I think maybe removing the "tested:" section would help?

jessebeach:
It might be failing because a Jest snapshot needed to be updated.

flarnie:
Thanks @spicyj - that was the issue.

Issue #1211, opened 2017-05-15
'blockRenderMap' may cause unhandled exception when 'blockRenderFn' is provided


southerncross:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

I want to create a block type to center text, so I defined a new block type 'CENTER_BLOCK_TYPE' and a new `blockRenderFn`.

But I got an error: `Uncaught TypeError: Cannot read property 'wrapper' of undefined.`
(Draft 0.10.1)

![image](https://cloud.githubusercontent.com/assets/3861465/26061644/140f7f0c-39bb-11e7-8ba4-e5295a0a9760.png)

The online example is: https://jsfiddle.net/m6z0xn4r/191/

After debugging for a while, I noticed that it was the `blockRenderMap` that caused the exception.

```javascript
const configForType = blockRenderMap.get(blockType); // 'configForType' will be undefined
const wrapperTemplate = configForType.wrapper; // can not read wrapper of undefined
```
(https://github.com/facebook/draft-js/blob/master/src/component/contents/DraftEditorContents.react.js#L147)

Since I didn't provide customized `blockRenderMap`, so the exception occurred when Draft tried to find out the wrapper.

I think it is not a normal behavior, because according to Draft's doc:
> Any block that is not recognized by the block redering mapping will be treated as unstyled

So I think the code should be changed as:

```javascript
const configForType = blockRenderMap.get(blockType) || blockRenderMap.get('unstyled'); // add a default value
const wrapperTemplate = configForType.wrapper;
```

hellendag:
That seems like a reasonable solution. Feel free to send a PR, it sounds like it'll be a simple one to merge.

southerncross:
No problem 😄

mitermayer:
This issue has been already fixed by https://github.com/facebook/draft-js/commit/3da87b518fa090da1c30b866914fe06586bb97a2

PR #1302, opened 2017-07-23
Add default value for blockRenderMap

southerncross:
**Summary**

PR for issue 1211
https://github.com/facebook/draft-js/issues/1211

**Test Plan**

<s>no need</s>
**Edit by Flarnie:** added unit test


flarnie:
Would still like to get this fix in - I'll add a quick test based on the example from the issue.

PR #1304, opened 2017-07-27
Remove old examples and move examples for 0.10 to /examples

juliankrispel:
- [x] remove 0.9 examples.
- [x] move 0.10 examples to /examples.
- [x] update all references to the above.

cc @flarnie 

addresses #1036


mitermayer:
cc @flarnie,

niveditc:
@juliankrispel - seems that we should merge this in, independent of the 0.11 release. It's always confused me that we've had examples from two versions in the repo.

What do you think? If it sounds good, please rebase to master & then I can merge it in :)

PR #1307, opened 2017-07-27
0.11.0 alpha

tberman:
This merged master with 0.11.0.

All tests pass.


flarnie:
Thanks again for doing this - I was just clicking around in the examples and noticed the 'universal' one doesn't build for me on this branch. Everything else looks good, so I plan on releasing the alpha soon.

tberman:
@flarnie I will take a look at that, it may be a pathing issue or it may just be missing from a build process.

What do you think about a patch that changes the keys in the entityMap to be random text similar to a ContentBlock's key?

PR #1314, opened 2017-07-29
Merge again and use random keys instead of autoincrement integers

tberman:
This merges 0.11.0-alpha up again w/ master.

It also adds a change to use a random key for entity keys instead of an auto-increment integer. The auto-increment integer can cause problems with entity keys being always reused if you are serializing content state including the entity map and reloading it.

This doesn't remove the possibility of duplicate keys, but it dramatically lowers the chance, since the first entity created isn't always a 0!


tberman:
Lemme get rid of this PR and split it up so that its easier to merge.

PR #1315, opened 2017-07-29
Use a random key instead of an autoincrement starting at 0

tberman:
By always starting entityIDs at 0 a user would always get entityID reuse when serializing entity state between browser refreshes.

This should make it significantly less likely to have a collision.

### DIFF
--- a/src/model/encoding/EntityRange.js
+++ b/src/model/encoding/EntityRange.js
@@ -14,12 +14,9 @@
 
 /**
  * A plain object representation of an entity attribution.
- *
- * The `key` value corresponds to the key of the entity in the `entityMap` of
- * a `ComposedText` object, not for use with `DraftEntity.get()`.
  */
 export type EntityRange = {
-  key: number,
+  key: string,
   offset: number,
   length: number,
 };
--- a/src/model/encoding/__tests__/encodeEntityRanges-test.js
+++ b/src/model/encoding/__tests__/encodeEntityRanges-test.js
@@ -37,22 +37,22 @@ describe('encodeEntityRanges', () => {
 
   it('must return an empty array if no entities present', () => {
     var block = createBlock(' '.repeat(20), Repeat(null, 20).toArray());
-    var encoded = encodeEntityRanges(block, {});
+    var encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([]);
 
-    encoded = encodeEntityRanges(block, {'0': '0'});
+    encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([]);
   });
 
   it('must return ranges with the storage-mapped key', () => {
     var entities = [null, null, '6', '6', null];
     var block = createBlock(' '.repeat(entities.length), entities);
-    var encoded = encodeEntityRanges(block, {'_6': '0'});
+    var encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([
       {
         offset: 2,
         length: 2,
-        key: 0,
+        key: '6',
       },
     ]);
   });
@@ -60,17 +60,17 @@ describe('encodeEntityRanges', () => {
   it('must return ranges with multiple entities present', () => {
     var entities = [null, null, '6', '6', null, '8', '8', null];
     var block = createBlock(' '.repeat(entities.length), entities);
-    var encoded = encodeEntityRanges(block, {'_6': '0', '_8': '1'});
+    var encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([
       {
         offset: 2,
         length: 2,
-        key: 0,
+        key: '6',
       },
       {
         offset: 5,
         length: 2,
-        key: 1,
+        key: '8',
       },
     ]);
   });
@@ -78,17 +78,17 @@ describe('encodeEntityRanges', () => {
   it('must return ranges with an entity present more than once', () => {
     var entities = [null, null, '6', '6', null, '6', '6', null];
     var block = createBlock(' '.repeat(entities.length), entities);
-    var encoded = encodeEntityRanges(block, {'_6': '0', '_8': '1'});
+    var encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([
       {
         offset: 2,
         length: 2,
-        key: 0,
+        key: '6',
       },
       {
         offset: 5,
         length: 2,
-        key: 0,
+        key: '6',
       },
     ]);
   });
@@ -96,23 +96,37 @@ describe('encodeEntityRanges', () => {
   it('must handle ranges that include surrogate pairs', () => {
     var str = 'Take a \uD83D\uDCF7 #selfie';
     var entities = [
-      null, null, null, null, null, null, // `Take a`
-      '6', '6', '6', '6', '6', '6',       // ` [camera] #s`
-      null, null, '8', '8', null,         // `elfie`
+      null,
+      null,
+      null,
+      null,
+      null,
+      null, // `Take a`
+      '6',
+      '6',
+      '6',
+      '6',
+      '6',
+      '6', // ` [camera] #s`
+      null,
+      null,
+      '8',
+      '8',
+      null, // `elfie`
     ];
 
     var block = createBlock(str, entities);
-    var encoded = encodeEntityRanges(block, {'_6': '0', '_8': '1'});
+    var encoded = encodeEntityRanges(block);
     expect(encoded).toEqual([
       {
         offset: 6,
         length: 5,
-        key: 0,
+        key: '6',
       },
       {
         offset: 13,
         length: 2,
-        key: 1,
+        key: '8',
       },
     ]);
   });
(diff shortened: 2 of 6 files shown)


flarnie:
This seems ok, and still I want to check about something - Here is what it sounds like we are talking about:
1. `contentState.createEntity(...)` and then the `EditorState` is serialized and persisted to the server.
2. User refreshes the browser, and the raw serialized `EditorState` is sent back to the client.
3. To re-hydrate the raw `EditorState` we need to *create a new DraftEntity for each of the entity ranges in the raw `EditorState`. You can't re-use the old `DraftEntityKey` because the client-side Entity store has been reset and doesn't have those entities any more.

For example - [in `convertFromRawToDraftState` we call `DraftEntity.create` to  create a DraftEntity instance for each entity range in the raw state](https://github.com/facebook/draft-js/blob/514490b1322e1c123524eae97d4aea25b4da16ce/src/model/encoding/convertFromRawToDraftState.js#L41).

What do you think of adding an informative '__DEV__' only warning asking people to create a new DraftEntity instance instead of trying to re-use a key from the server?

I've seen various permutations of this within FB, of varying correctness. I think we need better docs/examples around this too.

tberman:
So in our use-case we are serializing the raw content state across refreshes, and that has both the entities and the blocks. I would expect most people who are serializing are starting from a similar place.

Our issue is when we do that the first entity created in the new session always uses a key of 0 which is a collision every time.

flarnie:
Thanks for explaining again - I'm still looking for more context on how you end up getting a collision with the entity IDs.

This example shows generally what I expect folks do: https://jsfiddle.net/3x41ehbg/28/

The serialized raw content state is fetched in the `componentDidMount`, and you delay letting users create entities until that has loaded.

Then we would instantiate the `EditorState` like so:
```
    const contentState = convertFromRaw(rawContentStateFromServer);
    const editorState = 
      EditorState.createWithContent(contentState, decorator);
    this.setState({
      editorState,
    });
```

I am thinking there could be a collision if you let the user create new entities before loading the serialized `contentState`. Maybe it also would come up if you are loading content into multiple DraftEditor components?

flarnie:
Thanks for your patience - I do believe you that the collision issue is real. 

Also it looks like `convertFromRaw` ignores the keys from the server, because it creates a fresh instance of the `DraftEntity` for each entity range. So long as your server data is consistent between the `blockMap` and the `entityMap` then your keys could be anything. Here I turned the first key into `test` in the server data, and it still gets reset to `0` in the editor state.
https://jsfiddle.net/nko379wx/

If you are using `convertFromRaw` or a similar method it should create a new Entity when loading the serialized raw content state: https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromRawToDraftState.js#L38-L42

Do you mean it's a collision on the server? In that case it would make more sense to me.

tberman:
So here is the issue:

1) Document is created in browser session 1. An entity is created by uploading an image. It has a key of 0 given the way entities are created.
2) Document is serialized to the backend.
3) Document is loaded in browser session 2. This hydrates a content state from the serialized DraftRawContentState that the server gives it. In this, there is an entity with a key of 0 that is created, and the `key` variable in src/model/transaction/addEntityToEntityMap.js is 0 since this is a new browser session.
4) A new image entity is created, it is given a key of 0, overwriting the previous key with the id 0.

This is all using convertFromRaw/convertToRaw for serialization and deserialization.

tberman:
> Also it looks like convertFromRaw ignores the keys from the server, because it creates a fresh instance of the DraftEntity for each entity range. So long as your server data is consistent between the blockMap and the entityMap then your keys could be anything. Here I turned the first key into test in the server data, and it still gets reset to 0 in the editor state.
> https://jsfiddle.net/nko379wx/

This should likely also be fixed, as it should re-use the key that was given, since you should be able to roundtrip from raw -> Draft -> raw and get back the equal objects.

tberman:
> Maybe it also would come up if you are loading content into multiple DraftEditor components?

We do have multiples in a single page.

tberman:
Looking at it more, it appears that the other issue we have is that we don't retain keys as we roundtrip the data through Draft. We are doing some diffing between state snapshots, and Draft re-assigning the ids on conversion is definitely not great for us.

It may make sense for me to put together a different patch as well where it keeps the ids from the server.

Not sure how to resolve the multiple editors on a page issue w/o that.

Thoughts?

flarnie:
Thanks for talking through this @tberman - I think I understand now. It sounds like you have multiple editors on a page, and also something that triggers the content to go to the server, then come back, and you want to 'refresh' the current editor with the responts from the server. I think you are right that there still could be issues, since I don't think we were supporting cases of `raw -> Draft -> raw`.

I was just looking around at FB for any similar use cases to see what they did, but I don't think we have cases quite like this. The cases I saw do a full re-render of the `DraftEditor` and associated components when they get `contentState` from the server, or just don't support loading 'WIP' content from the server.

This does seem like it would come up though. It would be cool to have an `example` that covers some of this functionality, I think. And/or tests. That can also help me understand how it should work.

tberman:
We are doing live collaborative editing, which means we run into some especially 'fun' use cases. I think ours boils down to the ids generated by client 1 and client 2 can collide and cause issues in ways that other use cases may not.

Let me spend some time later digging in more, because this is definitely an issue we see regularly, and I am not 100% sure why given the great digging you are doing.

flarnie:
> We are doing live collaborative editing, which means we run into some especially 'fun' use cases.

Cool project! I believe you will run into some interesting challenges. :)

cc @mitermayer && @SamHowie who worked on a similar project in the past. They would probably have some great advice for you. Might be worth doing some other kind of intro, but for now tagging them on this PR seemed like the easiest way to connect you.

CameronAckermanSEL:
I have a similar problem, except mine sort of comes from pushing draft beyond the typical use case. 

In the software I am working on

1. Each block is assigned a unique key that we generate an inject into the content state because each block needs to be able to be retrieved and stored separately.
2. Each block is has its own arbitrary properties that users will need to bind to it by key. 
3. Blocks may need to be reused in other documents.

We haven't worked out how entities are going to fit in this system because draft doesn't accept keys that are not monotonically increasing integers, which is honestly kind of a strange design decision. I realize that most users will not need entities to have significant/unique keys, but for our system because blocks are reusable and are significant units of information, we need to have the ability to assign stronger keys to the entities. 

This PR would greatly simply some workarounds that I have to get around my problems. I think the most ideal solution would be the ability to give draft some delegates that can be used to generate block and entity keys.

tberman:
I am going to extend this branch to allow importing/setting of entity ids and will put it up for review when I get that working. Ultimately being able to roundtrip the entity ids is very important to us!

CameronAckermanSEL:
Oh man, thanks for doing this. I'm very excited to use it!

tberman:
@CameronAckermanSEL I have not yet tested it fully, but it seems to be doing what I want.

mitermayer:
@tberman On a previous project we managed to have collaborative editing using draft. In order for that to work we had to massage the draft raw format prior to sending to synchronization service.

@deliminator worked on the serialization of the draf  raw format. He may be able to add some advice or more info in this topic in here

adding  Tobias (@deliminator) to the thread

terencechow:
Hmmm, can anyone provide more guidance on collaborative editors? the @deliminator handle doesn't seem to work anymore and he hasn't responded


@mitermayer what do you mean by massaging the raw message before sending? What were you 'massaging' and why?

@tberman Can you provide a bit more detail on your approach? 

I'm trying to do collaborative editing and the current thought process for me is to hold a "source of truth" contentState on a websocket server. Client 1 makes some changes and sends it to the server. The server checks for any added blocks and adds them by key (easy part). For all existing blocks it will have to check if the entityRanges or inlineStyleRanges have been added or removed. This seems like it could become quite cumbersome and prone to bugs. Hence my hope that someone might be able to provide more guidance on their approach.

tberman:
Our approach is to use a json-based OT protocol across the wire and then reimplement it on the backend in go for storage. We are also going to evaluate a CRDT based approach as well.

ukd1:
Jumping in here - this isn't ideal and will cause slowness (due to having to store and check for collisions) and or collisions anyway, you're just reducing the chance of it by not starting from zero. The real fix here, imho, is to prefix incremented IDs with a UUID or similar, instead of using random each time.

flarnie:
Sorry about the lag in response to this PR - 
In order to consider this or the approach that @ukd1 suggested, we need a clear understanding of the problem that is being solved. I would want to see a repro case or otherwise get more clarity on that. Feel free to reopen with that info, sorry for the lag in response to this PR. For now closing this since it has not been active, but would consider if reopened with more info.

tberman:
The problem being solved is that in the case where you have a single page app that can have multiple editors over time (or multiple editors on a page) and you are also doing OT, the keys will not be even approximately unique and you will get a collision basically right away.

What if we exposed a key generation function that was used when keys were needed, and that can be customized for the use-case that the user has?

CameronAckermanSEL:
The auto increment keys for entities created an unnecessary amount of difficulty in our system and caused us to migrate to slate.

ukd1:
@tberman not the case, due to prefixing per instance.

ukd1:
@flarnie I'm no longer interested in pushing this further; if what I've explained isn't clear enough no problem! 👍

PR #1318, opened 2017-08-01
Update docs to clarify return value for handlePastedText

icd2k3:
**Summary**

Docs for `handlePastedText` are a bit unclear about what it should be returning. According to the [Cancelable Handlers](https://draftjs.org/docs/api-reference-editor.html#cancelable-handlers-optional) section, all should return `'handled'` or `'not-handled'`, but the section for `handlePastedText` mentions returning a boolean. This has caused some confusion in draft-js-plugins-editor here: https://github.com/draft-js-plugins/draft-js-plugins/issues/806

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -205,7 +205,7 @@ and to convert typed emoticons into images.
 ```
 handlePastedText?: (text: string, html?: string) => DraftHandleValue
 ```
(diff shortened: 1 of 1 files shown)


icd2k3:
Closing in favor of fixing `draft-js-plugins-editor` to accept boolean returns in addition to `'handled'|'not-handled'` see: https://github.com/draft-js-plugins/draft-js-plugins/issues/806

PR #1337, opened 2017-08-14
safety handle empty block arrays in raw states

mattkrick:
**Summary**

Currently, it is possible to make raw objects that have an empty block array.
These raw objects, when rehydrated, return undefined for methods like `getFirstBlock` since no blocks exist. A contentState without a single block is dangerous & can cause errors, so let's eliminate the possibility of that.

**Test Plan**

Happy to write tests, wanna get the general idea approved before doing so.


mitermayer:
Hi @mattkrick,

Thanks for submitting this PR, this behaviour sounds reasonable to me. You will need to rebase this first since that method has changed in master, and tests would be great.

Let me know if you are still interested on working on this

PR #1342, opened 2017-08-19
Support multiple entities on a single character

tberman:
Note: This also includes my changes from https://github.com/facebook/draft-js/pull/1315 which I can pull out if that is valuable.

This is a breaking change, so not sure how we would want to merge it, but I wanted to put it up for discussion before spending more time on mitigating the API breakage.

Love any thoughts!


flarnie:
Hi again Todd! Thanks for adding this. I know we discussed your use case in some detail on https://github.com/facebook/draft-js/pull/1315, and it would still be useful to know what new feature/abilities this change could enable us to do with Draft, in order to justify the added complexity.

I'm putting this and https://github.com/facebook/draft-js/pull/1315 on the agenda for discussing with the other Draft.js maintainers, will review it in more detail when I get time. I don't know of any blockers off the top of my head though. :)

flarnie:
I also wonder if this would be of interest as an alternative or supplement to the nested block refactor you have worked on @mitermayer

tberman:
So, I am using this to enable support for entities that look like a link across the middle of a paragraph and then a range that represents an inline comment across some of the link and some text outside of it. In our use case we actually have some pretty common cases where characters have upwards of 3 entities frequently.

For us, we definitely need both stable entity keys and multiple keys per character.

fraziermork:
This would be fantastic for my use case as well. I have ranges of text that, based on the results of an api request, get tagged with one or more pieces of metadata. Right now we're handling this by computing the ranges of each piece of metadata, figuring out the sections that have are tagged with a unique set of metadata, and then applying an entity to each those ranges. 

It would be substantially simpler if I could just tag each range of metadata with it's own entity. That would let me change/remove each tag independently of the others (right now that means having to recompute all of the ones that it overlaps with).

flarnie:
Hi again - we discussed this change a bit, and we're not opposed to this - we would need to approach it in a thoughtful way and minimize breaking changes. For example, would the current `addEntity` method overwrite an existing entity range, or be additive?

Also to be honest it's a big change and we (the core maintainers) won't have time to review it any time soon. We are still cleaning up the aftermath of some other breaking changes.

tberman:
I am happy to help in any way to get it upstreamed, we are depending on it so right now I am maintaining a fork of draft with this change which is no fun long term.

In this patch I just broke API (including getting rid of `applyEntity` and replacing it with `addEntity` and `removeEntity`).

As far as maintaining API, I intentionally kept it as an OrderedSet so in theory we could do the following:

1. Make `getEntity` return the first entity in the list, also give it a deprecation warning in 0.11 with support to be removed in 0.12.
1. Bring `applyEntity` back and have it work the same as it used to, where `null` clears all entities and non-null clears all and sets the only entity to the new value. Additionally, have applyEntity give a deprecation warning in 0.11 with support to be removed in 0.12.
1. Add a new method `getEntities` that returns the OrderedSet.
1. Leave `addEntity` and `removeEntity` and the other related changes in place for 0.11 and have them become the 0.12 API moving forward. This also looks like what is going on for inline styles, so the change should be coherent for the end-user.

Thoughts?

flarnie:
Hi again Todd - 
I have to let you know that I don't think we will be able to merge this any time soon.
For the next 6 months, Draft.js maintainers are focused on wrapping up other WIP API migrations and improving stability of existing features. We won't be able to support any new breaking changes or significant rewrites.

In about June it might be worth reaching out about expanding Draft.js features. My hope is that you're able to continue using your rewritten fork of Draft, or even use it as a basis for something new. We are rooting for you and can help answer questions as things come up.

xiaopow:
Currently, seems like only one entity can be added to a selection. How would I be able to implement highlight, color inline styles and link on an identical selection?

PR #1344, opened 2017-08-22
add support for data object to insertAtomicBlock function

willisplummer:
**Summary**

Adds an optional data parameter to the `AtomicBlockUtils.insertAtomicBlock` function. This is helpful if you don't need an entity and simply want to associate some metadata with the block itself. This is also a request that I've seen come up in the slack a few times

https://draftjs.slack.com/archives/C0PBTNQGY/p1503009836000334

If I'm missing something about why it's better to associate the metadata with an entity than directly with the atomic block, I'm all ears to that as well :)


willisplummer:
looks like builds are broken on master because of the flow stuff

would still love a review if anyone has time

PR #1346, opened 2017-08-23
Update handleKeyCommand examples and docs

aadsm:
**Summary**
handleKeyCommand receives a second parameter with the latest instance of the editor state. When this function needs the editor state it should use this instance instead of any other reference.

This update reflects a change made to handleKeyCommand to also receive the editor state. Since it can be potentially changed by Draft during key handling any previous references might be of an old version of the state.

**Test Plan**
Loaded `examples/draft-0-10-0/{media/media.html, rich/rich.html, tex/}` in the browser and verified the text input worked as expected and in particular that we could delete deadkeys like `.

### DIFF
--- a/examples/draft-0-10-0/rich/rich.html
+++ b/examples/draft-0-10-0/rich/rich.html
@@ -43,14 +43,13 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
           this.focus = () => this.refs.editor.focus();
           this.onChange = (editorState) => this.setState({editorState});
 
-          this.handleKeyCommand = (command) => this._handleKeyCommand(command);
-          this.onTab = (e) => this._onTab(e);
-          this.toggleBlockType = (type) => this._toggleBlockType(type);
-          this.toggleInlineStyle = (style) => this._toggleInlineStyle(style);
+          this.handleKeyCommand = this._handleKeyCommand.bind(this);
+          this.onTab = this._onTab.bind(this);
+          this.toggleBlockType = this._toggleBlockType.bind(this);
+          this.toggleInlineStyle = this._toggleInlineStyle.bind(this);
         }
 
-        _handleKeyCommand(command) {
-          const {editorState} = this.state;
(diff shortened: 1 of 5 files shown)


aadsm:
@flarnie, for some reason I'm not able to reproduce the flow and lint errors on my dev machine (the errors are also unrelated to my changes, so they're in master right now). I could easily put a PR up for the lint errors though as they are just about the "80 column" limit.

examples/draft-0-10-0/rich/rich.html
```diff
@@ -43,14 +43,13 @@
           this.focus = () => this.refs.editor.focus();
           this.onChange = (editorState) => this.setState({editorState});
 
-          this.handleKeyCommand = (command) => this._handleKeyCommand(command);
+          this.handleKeyCommand = this._handleKeyCommand.bind(this);
```
spicyj: What do you think of `(cmd, state) => this._handleKeyCommand(cmd, state)` instead?

aadsm: Initially I wrote that with `command` but it was greater than 80 columns. Then I decided to use `bind` because it's literally what we want (have the right `this` inside `_handleKeyCommand` and no need to update in the future if more parameters are added) and because it's also the style in other examples.
That was my rationale but I can update it to your suggestion if you prefer, I don't have a strong opinion here.

spicyj: If other examples have bind already let's do bind I guess.

aadsm: Yeah, this is the only one who doesn't actually, just checked. Both link and media do as well. I'll update this to make the other 3 functions use bind as well.

aadsm:
Do I also need to update `website/core/metadata.js` in this PR? or is this something only done in `gh-pages` branch?

spicyj:
Not sure, sorry. I think it's probably OK without changing.

flarnie:
Just added this to a list of PRs to merge after we fix CI by landing https://github.com/facebook/draft-js/pull/1349

Thanks for your patience~ ⏳🙏✨

PR #1355, opened 2017-09-01
Can't type normal text after link creating

peasemaker:
**SUMMARY**
In examples of link after creating link user can't type normal text or put a space, because link will be continued. User can type a character after link and remove link from it, but it's not convenient.

**TEST PLAN**
I think, for entities with type 'LINK' it should be checked current character and character before, and if entity keys are equal continue link, otherwise don't.

Such behavior will be similar with link from gmail editor.


thibaudcolas:
This is issue #430, except the issue goes beyond just links.

niveditc:
This was fixed in https://github.com/facebook/draft-js/commit/d16833b3dae77ccf13e3af7f5e42c8131b0d1d2c

PR #1356, opened 2017-09-01
Fix #1326 Safari can't focus with draggable parent

### LINKED ISSUE #1326: Text typing not working in Safari with draggable parent container
bug

When i add attribute `draggable="true"` on parent div text typing is not working

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/Aydar/ztsvbyep/3/.**

text typing works properly

Safari 10.1.2 (12603.3.8) / macOS Sierra
draft-js 0.10.1 ( i have tried  0.7.0, 0.9.0 )

mattkrick:
**Summary**

DraftEditors that lived inside draggable parents were useless in Safari. Now they work.
Solution found here: https://github.com/react-dnd/react-dnd/issues/178#issuecomment-192021227

**Test Plan**

Use the fiddle in #1326 with the updated version.
For a quick n dirty, just add this to the stylesheet:
`div[contenteditable="true"] { user-select: text;}`


jordanh:
👏 👏 👏

flarnie:
Thanks for submitting this!

The bug is easy to reproduce, and I'm still testing to verify that your change fixes it.

Issue #318, opened 2016-04-17
Not being able to update editor state


alexeygolev:
I wasn't sure how to describe the issue. I noticed the following behaviour while working on simple `Entity.create` scenarios. 
I have the following 

``` jsx
class SimpleEditorComponent extends Component {
  constructor(props) {
    super(props)
    this.state = { editorState: EditorState.createEmpty() }
    this.focus = () => this.refs.editor.focus()
    this.onChange = (editorState) => this.setState({ editorState })
    this.applyEntity = this.applyEntity.bind(this)
    this.logState = () => {
      const content = this.state.editorState.getCurrentContent()
      console.log(convertToRaw(content))
    }
  }
  applyEntity() {
    const { editorState } = this.state
    const entityKey = Entity.create('LINK', 'MUTABLE', { url: 'http://something.com' })
    this.setState({
      editorState: RichUtils.toggleLink(
        editorState,
        editorState.getSelection(),
        entityKey
      ),
    }, () => {
      setTimeout(() => this.refs.editor.focus(), 0)
    })
  }
  render() {
    const { editorState } = this.state
    return (
      <div className="editor" onClick={this.focus}>
        <Editor
          editorState={editorState}
          onChange={this.onChange}
          placeholder="Enter some text..."
          ref="editor"
          />
        <div onClick={this.applyEntity} style={{cursor: 'pointer'}}>Apply entity</div>
        <div onClick={this.logState}>Log editor state</div>
      </div>
    )
  }
}
```

Which is a very basic setup. However the entity is not added to the editor state. In fact if I set editorState to `{}` it won't change. Any other properties I set would set on state in `applyEntity` appear in state on next render. I'm sure it's something obvious but I don't even know where to start. 
I thought it might be a race condition because of the editor losing focus when I click `Apply entity` and then gaining it back. But then nothing would be updated in the state object.

hellendag:
I think your suggestion of a possible race condition is the likely culprit here. Try using `onMouseDown` instead of `onClick`. You might also want to `preventDefault()` on the event.

alexeygolev:
@hellendag `onMouseDown` works, thank you. `preventDefault()` is not needed.

alexeygolev:
This probably should be closed. However maybe we should somehow add it to docs. In fact now when I think about it `onMouseDown` is used in most of the examples so maybe we should add at least a little note about this.

hellendag:
[...]

PR #1371, opened 2017-09-07
Add EditorState Race Conditions documentation

aadsm:
**Summary**
This new documentation page is under "Advanced Topics" and between "Text Direction" and "Issues and Pitfalls".
It documents the ways to have race conditions based on consecutive EditorState updates and best practices to avoid this issue.
This was motivated by issue #318.

It also updates the CSS to make all images' width to 650px which is the width of the right column content, hope that's fine. (this version of markdown (Marked) doesn't allow to set the image width). 

**Test Plan**
Go to /docs/advanced-topics-editorstate-race-conditions.html and check the page loads.

Screenshot:
localhost-8080-docs-advanced-topics-editorstate-race-conditions html


docs/Advanced-Topics-EditorState-Race-Conditions.md
```diff
@@ -0,0 +1,53 @@
...
+
+As a rule of thumb avoid having different event handlers for the same event that manipulate the `EditorState`. Using setTimeout to run `setState` might also land you in the same situation.
+Anytime you feel you're “losing state” make sure you're not overriding it before the `Editor` re-rendering.
+
+## Best Practices
+
+Now that you understand the problem, what can you do to avoid it? In general be mindful of where you're getting the `EditorState` from. If you're using a local one (stored in `this.state`) then there's the potential for it to not be up to date.
+To minimize this problem Draft offers the latest `EditorState` instance in most of its callback functions. In your code you should use the provided `EditorState` instead of your local one to make sure you're basing your changes on the latest one.
+Here's a list of supported callbacks on the `Editor`:
+
```
flarnie: Nice! What would you think of adding an example of the right way to set up an `onPaste` handler? I think it would be something like this:

```
this.onPaste = (text, styles, editorState) => {
  this.setState({
    editorState: removeEditorStyles(text, editorState)
  });
}

//...
<Editor
  onChange={this.onChange}
  placeholder="Enter some text..."
  ref="editor"
  handlePastedText={this.onPaste}
/>
```

I will try to do a little test to verify that this is a good example.

I am also realizing we never updated [the docs](https://draftjs.org/docs/api-reference-editor.html#cancelable-handlers-optional) to show [which `handleX` callbacks receive the `editorState` as an argument.](https://github.com/facebook/draft-js/blob/6d584d4a7bb047a7719e2fe7077976935e26065e/src/component/base/DraftEditorProps.js) If you are open to a follow-up PR that would be great, or we could open a `documentation` `good-first-bug` issue.

flarnie:
PS adding this to the queue for merging and planning to call it out at the weekly meeting. Thanks again - the diagrams and explanation are amazing!

docs/Advanced-Topics-EditorState-Race-Conditions.md
aadsm: That's a really good idea, will add it. My other PR doesn't change the API reference though, I did make a mental note when I put this PR up but then forgot, thanks for reminding!

aadsm:
I've updated the PR with a running example of the race condition and some logging to show the JS timeline.

PR #1372, opened 2017-09-08
Update weekly meeting notes

flarnie:
Adding notes from the past two meetings. Doesn't really need reviewed.


flarnie:
Might as well add the notes from last Friday~

Issue #1364, opened 2017-09-06
Get CI passing


flarnie:
Commits from within FB have broken CI. Working on fixing this in https://github.com/facebook/draft-js/pull/1349

PR #1373, opened 2017-09-08
Add hook in travis to notify FB people when they break CI

flarnie:
**what is the change?:**
Add a hook similar to https://github.com/facebook/relay/commit/a1eff1a49aa511426c09c481e942256c751c9281

**why make this change?:**
We want people who sync commits out from FB to be notified (and maybe
stopped?) if their change breaks CI on Github.

**test plan:**
Do it live

**issue:**
Follow up of https://github.com/facebook/draft-js/issues/1364

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,3 +67,6 @@ script:
     fi
 
   fi
+
+notifications:
+  webhooks: https://code.facebook.com/travis/webhook/


flarnie:
I don't actually know - I think phabricator? Kind of want to try it and see. Paul didn't have time to explain yet but since Relay already did it I'm not too worried. :)

zpao:
Phabricator - meant to catch commits getting synced out that break CI for people not paying attention to GH or might not even be on GH. cf https://fburl.com/k0hctrfc

PR #1374, opened 2017-09-08
Update callbacks reference to include the EditorState

aadsm:
**Summary**
These 4 `Editor`'s properties: `handleReturn`, `handleKeyCommand`, `handleBeforeInput` and `handlePastedText` now also pass the latest editor state.

This is a follow up on #1371.

**Test Plan**
Go to `/docs/api-reference-editor.html#cancelable-handlers-optional` and check the method's signatures include the editorState.
Screenshot:
screen shot 2017-09-08 at 11 07 23 am


flarnie:
Land at will, and/or I will land this when I get time.

Issue #1377, opened 2017-09-09
Fix image rendering


flarnie:
**Do you want to request a *feature* or report a *bug*?**
A bug

**What is the current behavior?**
In Firefox and Chrome (haven't tested other browsers) since https://github.com/facebook/draft-js/commit/0b22d713d556ccc4820850099ad6231493b3f7aa we don't get the image blocks when using `convertFromHTMLToContentBlocks`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**

You can see this in the `examples/draft-0-10-0/convertFromHTML/convert.html` file.

**What is the expected behavior?**

It should render an image.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

Unreleased version of Draft.js, is only on master.

flarnie:
This should have been fixed. :)

michelson:
Hi @flarnie , in which version this is fixed ? Using 0.10.5 we I've hit this behavior .

Issue #231, opened 2016-03-25
Empty blocks not included by convertFromHTML


benbriggs:
HTML equivalent to a block with no text are omitted from the output of `Draft.convertFromHTML`.

Steps to reproduce:

```
const html = '<div>first block</div><div></div><div>third block</div>';
Draft.convertFromHTML(html) //returns an array of 2 blocks (the non-empty ones) instead of three
```

@hellendag mentioned in Slack that @CLowbrow might know most about this - I'm happy to PR a change to fix it but could use a bit more context on what the [condition](https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L183-L191) in `joinChunks` is used and not used for. Looks like this is the source of the removal of the block when it's called from [here](https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L361). When removing that last character of chunk `A` does its block ever need to be popped? My interpretation is that chunk `A` has a delimiter and chunk `B` is empty since its first character is `\r`.

Thank you!

CLowbrow:
This was on purpose to keep pastes of long text from having a lot of extra spaces between paragraphs. The content coming in can be pretty messy sometimes and this is one of the things we try to do to clean it up. I'm open to modifying this. What kinds of things are you trying to preserve?

benbriggs:
Ah I see. I'm using this module along with another that I built to go to/from HTML for saving the contents, so the empty block tags are actually coming from a transformation of `ContentState`. I didn't use this as my initial example since it added extra complexity but having three `ordered-list-item` blocks:

1. first
2. 
3. third

That `ContentState` is going to have three blocks that'll map to `<ol><li>first</li><li></li><li>third</li></ol>`. Throwing that back into `convertFromHTML` omits that middle block making the last one which causes noticeable differences from what Draft was rendering originally.

CLowbrow:
Storing draft.js content as html is probably not the way to go. Internally, we store a serialized content state and just re-instantiate/render from that. The paste handler is more tailored to minimize the amount of follow-up editing after a paste than faithfully recreate document states.

ianstormtaylor:
@CLowbrow question, what are the negatives you'd point to to storing content as HTML?

[...]

PR #1378, opened 2017-09-10
Make default text content of img blocks a little camera emoji

flarnie:
**what is the change?:**
We don't allow pasting whitespace-only blocks - see https://github.com/facebook/draft-js/issues/231

In the past we set the 'src' as the text content of an 'img' block when
pasting/converting HTML, but sometimes the 'src' is really long and can
even crash the editor. Plus that's not really what users expect when
they paste an image.

An empty string as a default would be better but the code in
`convertFromHTML` needs refactored and for now I don't want to touch it.
The camera emoji is a cute and easy fix.

**why make this change?:**
Right now images are being dropped by `convertFromHTML

**test plan:**
Manual testing with examples/draft-0-10-0/convertFromHTML/convert.html

We should do a follow-up change to add a test that catches this.

**issue:**
https://github.com/facebook/draft-js/issues/1377

4. Ensure that:

-

**Summary**

[...]

**Test Plan**

[...]

### DIFF
--- a/src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
+++ b/src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
@@ -50,18 +50,18 @@ describe('convertFromHTMLToContentBlocks', () => {
   ].forEach(tag => testConvertingAdjacentHtmlElementsToContentBlocks(tag));
 
   describe('img tag', function() {
-    test('img with http protocol should have empty content', function() {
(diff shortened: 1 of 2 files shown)


flarnie:
For future reference - here is where the space-only blocks are dropped:

https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L285-L296

aadsm:
I love the camera emoji replacement!

Issue #219, opened 2016-03-17
Binding left & right arrow keys?


coryasato:
Doesn't fall through to the default behavior when returning false from a custom key handler.  Crawled the source code but couldn't find anywhere this is being handled.  I guess I could manually reset the selection every time after Im done listening in.  Perhaps theres an easier way?

hellendag:
What is your use case?

coryasato:
I'm trying to listen if the cursor is currently in an inline that has an Entity type of LINK to fire off a custom popover component to edit the href.  This will do a bit more, but for now baby steps.  Currently I can achieve this in the onChange event and a little tweak to the 'currentBlockContainsLink' helper.

hellendag:
If you don't mind positioning the flyout absolutely, outside the `Editor` itself, you can declaratively set whether to render it based on the current selection state.

`getVisibleSelectionRect` may need to be modified for the collapsed cursor, but otherwise, I'm thinking that it could be something like this:

```javascript
render() {
  return (
    <div>
     <Editor ... />
     <LinkMenu editorState={this.state.editorState} />
    </div>
  );
}

class LinkMenu extends React.Component {
  render() {
    const isWithinLink = ...;
    if (isWithinLink) {
      const targetRect = getVisibleSelectionRect(...);
      return <div style={/* use targetRect to position */}>...</div>;
    }
    return <div />;
  }
}
```

coryasato:
Yup, thats pretty much what I'm doing. :)

We can close this issue if you think that binding to the left / right arrow keys is unnecessary or dangerous.  Thanks!

hellendag:
I think they could be exposed the same way as the up and down arrow keys, especially if it allows better keyboard navigation for flyouts. For instance, I could imagine a mentions result dropdown that allows two dimensions of navigation.

I haven't seen a need for it yet, so I'm fine with holding off for now. But if anyone should see this issue and want to offer up a PR to solve a need, I'm open to it. :)

PR #1384, opened 2017-09-12
Right and Left Arrow Event Handling

eessex:
Exposes right and left keyboard events, as suggested in [this issue](https://github.com/facebook/draft-js/issues/219) by @hellendag.

This will greatly simplify some of our arrow handling at [Artsy](https://github.com/artsy/positron/blob/master/client/apps/edit/components/content2/sections/text/index.coffee#L121).

The test suite is passing locally.


eessex:
This build was not able to complete, but the tests are passing locally.  Is it possible to trigger a rebuild?

flarnie:
Hi @eessex! Thanks for adding this.

I just triggered a rebuild, and in general you can also trigger a rebuild by rebasing your branch on master and force pushing your branch. Looks like tests were just failing spuriously, probably due to some issue with CI itself.

I think this looks good and makes sense. I'm trying to manually test it and when I manually added an `onLeftArrow` prop to the `rich` example, with a callback, the callback wasn't being called.

Going to debug this a little more and see if I can figure out what I'm missing.

eessex:
Thanks @flarnie! I'll try some manual debugging on my side as well

flarnie:
Maybe I forgot to re-run `yarn build` or something - it works for me now. Thanks!

happy_golden_retriever

Adding this to the queue for merging.

Issue #14, opened 2016-02-22
Demo Site: Scrolling down and clicking into Textarea jumps viewport to top.


thoughtpalette:
Chrome Version 48.0.2564.109 (64-bit)/Yosemite.

![](https://i.gyazo.com/e2c1671ad46bec9df1a2f090f06915a9.gif)

Steps:
1: Scroll down page
2: Click textarea

hellendag:
Weird. Are you using any browser plugins or extensions?

thoughtpalette:
Just Privacy Badger. Though the behavior is reproducible with or without it enabled. :|

hellendag:
I don't think I've seen this, and we can't repro it here. Do you see the same issue when focusing a comment input on www.facebook.com?

tjisousa:
I'm using the exact same Chrome version as you and I can't reproduce this.

thoughtpalette:
Hmmm. Just tested on an incognito tab with no add-ons and I still have the same behavior. 

I'm not able to reproduce this behavior on an FB comment input. 

I do however receive some react based console errors/warnings. Probably unrelated.

<img width="806" alt="screen shot 2016-02-22 at 4 31 27 pm" src="https://cloud.githubusercontent.com/assets/690624/13235100/c82888da-d981-11e5-831b-64c135b5bdaf.png">

JaRail:
Seeing similar issues in Safari on the Draft.js site.

Zoom in on text box, click somewhere to position to the cursor.

Result: Even though the textbox is completely in view, the viewport jumps to left: 0 and some top: 0+scroll(?) offset.

The window size can affect if you can scroll. So depending on if you've scrolled (unzoomed) or not, you might be jumped to the top-left or some lower position based on the scroll..

knowbody:
works fine for me on Chrome Version 48.0.2564.109 (64-bit) / Yosemite

brwr:
I'm able to reproduce in Chrome Version 48.0.2564.103 (64-bit) on OS X El Capitan Version 10.11.1.

Maybe an El Capitan problem? I've noticed a lot of problems with various projects/programs/things since upgrading.

hellendag:
I can repro the jump in Safari described by @JaRail with pinch+zoom. Not sure if it's the same issue, but still worth investigating.

Still can't repro the original issue in Chrome. I'm also on El Capitan 10.11.3.

JaRail:
**Also repros in the FB feed!**

[...]

PR #1401, opened 2017-09-22
Force update of  `fbjs` to get fix for #14, #1381

### LINKED ISSUE #1381: `focus()` cause unexpected scroll
bug

On chrome 61.0.3163.79, the view will scroll to top if the editor's `focus()` method were called.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Checkout this: https://jsfiddle.net/85t88Lat/

Just get the editor focused, without the extra scroll.

Draftjs/0.10.1 Chrome/61.0.3163.79

flarnie:
**what is the change?:**
Pin the `fbjs` dependency version to 0.8.15 or later.

**why make this change?:**
We need to get everyone updated to use the fixed version, to avoid a
scroll issue in Chrome 61+
See https://github.com/facebook/fbjs/pull/229

**test plan:**
Didn't test because we already are using this fix internally at FB.

**issue:**
fixes #1381

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "enzyme": "^2.9.1",
-    "fbjs": ">=0.8.15",
+    "fbjs": "^0.8.15",
     "immutable": "~3.7.4",
     "object-assign": "^4.1.0"
   },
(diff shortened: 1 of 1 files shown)


sophiebits:
(If people re-npm-install they'll get the new version already.)

flarnie:
Sure thing~

PR #1407, opened 2017-09-25
Fix when anchorNode is undefined instead of null

### LINKED ISSUE #1399: 0.10.2 fails on a test: global.getSelection().anchorNode undefined
A test with these 3 lines with an error. Removing the last line make it to succeed.
It uses enzyme.

```js
wrapper.find('.public-DraftEditor-content').simulate('beforeInput', { data: '@' });
wrapper.find('.public-DraftEditor-content').simulate('keydown', { which: KEY_ESC });
wrapper.find('.public-DraftEditor-content').simulate('beforeInput', { data: 't' });
```

So it seems this line `anchorNode` is undefined, not `null`.
https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnBeforeInput.js#L159

```
TypeError: Cannot read property 'nodeType' of undefined
      
      at editOnBeforeInput (node_modules/draft-js/lib/editOnBeforeInput.js:112:74)
      at node_modules/draft-js/lib/DraftEditor.react.js:146:19
      at Object.invokeGuardedCallback [as invokeGuardedCallbackWithCatch] (node_modules/react-dom/lib/ReactErrorUtils.js:26:5)
      at executeDispatch (node_modules/react-dom/lib/EventPluginUtils.js:83:21)
      at Object.executeDispatchesInOrder (node_modules/react-dom/lib/EventPluginUtils.js:108:5)
      at executeDispatchesAndRelease (node_modules/react-dom/lib/EventPluginHub.js:43:22)
      at executeDispatchesAndReleaseSimulated (node_modules/react-dom/lib/EventPluginHub.js:51:10)
      at forEachAccumulated (node_modules/react-dom/lib/forEachAccumulated.js:26:8)
      at Object.processEventQueue (node_modules/react-dom/lib/EventPluginHub.js:252:7)
      at node_modules/react-dom/lib/ReactTestUtils.js:351:22
      at ReactDefaultBatchingStrategyTransaction.perform (node_modules/react-dom/lib/Transaction.js:143:20)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactDefaultBatchingStrategy.js:62:26)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactUpdates.js:97:27)
      at node_modules/react-dom/lib/ReactTestUtils.js:349:18
      at ReactWrapper.<anonymous> (node_modules/enzyme/build/ReactWrapper.js:776:11)
      at ReactWrapper.single (node_modules/enzyme/build/ReactWrapper.js:1421:25)
      at ReactWrapper.simulate (node_modules/enzyme/build/ReactWrapper.js:769:14)
```

tleunen:
Fixes #1399

`nativeSelection.anchorNode` can be undefined instead of null, so I'm changing the condition to handle this use case.

### DIFF
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -155,7 +155,7 @@ function editOnBeforeInput(
     const nativeSelection = global.getSelection();
     // Selection is necessarily collapsed at this point due to earlier check.
     if (
-      nativeSelection.anchorNode !== null &&
+      nativeSelection.anchorNode &&
       nativeSelection.anchorNode.nodeType === Node.TEXT_NODE
     ) {
       // See isTabHTMLSpanElement in chromium EditingUtilities.cpp.


src/component/handlers/edit/editOnBeforeInput.js
```diff
@@ -155,7 +155,7 @@ function editOnBeforeInput(
     const nativeSelection = global.getSelection();
     // Selection is necessarily collapsed at this point due to earlier check.
     if (
-      nativeSelection.anchorNode !== null &&
+      nativeSelection.anchorNode &&
```
flarnie: I generally prefer to be more specific when checking for a value to be defined, but in this case we don't want any other falsy JS values either, so I think this is ok.

sophiebits:
In what situation can anchorNode be undefined?

tleunen:
@sophiebits You can see an example in #1399 when using Enzyme. But I don't have access to that specific bit of code to really investigate why

PR #1409, opened 2017-09-27
Support React v16 when editor is focused after removed from DOM

mattkrick:
**Summary**
- calling `Style.getScrollParent(editorNode)` with `null` will result in an uncaught error
- `Ref#focus()` is often called within a `setTimeout` (even in some example IIRC). `Ref` doesn't get GCd, but the DOM node it points to may no longer exist. Since React v16 is so blazing fast, it's possible that we're trying to focus something that no longer exists.  
- Without this bit of code, we'd have to call `findDOMNode` on the ref before calling `focus`. That's not great since putting `findDOMNode` is discouraged in application code.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -365,13 +365,13 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
     const alreadyHasFocus = editorState.getSelection().getHasFocus();
     const editorNode = ReactDOM.findDOMNode(this.refs.editor);
 
+    if (!(editorNode instanceof HTMLElement)) {
+      return;
+    }
+
     const scrollParent = Style.getScrollParent(editorNode);
     const {x, y} = scrollPosition || getScrollPosition(scrollParent);
 
-    invariant(
-      editorNode instanceof HTMLElement,
-      'editorNode is not an HTMLElement',
-    );
     editorNode.focus();
 
     // Restore scroll position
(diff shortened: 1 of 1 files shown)


flarnie:
Interesting - Seems like another option would be to release the timeout in `componentWillUnmount`, and that is generally what I would want to do.

That said, it looks like the main reason we added this invariant was to improve type strictness, and having an early return does the same thing effectively of validating the type before continuing execution.

I'm inclined to accept this, since I can't think of a big difference in effect between cleaning up the timeout on unmount and not throwing when it isn't cleaned up - but this way is more convenient for the engineer.

Adding this to my queue for merging - thanks @mattkrick !

sophiebits:
Maybe it would be cleaner to return if !editorNode then keep the invariant? Then we'll notice more explicitly if that's no longer true rather then it silently failing.

flarnie:
Oh - I like that! Will revise this and re-rebase/etc. tomorrow. :)

mattkrick:
i'm not sure i understand the change but as long as the function terminates without an error with the editorNode doesn't exist i'm happy!

src/component/base/DraftEditor.react.js
```diff
@@ -368,13 +368,19 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
...
     const editorNode = ReactDOM.findDOMNode(this.refs.editor);
 
-    if (!(editorNode instanceof HTMLElement)) {
+    if (!editorNode) {
+      // once in a while people call 'focus' in a setTimeout, and the node has
+      // been deleted, so it can be null in that case.
       return;
     }
 
     const scrollParent = Style.getScrollParent(editorNode);
```
mattkrick: correct me if I'm wrong, but Style.getScrollParent has 1 param & that is `node`, so we don't want to call it unless we have an `HTMLElement`, right? It doesn't matter either way (code below), but if we were writing a unit test for this, then it'd be REALLY hard to reach that invariant...

```
getScrollParent: function getScrollParent(node) {
	    if (!node) {
	      return null;
	    }
            ...
```

flarnie:
radish_spirit_dance

Issue #1411, opened 2017-09-28
Allow developing/testing Draft with React 16


flarnie:
Updating `devDependencies` to allow React 16 and running the tests leads to some failures. We should debug and fix this, so that we are developing Draft to run with the latest version of React.

**Edit:** Here is more detail about how to fix this, and if you'd like to pick it up please comment on the issue. :)

1. Update `devDependencies` in `package.json` to add `| ^16.0` to the current `react` and `react-dom` versions. *NOTE* that this is a different section from `dependencies` or `peerDependencies`.
2. Update any other dependencies (like Enzyme) which have released a recent version which accepts React 16 as a dependency.
3. Fix any failing tests, lints, flow errors resulting from the changes.

itssujan:
@flarnie I can take this up. Might be a good starter for me

Looks like Enzyme has a major upgrade from 2.x to 3.x and also uses an adapter `enzyme-adapter-react-16 ` for supporting React 16.

And also the ReactTestUtils seems to have been moved out to the ReactDOM package in one of the 2.x revisions.

flarnie:
Thanks @issujan!

itssujan:
https://github.com/facebook/draft-js/pull/1464

claudiopro:
This issue was apparently resolved with 5a965e6.

PR #1422, opened 2017-10-07
Support React 16

haydenbleasel:
PR for #1411

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
     "object-assign": "^4.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


flarnie:
Thanks! This is a great start. I think the `^16.0.0-rc` version should be compatible with `^16.0.0` but this won't hurt, and could make it even more clear to folks that we support with React 16.

For issue #1411 I would like `react` and `react-dom` updated in the `devDependencies`, which led to some errors last time I checked. Probably will also need to update to the latest Enzyme, and fix any breaking changes there. I'll update the issue description with those details.

PR #1428, opened 2017-10-12
Page scrolls to top when cutting text in block with scrollbar

### LINKED ISSUE #1427: Page scrolls to top when cutting text in block with scrollbar
To reproduce:
* Open the 'Rich' example in a browser.
* Open the developer console of the browser and add the following CSS rules:

```
pre.public-DraftStyleDefault-pre pre {
  overflow-x: auto;
}

pre.public-DraftStyleDefault-pre pre span {
  white-space: pre;
}
```

* Enter a lot of text in the editor, enough for the browser page to scroll.
* Add a code block at the end and type enough text in the block so that a horizontal scrollbar appears in the code block.
* Select some text in the code block and press the 'cut text' shortcut (Ctrl+X on my OS)

Result: the browser page scrolls all the way to the top.

I think this is a bug in `editOnCut.js`. It looks at the scroll parent of the event target to figure out how much the editor is scrolled, but it should look at the scroll parent of the editor itself (i.e., the same as in the `_focus()` function in `DraftEditor.react.js`).

robbertbrak:
This fixes #1427.


mitermayer:
Hi @robbertbrak ,

Thanks for this fix, would you mind rebasing this ?  that file has changed in master

robbertbrak:
Hi @mitermayer, I have rebased it.

src/component/handlers/edit/editOnCut.js
```diff
@@ -45,9 +44,9 @@ function editOnCut(editor: DraftEditor, e: SyntheticClipboardEvent<>): void {
 
   // Track the current scroll position so that it can be forced back in place
   // after the editor regains control of the DOM.
-  if (element instanceof Node) {
-    scrollPosition = getScrollPosition(Style.getScrollParent(element));
-  }
+  const editorNode = ReactDOM.findDOMNode(editor.editor);
```
mitermayer: Was wondering if the previous line `Style.getScrollParent(element)` was meant to always get the 'editor' dom reference ? Is this new line having equivalent behaviour to the deleted code ? 

cc @flarnie

robbertbrak: @mitermayer I think the answer is 'No' to both of your questions. The line `Style.getScrollParent(element)` was meant to get the _scroll parent_ of the editor dom reference, which in many cases will be the `document.body`.

The new line is _not_ equivalent to the deleted code. In many cases it is, because the scroll parent of the cut event target will also be the scroll parent of the editor dom node. But in some cases, as described in #1427, the scroll parent is something else, leading to erroneous behaviour. The point of my pull request is to fix this and make the behaviour equivalent to https://github.com/facebook/draft-js/blob/f50c4ca9a5881763eb3028ecacf8fd56f430bc3f/src/component/base/DraftEditor.react.js#L371.

juliankrispel:
@robbertbrak @mitermayer wonder how stuff like this could be tested. 

Thing is - whoever wrote the original code might have had a use-case that we're now breaking...

Would be nice to see a test for stuff like this.

PR #1449, opened 2017-10-24
0.10.4

flarnie:


CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ Notable changes to Draft.js will be documented in this file.
...
 Changes to `src` are live in production on facebook.com at the time of release.
 
+## 0.10.4 (October 24th, 2017)
+
+### Added
+* Expose `onRightArrow` and `onLeftArrow` props to allow handling keyboard
+  events when right or left arrow is pressed.
+  ([@eessex](https://github.com/eessex)
+  in [#1384](https://github.com/facebook/draft-js/pull/1384))
+* Expose path to Draft.css for usage by CSS pre-processors.
```
sophiebits: Can we say more like: "Expose Draft.css as default CSS export in package.json for use by CSS preprocessors"? "Expose path" doesn't quite capture it especially since the whole point is for users to *not* think about the path.

flarnie: Thanks for the review!

Sounds good - will fix.

PR #1467, opened 2017-10-27
Fixing FlowFixme errors

mitermayer:
This PR removes the $FlowFixMe and fixes the flow errors as well as minor things uncovered by the flow parser.

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -190,14 +190,9 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
   _renderPlaceholder(): ?React.Element<any> {
     if (this._showPlaceholder()) {
       return (
-        /* $FlowFixMe(>=0.53.0 site=www,mobile) This comment suppresses an
-         * error when upgrading Flow's support for React. Common errors found
-         * when upgrading Flow's React support are documented at
-         * https://fburl.com/eq7bs81w */
         <DraftEditorPlaceholder
           text={nullthrows(this.props.placeholder)}
           editorState={this.props.editorState}
-          textAlignment={this.props.textAlignment}
           accessibilityID={this._placeholderAccessibilityID}
         />
       );
@@ -286,10 +281,6 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
             style={contentStyle}
             suppressContentEditableWarning
             tabIndex={this.props.tabIndex}>
-            {/* $FlowFixMe(>=0.53.0 site=www,mobile) This comment suppresses an
-              * error when upgrading Flow's support for React. Common errors
-              * found when upgrading Flow's React support are documented at
-              * https://fburl.com/eq7bs81w */}
             <DraftEditorContents
               blockRenderMap={this.props.blockRenderMap}
               blockRendererFn={this.props.blockRendererFn}
--- a/src/component/base/DraftEditorPlaceholder.react.js
+++ b/src/component/base/DraftEditorPlaceholder.react.js
@@ -13,7 +13,6 @@
 
 'use strict';
 
-import type {DraftTextAlignment} from 'DraftTextAlignment';
 import type EditorState from 'EditorState';
 
 const React = require('React');
@@ -24,7 +23,6 @@ type Props = {
   accessibilityID: string,
   editorState: EditorState,
   text: string,
-  textAlignment: DraftTextAlignment,
 };
 
 /**
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -60,10 +60,10 @@ export type DraftEditorProps = {
   // For a given `ContentBlock` object, return an object that specifies
   // a custom block component and/or props. If no object is returned,
   // the default `DraftEditorBlock` is used.
-  blockRendererFn?: (block: ContentBlock) => ?Object,
+  blockRendererFn: (block: ContentBlock) => ?Object,
 
   // Function that returns a cx map corresponding to block-level styles.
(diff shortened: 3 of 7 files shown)


src/component/base/DraftEditorPlaceholder.react.js
```diff
@@ -25,7 +24,6 @@ type Props = {
   accessibilityID: string,
   editorState: EditorState,
   text: string,
-  textAlignment: DraftTextAlignment,
```
mitermayer: We never used the alignment property in this component. Is that a feature that folks would need ?

src/component/base/DraftEditor.react.js
```diff
@@ -300,10 +295,6 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
             tabIndex={this.props.tabIndex}>
             {}
```
mitermayer: This rogue '{}' is fixed on my other PR

src/component/base/DraftEditorPlaceholder.react.js
flarnie: Folks in open source may be using it - if we are deprecating we should warn through the 0.11.* series and then remove in 0.12.*, similar to how React handles deprecations.

src/component/base/DraftEditor.react.js
```diff
@@ -202,14 +202,9 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
...
     if (this._showPlaceholder()) {
       return (
-        /* $FlowFixMe(>=0.53.0 site=www,mobile) This comment suppresses an
-         * error when upgrading Flow's support for React. Common errors found
-         * when upgrading Flow's React support are documented at
-         * https://fburl.com/eq7bs81w */
         <DraftEditorPlaceholder
           text={nullthrows(this.props.placeholder)}
           editorState={this.props.editorState}
-          textAlignment={this.props.textAlignment}
```
flarnie: I think we still want to do this with `textAlignment` - and I did see uses of `textAlignment` internally at FB.

src/component/base/DraftEditorPlaceholder.react.js
flarnie: We did - see your previous fix. It should match the `textAlignment` passed in as a prop to `DraftEditor`.

src/component/base/DraftEditorProps.js
```diff
@@ -60,10 +60,10 @@ export type DraftEditorProps = {
   // For a given `ContentBlock` object, return an object that specifies
   // a custom block component and/or props. If no object is returned,
   // the default `DraftEditorBlock` is used.
-  blockRendererFn?: (block: ContentBlock) => ?Object,
+  blockRendererFn: (block: ContentBlock) => ?Object,
```
flarnie: What if a user doesn't pass this into Draft? And same question about the props below.

src/component/contents/DraftEditorBlock.react.js
```diff
@@ -43,15 +44,15 @@ type Props = {
   contentState: ContentState,
   block: ContentBlock,
   customStyleMap: Object,
-  customStyleFn: Function,
+  customStyleFn: (style: DraftInlineStyle, block: ContentBlock) => ?Object,
```
flarnie: :D Love the specific typing here.

src/component/contents/DraftEditorBlock.react.js
```diff
@@ -43,15 +44,15 @@ type Props = {
...
   tree: List<any>,
   selection: SelectionState,
-  decorator: DraftDecoratorType,
+  decorator: ?DraftDecoratorType,
   forceSelection: boolean,
   direction: BidiDirection,
   blockProps?: Object,
   startIndent?: boolean,
-  blockStyleFn: Function,
```
flarnie: Nice catch - looks like we now handle this in a parent component.

src/component/contents/DraftEditorLeaf.react.js
```diff
@@ -45,7 +45,7 @@ type Props = {
 
   // The current `SelectionState`, used to represent a selection range in the
   // editor
-  selection: SelectionState,
+  selection: ?SelectionState,
```
flarnie: Would it work to do `SelectionState | null`?

flarnie:
PS Thanks so much for this clean-up! Great to see this crufty Flow error stuff being washed away. 💦✨

src/component/base/DraftEditor.react.js
mitermayer: The reason why I remove was due to the fact that `DraftEditorPlaceholder` does not use the prop `textAlignment`.

If you refer to `https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorPlaceholder.react.js` you will see no usage of `textAlignment` apart from the props definition itself

But if you prefer I can updated this PR to still have to pass `textAlignment` and make it a nullable `?textAlignment` instead.

src/component/base/DraftEditorProps.js
mitermayer: The reason why we were able to delete it was that we already supply a default empty function for it https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditor.react.js#L73

src/component/contents/DraftEditorLeaf.react.js
mitermayer: We could do that, I have no strong opinion around it, we are currently doing a non strict check for it on https://github.com/facebook/draft-js/blob/master/src/component/contents/DraftEditorLeaf.react.js#L86

mitermayer:
Will import this internally to see if this flow type changes affect any internal project and do further amends from within phabricator

PR #1468, opened 2017-10-27
Refactoring function binding to use arrow functions

mitermayer:
Run a codemod to remove function binding in favour of arrow functions

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -160,20 +160,9 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
     this._onPaste = this._buildHandler('onPaste');
     this._onSelect = this._buildHandler('onSelect');
 
-    // Manual binding for public and internal methods.
-    this.focus = this._focus.bind(this);
-    this.blur = this._blur.bind(this);
-    this.setMode = this._setMode.bind(this);
-    this.exitCurrentMode = this._exitCurrentMode.bind(this);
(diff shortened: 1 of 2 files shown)


mitermayer:
Will submit a new commit to fix travis

src/component/base/DraftEditor.react.js
```diff
@@ -160,20 +160,9 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
...
-    this.setClipboard = this._setClipboard.bind(this);
-    this.getClipboard = this._getClipboard.bind(this);
     this.getEditorKey = () => this._editorKey;
-    this.update = this._update.bind(this);
-    this.onDragEnter = this._onDragEnter.bind(this);
-    this.onDragLeave = this._onDragLeave.bind(this);
 
-    // See `_restoreEditorDOM()`.
+    // See `restoreEditorDOM()`.
     this.state = {contentsKey: 0};
```
flarnie: As a follow-up let's add an underscore to any of the methods which are not supposed to be public - so anything except `focus` and `blur`. https://draftjs.org/docs/api-reference-editor.html#methods

You could make a `good-first-bug` type issue for this too if you are ready to jump into other stuff.

mitermayer: It was failling before due to me pushing a broken commit, the reason why it is no longer broken is because i have since fix it and  been pushing using `git --amend` and `git push --force` since nobody was reviewing it at that stage

Issue #1455, opened 2017-10-25
ContentBlock auto generate CharacterList


camurai:
**Do you want to request a *feature* or report a *bug*?**
Feature Request

**What is the current behavior?**
Currently when inserting a new ContentBlock into the ContentState, It appears that you need to generate the appropriate default CharacterList for the text set in the new block, otherwise the block text does not display. In order to work around this, I have used:
```js
convertFromRaw(convertToRaw(contentState))
```
as the convert back from Raw generates the CharacterList.

Here is a jsfiddle highlighting the problem. there is one line of code commented out that is my convertToRaw and back again work around:
https://jsfiddle.net/v7o1h76d/
Let me know if there is something else i should be doing to insert the block that would get around this problem.

**What is the expected behavior?**
Preferably the default CharacterList would be auto generated when the ContentBlock instance is first created as part of its constructor.

mitermayer:
Hi @camurai,

Thanks for this feature request. To give a bit more context, the way a ContentBlock is render it requires each character of the block text to have a matching CharacterMetadata.

On your above example you are providing an initial text to the ContentBlock creating without supplying a characterList. The default value for ContentBlock.characterList record is of Immutable.List(), so that means if you change your example to supply and empty string `''` instead of `"blargh"` it will work as expected.

If you want to have blocks created with the string `"blargh"` you would have to build the characterList object, that could be done by creating an Immutable.List of CharacterMetadata.EMPTY that matches same length as initial string.

so for the object would look like:

```
new ContentBlock({
  key: newBlockKey,
  type: 'unstyled',
  type: 'unstyled',
  text: 'blargh',
  characterList: Immutable.List([
    CharacterMetadata.create(),
    CharacterMetadata.create(),
    CharacterMetadata.create(),
    CharacterMetadata.create(),
    CharacterMetadata.create(),
    CharacterMetadata.create(),
  ])
})
```

I have updated the your fiddle to show the above behaviour https://jsfiddle.net/v7o1h76d/2/ 

*** 

I also agree with you, we could do a better job on making easier to create ContentBlock, by  defaulting characterList to a List of **CharacterMetadata.EMPTY** that matches the size of the **text** passed.

mitermayer:
Created a PR addressing this issue in here https://github.com/facebook/draft-js/pull/1472

PR #1472, opened 2017-10-30
Adding new static method ContentBlock.create

mitermayer:
Background:

Relying on the ContentBlock constructor for record creation means that the user would have to create a valid characterList matching the length of the text passed. It makes sense to simplify the API a bit by allowing the characterList to be optional.

This fixes: https://github.com/facebook/draft-js/issues/1455


flarnie:
First off - I think you are right that folks commonly do this. We have several examples of it internally.

That said - I remember you and I were just discussing how there are too many ways to do the same thing with Draft.js, and how our overall goal is to shrink the API.

Could we achieve this same operation with `Modifier.insertText`, or even `AtomicBlockUtils.insertAtomicBlock`? And if not, should we consider deprecating some of those helpers in favor of this? 

What do you think of having some guidelines about what we will consider for new APIs? We could discuss it at the next Draft sync.

mitermayer:
I agree with you 100%, will revisit this PR to figure out a long term solution for it.

mitermayer:
Hi @flarnie, 

Updated the logic to be inside of the constructor and removed the static `create` method.

PR #1477, opened 2017-11-01
Exploration - Feature tree / Setting up BlockNode interface

mitermayer:
### Context

**Setting up BlockNode interface**

This PR will be laying down the foundations by creating a tree block compatible node.

In order to avoiding including conditional logic as much as possible to existing draft library while exploring we will use interfaces to guarantee that the new tree block structure is some what compatible with the existing ContentBlock

***

**Note:**
production systems.

### DIFF
--- a/src/model/immutable/BlockNode.js
+++ b/src/model/immutable/BlockNode.js
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule BlockNode
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+import type CharacterMetadata from 'CharacterMetadata';
+import type {DraftBlockType} from 'DraftBlockType';
+import type {DraftInlineStyle} from 'DraftInlineStyle';
+import type {List, Map} from 'immutable';
+
+export type BlockNodeKey = string;
+
+export type BlockNodeConfig = {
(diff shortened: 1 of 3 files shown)


src/model/immutable/BlockNode.js
```diff
@@ -0,0 +1,60 @@
...
+
+  getData(): Map<any, any>;
+
+  getDepth(): number;
+
+  getEntityAt(offset: number): ?string;
+
+  getInlineStyleAt(offset: number): DraftInlineStyle;
+
+  getKey(): string;
```
flarnie: Should this return type `BlockNodeKey`?

src/model/immutable/ContentBlockNode.js
```diff
@@ -0,0 +1,160 @@
...
+const findRangesImmutable = require('findRangesImmutable');
+
+import type {BlockNode, BlockNodeConfig, BlockNodeKey} from 'BlockNode';
+import type {DraftBlockType} from 'DraftBlockType';
+import type {DraftInlineStyle} from 'DraftInlineStyle';
+
+import {List, Map, OrderedSet, Record, Repeat} from 'immutable';
+
+type ContentBlockNodeConfig = BlockNodeConfig & {
+  nodes?: List<BlockNodeKey>,
```
flarnie: `nodes` are the children, right? What do you think of `children` or `childNodes`?

src/model/immutable/ContentBlockNode.js
```diff
@@ -0,0 +1,160 @@
...
+  parent?: ?BlockNodeKey,
+  prevSiblling?: ?BlockNodeKey,
```
flarnie: nit - typos in spelling "sibling" above and below.

src/model/immutable/BlockNode.js
mitermayer: great catch

Issue #1453, opened 2017-10-25
Typescript DraftBlockType too strict?


camurai:
Its possible that I am missing an obvious way I am supposed to be doing this, but it seems like the DraftBlockType is too specific, making it impossible to add in custom block types in typescript. I was following the example found in the docs here:
https://draftjs.org/docs/advanced-topics-block-components.html#content 

Trying to use the code in the example in a typescript project, I got the following error, coming from the line where the Editor object is created in the render function:
```ts
 Type 'string' is not assignable to type '"unstyled" | "paragraph" | "header-one" | "header-two" | "header-three" | "header-four" | "header...'.
```
This seems to stem from the fact that blockRenderMap is expecting a Immutable.Map<DraftBlockType, any>, while anything I create with the merge will have to be Immutable.Map<string, any>

I've tried setting the types, but since DraftBlockType does not contain my custom block type, and blockRenderMap expects the key to be DraftBlockType, no custom Block types can be added.

To test things out I edited the type for DraftBlockType in the node-module and set it to just be "string" instead of "unstyled" | "paragraph" | ....
This seems to make it so I am able to compile without the error messages.

So is there a different way to create a Custom Block Component that gets around the DraftBlockType, or was DraftBlockType just made too strict when the type file was setup?

mitermayer:
Hi @camurai,

Thanks for reporting this issue

mitermayer:
Hi @xFloooo,

`DefinitelyTyped` is a community driven repository, you should submit an issue/PR to that repo for requesting this changes.

PR #1480, opened 2017-11-01
Making DraftBlockType less restrictive

mitermayer:
DraftBlockType needs to be less strict in order to allow user defined block types however we also want to retain a way to identify our core draft blocks.

fixes: https://github.com/facebook/draft-js/issues/1453

### DIFF
--- a/src/model/immutable/ContentBlock.js
+++ b/src/model/immutable/ContentBlock.js
@@ -26,7 +26,7 @@ const {List, Map, OrderedSet, Record, Repeat} = Immutable;
 const EMPTY_SET = OrderedSet();
 
 type ContentBlockConfig = {
-  key?: string,
+  key?: DraftBlockType,
   type?: string,
   text?: string,
   characterList?: List<CharacterMetadata>,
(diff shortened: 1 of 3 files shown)


src/model/immutable/ContentBlock.js
```diff
@@ -26,7 +26,7 @@ const {List, Map, OrderedSet, Record, Repeat} = Immutable;
 const EMPTY_SET = OrderedSet();
 
 type ContentBlockConfig = {
-  key?: string,
+  key?: DraftBlockType,
```
mitermayer: Just adding back, had removed this temporarily while shipping the other PR internally, but now since DraftBlockType is more permissive we wont run in the same problem again

flarnie:
lgtm - thanks! :)

src/model/immutable/ContentBlock.js
mitermayer: My mistake in here,  this should have been added to `type` not `key` but fixing it on a follow up PR https://github.com/facebook/draft-js/pull/1485

PR #1485, opened 2017-11-02
Fixing mixed up typing

mitermayer:
While amending diff in phabricator I got typing mixed up, this PR fixes it


flarnie:
Thanks for fixing!

dogge_thanks

PR #1486, opened 2017-11-03
Added documentation for DraftEntityMutability

trevtrich:
**Summary**
I often come across docs that show the raw types for a method signature (e.g., `Entity.create` has `DraftEntityMutability` as its second arg). Seeing this in raw text with no link to the true options that are available can be frustrating to a new user. They're essentially left to dive into the code on github to figure it out.

Would it be useful for me to start building some of these out like I've done in this PR? My apologies if there is some 'documentation guidelines' or decisions that have been made on what or what not to document. My primary goal here is a conversation rather than necessarily this specific PR's content.

**Test Plan**
n/a

### DIFF
--- a/docs/APIReference-ContentState.md
+++ b/docs/APIReference-ContentState.md
@@ -336,7 +336,7 @@ Returns whether the contents contain any text at all.
 ```
 createEntity(
   type: DraftEntityType,
-  mutability: DraftEntityMutability,
+  mutability: [DraftEntityMutability](/docs/advanced-topics-entities.html#mutability),
   data?: Object
 ): ContentState
 ```
--- a/docs/APIReference-Entity.md
+++ b/docs/APIReference-Entity.md
@@ -62,7 +62,7 @@ be used only for retrieval.
 ```
 create(
   type: DraftEntityType,
-  mutability: DraftEntityMutability,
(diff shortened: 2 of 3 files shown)


mitermayer:
Hi @richardson-trevor,

Thank you for this!

@flarnie wold this doc still be applicable with the entity creation API being deprecated ?

flarnie:
This info is still accurate - we would just need to change `DraftEntity.create` to `contentState.createEntity`.

This is a good thing to document - I see a section covering this in the `Entities` page of the docs, here - https://draftjs.org/docs/advanced-topics-entities.html#mutability

What would you think of adding info to the existing mutability section, and linking to that from the places that mention `DraftEntityMutability` in the docs?

richardson-trevor:
That seems totally fair. When I get some time I'll get this updated. Thanks for the feedback 👍 .

richardson-trevor:
Alright, finally got around to taking a crack at this update. Let me know what you think. I wasn't able to run the site locally to test my doc changes (currently requesting some input from folks on our Slack but haven't heard back yet). Probably my unfamiliarity and nothing wrong with the local server.

niveditc:
@richardson-trevor - just wanted to ping this PR. We'd still love to merge it in once you get a chance to make the changes :)

PR #1491, opened 2017-11-06
Fix syntax nit for prettier in BlockType

flarnie:
Just fixing a lint errors.

### DIFF
--- a/src/model/immutable/BlockNode.js
+++ b/src/model/immutable/BlockNode.js
@@ -33,28 +33,28 @@ export interface BlockNode {
   findEntityRanges(
     filterFn: (value: CharacterMetadata) => boolean,
     callback: (start: number, end: number) => void,
-  ): void,
+  ): void;
 
   findStyleRanges(
     filterFn: (value: CharacterMetadata) => boolean,
     callback: (start: number, end: number) => void,
-  ): void,
+  ): void;
 
-  getCharacterList(): List<CharacterMetadata>,
+  getCharacterList(): List<CharacterMetadata>;
 
-  getData(): Map<any, any>,
+  getData(): Map<any, any>;
 
-  getDepth(): number,
+  getDepth(): number;
 
-  getEntityAt(offset: number): ?string,
+  getEntityAt(offset: number): ?string;
 
-  getInlineStyleAt(offset: number): DraftInlineStyle,
+  getInlineStyleAt(offset: number): DraftInlineStyle;
 
-  getKey(): BlockNodeKey,
+  getKey(): BlockNodeKey;
 
-  getLength(): number,
+  getLength(): number;
 
-  getText(): string,
+  getText(): string;
 
-  getType(): DraftBlockType,
+  getType(): DraftBlockType;
 }


flarnie:
It appears we prefer commas in our internal `Prettier` config, and semicolons in the Draft.js `Prettier` config.

Looking into that is the real fix for this. If you get a chance @mitermayer go ahead, otherwise I'll fix it after I finish reviewing your Jest Snapshot PR.

mitermayer:
Just quoting @vjeux  in here for history tracking (https://github.com/facebook/draft-js/issues/1492#issuecomment-342213957)

> At flow's team request, prettier now uses ; instead of , but I haven't updated www yet to the version that contains this change.

We can solve this by doing one of two things

* downgrade draft prettier to match facebook ones 

or: 

* ignore internal linting since eventually it will be fixed when @vjeux  upgrade prettier internally

Any thoughts around it ?

mitermayer:
Just talked with @vjeux, the main take a way is, lets merge this PR to fix travis, and while we are importing draft internally we can ignore the 'prettier linting' rules caused by our internal, since in the next couple weeks prettier will be bumped internally.

One thing that I was thinking is if would make sense to eventually move prettier settings and configurations to 'fbjs' so that all our open source projects use it and are on sync with 'www' future prettier versions ? 

cc @zpao

mitermayer:
Will update this PR to keep fb internal formatting and add this to fix the CI:

```
// https://github.com/facebook/draft-js/issues/1492
// prettier-ignore
```

The reason for is that our internal linting rules ignores // prettier-ignore

zpao:
eslint-plugin-prettier has a config that matches FB, maybe we can piggyback on that? (https://github.com/prettier/eslint-plugin-prettier#options)

That said, I don't think this change was configurable, just an incompatability between versions. You should probably ensure you're using the same version of prettier here that is being used in www. Currently www is using ^1.6.1 and you have ^1.7.4 here, so it's not surprising to see discrepencies. Your eslint-plugin-prettier is also newer here. I would recommend you start using yarn in the repo here and lock that at the www version as well (though it should matter less since it mostly just runs prettier).

mitermayer:
Hi @zpao,

Thanks for the suggestions.

> You should probably ensure you're using the same version of prettier here that is being used in www

Would be great if this could be handled by `fbjs` this way would always be on sync with www. Is it possible ? I can imagine that other fb open source projects could benefit from this as well

zpao:
Is it possible? Not in a great sustainable way. Could maybe hardcode it into something like fbjs's check dependencies script but if you fail to update that then all bets are off. Or maybe have a file somewhere on the internet that we (hopefully remember to) update, but if you're building on plane, the build will fail. So you'll have some manual step. I'm not sure if it's worth the effort in all honesty - Draft is the only project actively syncing out of www (couple other projects are not actively worked on/synced). When internal prettier upgrades happen, there's usually a post somewhere so when that happens you can update the version here. Or just wait for builds to fail on master and if they're formatting failures, that's the clue to upgrade your version here to match (in other words, exactly what happened but don't edit your code, edit your package.json/yarn.lock).

PR #1493, opened 2017-11-06
Exploration - Feature tree / Wiring up BlockNode interface

mitermayer:
### Context

**Wiring up BlockNode interface**

This PR will make sure that ContentBlock references are now referencing the BlockNode interface in order for us to be able to support both ContentBlock and ContentBlockNode operations

***

**Note:**
production systems.

### DIFF
--- a/src/model/immutable/BlockNode.js
+++ b/src/model/immutable/BlockNode.js
@@ -16,7 +16,7 @@
 import type CharacterMetadata from 'CharacterMetadata';
 import type {DraftBlockType} from 'DraftBlockType';
 import type {DraftInlineStyle} from 'DraftInlineStyle';
-import type {List, Map} from 'immutable';
+import type {List, Map, Record} from 'immutable';
 
 export type BlockNodeKey = string;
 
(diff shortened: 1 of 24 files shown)


juliankrispel:
Omg yes please! If I can help at all somehow please lmk @mitermayer

src/model/immutable/BlockNode.js
```diff
@@ -31,7 +31,7 @@ export type BlockNodeConfig = {
 
 // https://github.com/facebook/draft-js/issues/1492
 // prettier-ignore
-export interface BlockNode {
+export interface BlockNode extends Record {
```
flarnie: I'm missing something here - why do we no longer extend Record?
Would that make BlockNode incompatible with the ContentBlock type?

mitermayer: This is becase we have some places in the codebase where we do things like this:

```
src/model/transaction/getContentStateFragment.js
75:    return block.set('key', newKey);

src/model/transaction/adjustBlockDepthForContentState.js
36:      return block.set('depth', depth);

src/model/transaction/applyEntityToContentBlock.js
34:  return contentBlock.set('characterList', characterList);

src/model/transaction/ContentStateInlineStyle.js
81:      return block.set('characterList', chars);

src/model/transaction/removeEntitiesAtEdges.js
112:      return block.set('characterList', chars);
```

TLDR: 
We can eventually clean this up later and make sure we are not setting it directly. But due to the fact that we have places that we are using 'blocks' assuming the immutable Record API we need to make sure we extend its type

PR #1494, opened 2017-11-07
Exploration - Feature tree / Serializing Draft state to tree data structure

mitermayer:
### Context

**Serializing Draft state to tree data structure**

This PR will make sure that we can serialize existing draft state into a tree data block structure for ContentBlockNode and flat data structure for the existing ContentBlock

***

**Note:**
production systems.

### DIFF
--- a/src/model/encoding/__tests__/encodeInlineStyleRanges-test.js
+++ b/src/model/encoding/__tests__/encodeInlineStyleRanges-test.js
@@ -91,13 +91,13 @@ describe('encodeInlineStyleRanges', () => {
 
   it('must encode custom styles', () => {
     const custom = List([FOO, FOO, FOO_BAR, FOO_BAR, BOLD, BOLD]);
-    expect(
-      encodeInlineStyleRanges(createBlock(' '.repeat(6), custom)),
-    ).toEqual([
-      {offset: 0, length: 4, style: 'foo'},
-      {offset: 2, length: 2, style: 'bar'},
-      {offset: 4, length: 2, style: 'BOLD'},
-    ]);
+    expect(encodeInlineStyleRanges(createBlock(' '.repeat(6), custom))).toEqual(
+      [
+        {offset: 0, length: 4, style: 'foo'},
+        {offset: 2, length: 2, style: 'bar'},
+        {offset: 4, length: 2, style: 'BOLD'},
+      ],
+    );
   });
 
   it('must encode for a complex styled document', () => {
--- a/src/model/encoding/convertFromDraftStateToRaw.js
+++ b/src/model/encoding/convertFromDraftStateToRaw.js
@@ -16,24 +16,36 @@
 import type ContentState from 'ContentState';
 import type {RawDraftContentState} from 'RawDraftContentState';
 
-var DraftStringKey = require('DraftStringKey');
+const DraftStringKey = require('DraftStringKey');
 
-var encodeEntityRanges = require('encodeEntityRanges');
-var encodeInlineStyleRanges = require('encodeInlineStyleRanges');
+const encodeEntityRanges = require('encodeEntityRanges');
+const encodeInlineStyleRanges = require('encodeInlineStyleRanges');
 
-function convertFromDraftStateToRaw(
+const encodeBlock = (block, entityStorageMap) => {
+  return {
+    key: block.getKey(),
+    text: block.getText(),
+    type: block.getType(),
+    depth: block.getDepth(),
+    inlineStyleRanges: encodeInlineStyleRanges(block),
+    entityRanges: encodeEntityRanges(block, entityStorageMap),
+    data: block.getData().toObject(),
+  };
+};
+
+const convertFromDraftStateToRaw = (
   contentState: ContentState,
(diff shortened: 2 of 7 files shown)


src/model/immutable/BlockNode.js
```diff
@@ -29,6 +29,8 @@ export type BlockNodeConfig = {
   type?: DraftBlockType,
 };
 
+// https://github.com/facebook/draft-js/issues/1492
```
mitermayer: Not sure why github is showing this as changes, this is already in master https://github.com/facebook/draft-js/blob/master/src/model/immutable/BlockNode.js#L32 and have been merged on https://github.com/facebook/draft-js/commit/652afc1a073702c351ad5be107c3febdcfe9936e

src/model/encoding/__tests__/encodeInlineStyleRanges-test.js
```diff
@@ -91,13 +91,13 @@ describe('encodeInlineStyleRanges', () => {
...
   it('must encode custom styles', () => {
     const custom = List([FOO, FOO, FOO_BAR, FOO_BAR, BOLD, BOLD]);
-    expect(
-      encodeInlineStyleRanges(createBlock(' '.repeat(6), custom)),
-    ).toEqual([
-      {offset: 0, length: 4, style: 'foo'},
-      {offset: 2, length: 2, style: 'bar'},
-      {offset: 4, length: 2, style: 'BOLD'},
-    ]);
+    expect(encodeInlineStyleRanges(createBlock(' '.repeat(6), custom))).toEqual(
```
mitermayer: This was due to a new prettier been version been released, and was failing the ci. At least was only this that needed change!

After @vjeux  release a new version of prettier in www we should prob then pin draft.js to that version with a yarn.lock.

src/model/encoding/convertFromDraftStateToRaw.js
```diff
@@ -16,60 +16,144 @@
...
+  const parentKey = block.getParentKey();
+  const rawBlock = (blockCacheRef[block.getKey()] = {
+    ...createRawBlock(block, entityMap),
+    children: [],
+  });
+
+  if (parentKey) {
+    blockCacheRef[parentKey].children.push(rawBlock);
+    return;
+  }
```
flarnie: I like that!

src/model/encoding/convertFromDraftStateToRaw.js
```diff
@@ -16,60 +16,144 @@
...
+  let rawDraftContentState = {
+    entityMap: {},
+    blocks: [],
   };
-}
+
+  // add blocks
+  rawDraftContentState = encodeRawBlocks(contentState, rawDraftContentState);
+
+  // add entity
```
flarnie: nit - 'entity' -> 'entities'

src/model/transaction/getSampleStateForTesting.js
```diff
@@ -13,19 +13,19 @@
...
+const ContentState = require('ContentState');
+const EditorState = require('EditorState');
+const Immutable = require('immutable');
+const SampleDraftInlineStyle = require('SampleDraftInlineStyle');
+const SelectionState = require('SelectionState');
 
-var {BOLD, ITALIC} = SampleDraftInlineStyle;
-var ENTITY_KEY = '123';
+const {BOLD, ITALIC} = SampleDraftInlineStyle;
+const ENTITY_KEY = '1';
```
flarnie: Just  curious - why change this to 1?

flarnie:
FYI I also pinged @SirensOfTitan to keep him in the loop and get more eyes on this.

src/model/encoding/convertFromDraftStateToRaw.js
```diff
@@ -16,60 +16,144 @@
 import type ContentState from 'ContentState';
```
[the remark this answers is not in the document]
mitermayer: You are correct, we will be able to revisit this later once we fully complete the migration

src/model/transaction/getSampleStateForTesting.js
mitermayer: TLDR:

Entitymap is indexed based on insertions (1,2,3) when processed on the convertStateToRaw, having the previous `123` as key meant we would not retrieve the key correctly from the entitymap

mitermayer: for reference:

https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromDraftStateToRaw.js#L27

```
function convertFromDraftStateToRaw(
  contentState: ContentState,
): RawDraftContentState {
  var entityStorageKey = 0;
```

and https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromDraftStateToRaw.js#L56


```
  // Flip storage map so that our storage keys map to global
  // DraftEntity keys.
  var entityKeys = Object.keys(entityStorageMap);
  var flippedStorageMap = {};
  entityKeys.forEach((key, jj) => {
    var entity = contentState.getEntity(DraftStringKey.unstringify(key));
    flippedStorageMap[jj] = {
      type: entity.getType(),
```

PR #1508, opened 2017-11-13
Exploration - Feature tree / Rendering tree data format

mitermayer:
### Context

**Rendering tree data format**

This PR adds support for react components to render the tree data structure

***

**Note:**
production systems.


src/component/contents/DraftEditorContents.react.js
```diff
@@ -117,109 +143,144 @@ class DraftEditorContents extends React.Component<Props> {
...
-      if (Element === 'li') {
-        const shouldResetCount =
-          lastWrapperTemplate !== wrapperTemplate ||
-          currentDepth === null ||
-          depth > currentDepth;
-        className = joinClasses(
-          className,
-          getListItemClasses(blockType, depth, shouldResetCount, direction),
-        );
+    if (rootBlock && rootBlock instanceof ContentBlockNode) {
```
mitermayer: For now left the rendering logic for tree inside `DraftEditorContents.react.js`, but when we introduce a new feature flag to control this logic we can move this out to a tree aware variant of `DraftEditorContents.react.js` and use the feature flag to chose between both.

Decided to integrate in here for now to make it easier to review and get early end to end progression

src/component/contents/exploration/__tests__/DraftEditorContents.react-test.js
```diff
@@ -0,0 +1,142 @@
+/**
```
mitermayer: We have a file called `DraftEditorContent.react-test.js` however since I plan to move the logic for rendering tree outside of that file it made sense to create a separate test file as opposed to add to the existing one.

src/component/contents/DraftEditorContents.react.js
```diff
@@ -117,109 +143,144 @@ class DraftEditorContents extends React.Component<Props> {
...
+          forceSelection,
+          selection,
+          block: nodeBlock,
+          direction: textDirectionality
+            ? textDirectionality
+            : directionMap.get(blockKey),
+          tree: editorState.getBlockTree(blockKey),
+        };
+
+        const configForType =
```
mitermayer: Due to DraftEditorBlockNode only handle wrapping for its own children had to still rely on the augment hack for wraping the rootBlock nodes. 

Once we introduce featutre control flag move this logic out of this file we can then avoid this

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+): boolean => {
+  const nextSiblingKey = block.getNextSiblingKey();
+
+  return nextSiblingKey
+    ? contentState.getBlockForKey(nextSiblingKey).getType() === block.getType()
+    : false;
+};
+
+const applyWrapperElementToSiblings = (
+  wrapperTemplate: *,
```
flarnie: Could the type here be `function |  React.ComponentType<any>`?

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+    ),
+  );
+
+  return nodes;
+};
+
+const getDraftRenderConfig = (
+  block: BlockNodeRecord,
+  blockRenderMap: DraftBlockRenderMap,
+): * => {
```
flarnie: We could at least say it returns type `Object` and leave a TODO to add a specific type. Same below with `getCustomRenderConfig`.

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+    customEditable,
+  };
+};
+
+const getElementPropsConfig = (
+  block: BlockNodeRecord,
+  editorKey: string,
+  offsetKey: string,
+  blockStyleFn: BlockStyleFn,
+  customConfig: *,
```
flarnie: if this `customConfig` param is the same as the return value of `getCustomRenderConfig` we could add a specific type and reuse it. Could be done in a follow-up commit.

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+
+  return elementProps;
+};
+
+class DraftEditorBlockNode extends React.Component<Props> {
+  shouldComponentUpdate(nextProps: Props): boolean {
+    const {block, direction, tree} = this.props;
+
+    return (
+      block.getChildKeys() !== List() ||
```
flarnie: Just noting that we talked about updating this conditional.

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+        const elementProps = getElementPropsConfig(
+          child,
+          editorKey,
+          offsetKey,
+          blockStyleFn,
+          customConfig,
+        );
+        const childProps = {
+          ...this.props,
+          tree: editorState.getBlockTree(key),
```
flarnie: I hate to pass the editorState through in order to use it just once. Not sure how to extract what we need here though.

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+        }
+
+        // if we are here it means we are the last block
+        // that has a wrapperTemplate so we should wrap itself
+        // and all other previous siblings that share the same wrapper
+        applyWrapperElementToSiblings(wrapperTemplate, Element, acc);
+
+        return acc;
+      }, []);
+    }
```
flarnie: This might be nice as a helper function, like `this._renderChildren`.

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+const applyWrapperElementToSiblings = (
+  wrapperTemplate: *,
+  Element: string,
+  nodes: Array<React.Node>,
+): Array<React.Node> => {
+  const wrappedSiblings = [];
+
+  // we check back until we find a sibbling that does not have same wrapper
+  for (const sibling: any of nodes.reverse()) {
+    if (sibling.type !== Element) {
```
SirensOfTitan: Why are we tagging this block as `any`?  React.Node could be null or a number, amongst other things, so we should insulate against that?

+nit, but since this is so simple why don't we avoid a break?
```
for (const sibling of nodes.reverse()) {
  if (sibling && sibling.type === Element) {
    wrappedSiblings.push(sibling);
  }
}
// or this could be reduced down to a simple filter:
const wrappedSiblings = nodes.reverse().filter(node ==> node && node.type === Element);
```

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+};
+
+const getElementPropsConfig = (
+  block: BlockNodeRecord,
+  editorKey: string,
+  offsetKey: string,
+  blockStyleFn: BlockStyleFn,
+  customConfig: *,
+): Object => {
+  let elementProps: Object = {
```
SirensOfTitan: Again, I don't know what flow internals are around props, but using a CovariantObject is safer, so perhaps getElementPropsConfig should return the type {+[key: string]: any};

src/component/contents/exploration/DraftEditorBlockNode.react.js
SirensOfTitan: `!block.getChildKeys.isEmpty()` as perhaps something more readable?

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+class DraftEditorBlockNode extends React.Component<Props> {
+  shouldComponentUpdate(nextProps: Props): boolean {
+    const {block, direction, tree} = this.props;
+
+    return (
+      block.getChildKeys() !== List() ||
+      block !== nextProps.block ||
+      tree !== nextProps.tree ||
+      direction !== nextProps.direction ||
+      (isBlockOnSelectionEdge(nextProps.selection, nextProps.block.getKey()) &&
```
SirensOfTitan: This line perhaps deserves a comment.  So why would we update if we're on a selection edge and want to forceSelection?

src/component/contents/exploration/DraftEditorDecoratedLeaves.react.js
```diff
@@ -0,0 +1,91 @@
...
+      contentState,
+      decorator,
+      decoratorKey,
+      direction,
+      leafSet,
+      text,
+    } = this.props;
+
+    const blockKey = block.getKey();
+    const leavesForLeafSet = leafSet.get('leaves');
```
SirensOfTitan: Perhaps I'm missing something here, but leafSet is a List, which means get takes numbers.  Is `leafSet` typed correctly?

src/component/contents/DraftEditorBlock.react.js
```diff
@@ -16,6 +16,7 @@
 import type {BlockNodeRecord} from 'BlockNodeRecord';
```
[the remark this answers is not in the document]
mitermayer: Thats correct! only cosmetic chanes

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+const getElementPosition = require('getElementPosition');
+const getScrollPosition = require('getScrollPosition');
+const getViewportDimensions = require('getViewportDimensions');
+const invariant = require('invariant');
+
+const SCROLL_BUFFER = 10;
+
+const {List} = Immutable;
+
+type BlockRenderFn = (block: BlockNodeRecord) => ?Object;
```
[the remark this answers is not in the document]
mitermayer: I got this from the main file https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorProps.js#L183

I suggest we update all of that later, could even raise this as a good first bug, updating the flow definitions

src/component/contents/exploration/DraftEditorBlockNode.react.js
```diff
@@ -0,0 +1,370 @@
...
+  customStyleFn: (style: DraftInlineStyle, block: BlockNodeRecord) => ?Object,
+  customStyleMap: Object,
+  decorator: ?DraftDecoratorType,
+  direction: BidiDirection,
+  editorKey: string,
+  editorState: EditorState,
+  forceSelection: boolean,
+  selection: SelectionState,
+  startIndent?: boolean,
+  tree: List<any>,
```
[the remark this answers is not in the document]
mitermayer: I agree,  do you have some suggestions ?

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: could this also be null ?

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: Sounds good to me, I was leaning towards adding typing now but the PR was already too big, happy to do this on follow up PR's

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: That's correct this conditional logic as it is is wrong, will update it

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: I could pass just the `editorState.getBlockTree` method reference, what do you reckon ?

src/component/contents/DraftEditorBlock.react.js
```diff
@@ -39,18 +40,29 @@ const nullthrows = require('nullthrows');
 const SCROLL_BUFFER = 10;
 
 type Props = {
-  contentState: ContentState,
   block: BlockNodeRecord,
+  blockProps?: Object,
```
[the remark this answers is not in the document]
mitermayer: This is already existent draft editor block, we definitely should improve typing but this should probably come up on a different PR

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: We want to avoid unnecessary iterations on larger documents

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: This is legacy from the port of the other BlockNode

src/component/contents/exploration/DraftEditorDecoratedLeaves.react.js
mitermayer: Will double check, could be that it's typed incorrectly

src/component/contents/DraftEditorBlock.react.js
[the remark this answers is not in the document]
flarnie: I agree that we could improve the types - seems like a good thing to work on as a follow up. We can make a list and either bootcamp or share with @juliankrispel .

We have run into some type definitions that pass CI on Github but break internally, so we are also being careful so we don't get blocked by that.

src/component/contents/exploration/DraftEditorBlockNode.react.js
mitermayer: Been trying to play safe on this PR since I don't want to get blocked by flow types, a lot of times flow internal is behaving different than here so I think we can create a follow up PR / issue to harden up draft.js typings in general

src/component/contents/exploration/DraftEditorBlockNode.react.js
flarnie: Sounds good - passing a bound function like that is a bit unconventional, but I can't think of any strong reason not to.

PR #1518, opened 2017-11-17
Exploration - Feature tree / Convert from html tree data support

mitermayer:
### Context

**Convert from html tree data support**

This PR adds support for converting from HTML to tree data, as a follow up we will then hook the paste processor with this too

***

**Note:**
production systems.


src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -249,15 +220,19 @@ function processInlineTag(
...
+): Chunk => {
   // Sometimes two blocks will touch in the DOM and we need to strip the
   // extra delimiter to preserve niceness.
-  var lastInA = A.text.slice(-1);
-  var firstInB = B.text.slice(0, 1);
+  const lastInA = A.text.slice(-1);
+  const firstInB = B.text.slice(0, 1);
 
-  if (lastInA === '\r' && firstInB === '\r') {
+  if (lastInA === '\r' && firstInB === '\r' && !experimentalHasNestedBlocks) {
```
mitermayer: This is the area where we updated in the `joinChunks` section. This is responsible for previously flattning blocks making nested blocks content collapse to root container elemenets. Since we want to support nested blocks we don't want this to happen when the flag is enabled

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -281,32 +256,77 @@ function joinChunks(A: Chunk, B: Chunk): Chunk {
...
+  parentKey: ?string = null,
+): Chunk => {
+  return {
+    text: '\r',
+    inlines: [OrderedSet()],
+    entities: new Array(1),
+    blocks: [
+      getChunkedBlock({
+        parent: parentKey,
+        key: generateRandomKey(),
```
mitermayer: The change in here is that when we create a `block chunk` we will always want a key since we will use that latter to build the tree data blocks.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -353,12 +380,11 @@ function genFragment(
...
   if (nodeName === 'br') {
-    if (
-      lastLastBlock === 'br' &&
-      (!inBlock ||
-        getBlockTypeForTag(inBlock, lastList, blockRenderMap) === 'unstyled')
-    ) {
-      return {chunk: getBlockDividerChunk('unstyled', depth), entityMap};
+    if (lastLastBlock === 'br' && (!inBlock || inBlockType === 'unstyled')) {
+      return {
+        chunk: getBlockDividerChunk('unstyled', depth, parentKey),
```
mitermayer: Only difference here is that we pass the `parentKey`, this is always undefined for root containers but since we call this function recursively when applying it for child-nodes it will be populated with the parent block key

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -404,32 +427,37 @@ function genFragment(
...
+
   // Block Tags
-  if (!inBlock && blockTags.indexOf(nodeName) !== -1) {
-    chunk = getBlockDividerChunk(
-      getBlockTypeForTag(nodeName, lastList, blockRenderMap),
-      depth,
-    );
+  if (lastList && inBlock === 'li' && nodeName === 'li') {
+    chunk = getBlockDividerChunk(blockType, depth, parentKey);
+    blockKey = chunk.blocks[0].key;
```
mitermayer: We get a reference of the generated `blockKey` so that wen we recurse we can pass it as the `parentKey` for the next block

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -404,32 +427,37 @@ function genFragment(
...
     inBlock = nodeName;
-    newBlock = true;
-  } else if (lastList && inBlock === 'li' && nodeName === 'li') {
+    nextBlockType =
+      lastList === 'ul' ? 'unordered-list-item' : 'ordered-list-item';
+    newBlock = !experimentalTreeDataSupport;
```
mitermayer: this flag `newBlock` was previously used to split nested blocks, when we have tree support enabled we dont want this to be active

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -404,32 +427,37 @@ function genFragment(
...
       getBlockTypeForTag(nodeName, lastList, blockRenderMap),
       depth,
+      parentKey,
     );
+    blockKey = chunk.blocks[0].key;
     inBlock = nodeName;
-    newBlock = true;
-    nextBlockType =
-      lastList === 'ul' ? 'unordered-list-item' : 'ordered-list-item';
+    newBlock = !experimentalTreeDataSupport;
```
mitermayer: This is the same as the above, except that the above one create list blocks and this create all the other blocks

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -467,16 +495,18 @@ function genFragment(
       depth,
       blockRenderMap,
       entityId || inEntity,
+      experimentalTreeDataSupport,
+      blockKey,
```
mitermayer: we pass the bockKey as the `parentKey` prop of this recursive function

src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -54,7 +43,83 @@ const testConvertingAdjacentHtmlElementsToContentBlocks = (tag: string) => {
...
+      experimentalTreeDataSupport,
+    ).contentBlocks.map(block => block.toJS()),
+  ).toMatchSnapshot();
+};
+
+const testConvertingAdjacentHtmlElementsToContentBlocks = (
+  tag: string,
+  experimentalTreeDataSupport?: boolean = false,
+) => {
+  test(`must not merge tags when converting adjacent <${tag} />`, () => {
```
flarnie: I love this. Sometimes it's better to be repetetive in tests, but I think this is DRY while still being really clear.

src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -69,3 +134,43 @@ test('img with data protocol should be correctly parsed', () => {
...
+  assertConvertFromHTMLToContentBlocks(html_string, {
+    experimentalTreeDataSupport: true,
+  });
+});
+
+SUPPORTED_TAGS.forEach(tag =>
+  testConvertingAdjacentHtmlElementsToContentBlocks(tag, true),
+);
+
+// assert that using tree blocks wtesth
```
flarnie: nit - a typo in the above comment.

src/model/encoding/__tests__/__snapshots__/convertFromHTMLToContentBlocks-test.js.snap
```diff
@@ -1,10 +1,568 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`converts deeply nested html blocks when experimentalTreeDataSupport is enabled 1`] = `
```
flarnie: Thanks for improving the test coverage here so much!!!

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -404,32 +427,37 @@ function genFragment(
...
-  } else if (lastList && inBlock === 'li' && nodeName === 'li') {
+    nextBlockType =
+      lastList === 'ul' ? 'unordered-list-item' : 'ordered-list-item';
+    newBlock = !experimentalTreeDataSupport;
+  } else if (
+    (!inBlock || experimentalTreeDataSupport) &&
+    blockTags.indexOf(nodeName) !== -1
+  ) {
     chunk = getBlockDividerChunk(
       getBlockTypeForTag(nodeName, lastList, blockRenderMap),
```
flarnie: Nit - we can now use the variable `blockType` here instead of `getBlockTypeForTag(nodeName, lastList, blockRenderMap)`.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -404,32 +427,37 @@ function genFragment(
...
+  const blockType = getBlockTypeForTag(nodeName, lastList, blockRenderMap);
+
   // Block Tags
-  if (!inBlock && blockTags.indexOf(nodeName) !== -1) {
-    chunk = getBlockDividerChunk(
-      getBlockTypeForTag(nodeName, lastList, blockRenderMap),
-      depth,
-    );
+  if (lastList && inBlock === 'li' && nodeName === 'li') {
+    chunk = getBlockDividerChunk(blockType, depth, parentKey);
```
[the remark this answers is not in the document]
flarnie: Ok - I see now that you switched the 'list hack' conditional to be first and the other conditional to be second.

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -281,32 +256,77 @@ function joinChunks(A: Chunk, B: Chunk): Chunk {
...
-  var protocol = link.protocol;
+  const protocol = link.protocol;
   return (
     protocol === 'http:' || protocol === 'https:' || protocol === 'mailto:'
   );
-}
+};
+
+const getWhitespaceChunk = (inEntity: ?string): Chunk => {
+  return {
```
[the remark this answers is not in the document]
mitermayer: That's correct!

src/model/encoding/convertFromHTMLToContentBlocks.js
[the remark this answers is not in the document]
mitermayer: Just updated the list hack to come first

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -564,60 +607,128 @@ function getChunkForHTML(
...
+ *
+ * This is unstable and not part of the public API and should not be used by
+ * production systems. This functionality may be update/removed without notice.
+ */
+const convertFromHTMLtoContentBlocks = (
   html: string,
   DOMBuilder: Function = getSafeBodyFromHTML,
   blockRenderMap?: DraftBlockRenderMap = DefaultDraftBlockRenderMap,
-): ?{contentBlocks: ?Array<BlockNodeRecord>, entityMap: EntityMap} {
+  experimentalTreeDataSupport?: boolean = false,
```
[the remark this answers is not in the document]
mitermayer: The gk task is the next task i want to cover on a PR this was to avoid bloating this already complicated PR to review.

My goal is having a follow up PR which will move all the `experimentalTreeDataSupport` flags to be controlled by a `GK` flag, was hoping to discuss a bit in person with you prior to starting it

Issue #1514, opened 2017-11-16
Flow errors with basic Draft.js usage


bjohn465:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When creating a very simple Draft.js project, Flow errors within Draft.js code.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
I have a demo of this problem here: https://github.com/bjohn465/draft-and-flow-errors

**What is the expected behavior?**
Draft.js should not cause Flow errors.

As noted [in the demo's README](https://github.com/bjohn465/draft-and-flow-errors#what-i-think-is-wrong), I believe the problem comes from `blockRendererFn` and `blockStyleFn` being defined as optional props to the `DraftEditor` component, when both of these props have default values specified via `defaultProps`. They should be marked as required props, as Flow will figure out that they are optional because of the `defaultProps` property.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
I found this in Draft.js version 0.10.4, using Flow 0.59.0 (the issue is also present using Flow 0.53.1).

mitermayer:
Hi @bjohn465,

Thanks for reporting this issue, could you provide an example of how to replicate this ? Also we would love to upgrade Draft to flow 0.59.0, would you be interested in a submitting a PR for it ?

bjohn465:
Thanks for the response, @mitermayer. I have a demo repository here: https://github.com/bjohn465/draft-and-flow-errors

I can at least get a pull request together for the errors I was seeing. Those were errors when attempting to use Draft.js and not when running Flow on the Draft.js project itself.

bjohn465:
It turned out that there weren't any additional changes needed to upgrade Flow, so I fixed this issue and upgraded Flow in #1520.

PR #1520, opened 2017-11-17
Mark props that have defaults as required

bjohn465:
Fixes facebook/draft-js#1514

**Summary**

According to the [Flow documentation](https://flow.org/en/docs/react/components/#toc-using-default-props), when using default props, the prop should be marked as required, as "Flow will make sure that [the prop] is optional if you have a default prop for [it]."

This fixes Flow errors that would occur when using Draft.js in a project with Flow. Flow would complain that `blockRendererFn` and `blockStyleFn` could potentially be `undefined` when passed to `DraftEditorContents`, even though these two properties have default values, so they will never be undefined.

This commit also upgrades Flow to 0.59.0, since there weren't any additional Flow errors between the version of Flow that was being used (0.53.1) and the latest version of Flow (0.59.0).

**Test Plan**

 * Flow check passes
 * A sample project, such as https://github.com/bjohn465/draft-and-flow-errors, does not produce Flow errors when using Draft.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "eslint-plugin-react": "^7.3.0",
     "eslint-plugin-relay": "^0.0.8",
     "fbjs-scripts": "^0.8.0",
-    "flow-bin": "^0.53.1",
+    "flow-bin": "^0.59.0",
     "gulp": "^3.9.0",
     "gulp-babel": "^6.1.2",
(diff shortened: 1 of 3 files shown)


mitermayer:
Hi @bjohn465,

Thank you so much for submitting this PR! Will try importing it internally and see how this changes will affect our internal usage.

PR #1521, opened 2017-11-18
Exploration - Feature tree / Tree adapter transformer

mitermayer:
### Context

**Tree adapter to transformer**

This creates a raw tree data adapter to transform back and forth between raw tree and draft raw data

***

**Note:**
production systems.

### DIFF
--- a/src/component/utils/exploration/__tests__/DraftTreeAdapter-test.js
+++ b/src/component/utils/exploration/__tests__/DraftTreeAdapter-test.js
@@ -1,0 +1,232 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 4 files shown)


src/component/utils/exploration/__tests__/DraftTreeAdapter-test.js
```diff
@@ -0,0 +1,232 @@
...
+    ],
+    entityMap: {},
+  };
+
+  assertFromRawTreeStateToRawState(rawState);
+});
+
+test('must be able to convert from tree raw state with nested list blocks to raw state preserving lists depth', () => {
+  const rawState = {
+    blocks: [
```
flarnie: The snapshot output for this has 'depth' of 0 for all blocks. Should the depth reflect the nesting?

mitermayer: This is definitly a bug, must have introduced it while refactoring. Thank you for spotting this!

PR #1531, opened 2017-11-26
Bugfix - Making sure empty spaces leading prior to ul/ol are not

mitermayer:
**Converting HTML that had ol/ul with spaces prior to first list item would create extra unstyled blocks**

Before if we tried to convert:

```html
<ul>
   <li>a</li>
</ul>
```

we would get 2 content blocks, with a unstyled one for the UL.

note that the same without leading spaces was converting fine:

```html
<ul><li>a</li></ul>
```

**Test Plan**

Added unit tests and updated old snapshots

### DIFF
--- a/src/model/encoding/__tests__/__snapshots__/convertFromHTMLToContentBlocks-test.js.snap
+++ b/src/model/encoding/__tests__/__snapshots__/convertFromHTMLToContentBlocks-test.js.snap
@@ -1,9 +1,91 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`converts deeply nested html blocks when experimentalTreeDataSupport is enabled 1`] = `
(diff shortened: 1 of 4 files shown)


src/model/encoding/__tests__/__snapshots__/convertFromHTMLToContentBlocks-test.js.snap
```diff
@@ -407,19 +590,6 @@ exports[`does not convert deeply nested html blocks when experimentalTreeDataSup
 
 exports[`does not convert deeply nested html blocks when experimentalTreeDataSupport is disabled 2`] = `
 Array [
-  Object {
```
mitermayer: this demonstrates that this bug was already reproducible with the old non tree code path.

PR #1532, opened 2017-11-27
Exploration - Feature tree / Wiring up tree adapter on the raw

mitermayer:
### Context

**Wiring up tree adapter on the raw**

Making sure when the darkfeature flag is enabled/disabled we can handle both data formats to guarantee some degree of migration/backwards compatibility

***

**Note:**
production systems.

### DIFF
--- a/src/model/encoding/convertFromRawToDraftState.js
+++ b/src/model/encoding/convertFromRawToDraftState.js
@@ -23,6 +23,8 @@ const ContentBlock = require('ContentBlock');
 const ContentBlockNode = require('ContentBlockNode');
 const ContentState = require('ContentState');
 const DraftEntity = require('DraftEntity');
+const DraftFeatureFlags = require('DraftFeatureFlags');
+const DraftTreeAdapter = require('DraftTreeAdapter');
 const Immutable = require('immutable');
 const SelectionState = require('SelectionState');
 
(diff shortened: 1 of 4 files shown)


src/model/encoding/convertFromRawToDraftState.js
```diff
@@ -210,10 +214,19 @@ const decodeRawBlocks = (
...
 ): BlockMap => {
-  const rawBlocks = rawState.blocks;
-
-  if (!Array.isArray(rawBlocks[0].children)) {
-    return decodeContentBlocks(rawBlocks, entityMap);
+  const isTreeRawBlock = Array.isArray(rawState.blocks[0].children);
+  const rawBlocks =
+    experimentalTreeDataSupport && !isTreeRawBlock
+      ? DraftTreeAdapter.fromRawStateToRawTreeState(rawState).blocks
+      : rawState.blocks;
```
flarnie: This is a nit, but I would find this more clear if we split out the behavior into two branches based on whether `experimentalTreeDataSupport` is true.

If it's true, we want to:
- convert all blocks into TreeBlocks if needed
- return `decodeContentBlockNodes(rawBlocks, entityMap)`

If it's false, we want to:
- convert all blocks into non-treeBlocks if needed
- return `decodeContentBlocks(rawBlocks, entityMap)`

This is mainly for my own ability to read things, let me know if you want me to take a pass at it after you land this.

mitermayer: This makes total sense to me, will update this PR to the above recommendation

mitermayer: updating this inside phabricator

PR #1533, opened 2017-11-27
Exploration - Feature tree / Adding support for container text nodes

mitermayer:
### Context

**Adding support for container text nodes**

This PR will add support for handling container text nodes that precedes real block nodes.

example:

```html
<blockquote>
   some text
   <h1>header block</h1>
</blockquote>
```

***

**Note:**
production systems.

### DIFF
--- a/src/model/encoding/convertFromHTMLToContentBlocks.js
+++ b/src/model/encoding/convertFromHTMLToContentBlocks.js
@@ -643,6 +643,42 @@ const convertChunkToContentBlocks = (chunk: Chunk): ?Array<BlockNodeRecord> => {
     const {depth, type, parent} = block;
 
     const key = block.key || generateRandomKey();
+    let parentTextNodeKey = null; // will be used to store container text nodes
+
+    // childrens add themselves to their parents since we are iterating in order
+    if (parent) {
+      const parentIndex = acc.cacheRef[parent];
+      let parentRecord = acc.contentBlocks[parentIndex];
+
+      // if parent has text we need to split it into a separate unstyled element
+      if (parentRecord.getChildKeys().isEmpty() && parentRecord.getText()) {
+        const parentCharacterList = parentRecord.getCharacterList();
+        const parentText = parentRecord.getText();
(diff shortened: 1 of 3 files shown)


src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -643,6 +643,42 @@ const convertChunkToContentBlocks = (chunk: Chunk): ?Array<BlockNodeRecord> => {
...
     const key = block.key || generateRandomKey();
+    let parentTextNodeKey = null; // will be used to store container text nodes
+
+    // childrens add themselves to their parents since we are iterating in order
+    if (parent) {
+      const parentIndex = acc.cacheRef[parent];
+      let parentRecord = acc.contentBlocks[parentIndex];
+
+      // if parent has text we need to split it into a separate unstyled element
+      if (parentRecord.getChildKeys().isEmpty() && parentRecord.getText()) {
```
flarnie: If I'm reading this right, then this conditional will be true whenever the parent has no children and has any text. It will take all the text and put it into a textNode, and add that to the children.

It seems this will happen even for `<blockquote>hi</blockquote>` as well as `<blockquote>hi <h1>world</h1></blockquote>`. Why do it this way? Maybe more consistent?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -652,25 +688,16 @@ const convertChunkToContentBlocks = (chunk: Chunk): ?Array<BlockNodeRecord> => {
       text: textBlock,
       characterList,
       prevSibling:
-        index === 0 || rawBlocks[index - 1].parent !== parent
+        parentTextNodeKey ||
```
flarnie: I could use a bit more explanation for why we put the `parentTextNodeKey` as the prevSibling. It makes sense that it's what we want - like in the case of `<blockquote>hi <h1>world</h1></blockquote>`. I'll re-read this again and get more clarity.

PR #1540, opened 2017-11-30
Exploration - Feature tree / Adding support for moving tree blocks

mitermayer:
### Context

**Adding support for moving tree blocks**

This PR adds support for moving tree blocks via the moveBlockInContentState transaction.

***

**Note:**
production systems.

### DIFF
--- a/src/model/transaction/__tests__/moveBlockInContentState-test.js
+++ b/src/model/transaction/__tests__/moveBlockInContentState-test.js
@@ -68,7 +68,7 @@ const contentBlockNodes = [
   new ContentBlockNode({
     key: 'E',
     parent: 'D',
-    text: 'Elefant',
+    text: 'Elephant',
   }),
 ];
 
@@ -95,15 +95,16 @@ const assertMoveBlockInContentState = (
       .getBlockMap()
       .toSetSeq()
       .toArray()
+      // doing this filtering to make the snapshot more precise/concise in what we test
       .map(filter => {
-          const {
-             data,
-             characterList,
-             depth,
-             type,
-             text,
-             ...other,
-          } = filter.toJS();
+        const {
+          data,
+          characterList,
+          depth,
+          type,
+          text,
+          ...other
+        } = filter.toJS();
         return other;
       }),
   ).toMatchSnapshot();
--- a/src/model/transaction/__tests__/moveBlockInContentState-test.js
+++ b/src/model/transaction/__tests__/moveBlockInContentState-test.js
@@ -1,0 +1,150 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @emails oncall+ui_infra
+ * @format
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
+jest.mock('generateRandomKey');
+
+const EditorState = require('EditorState');
+const ContentBlock = require('ContentBlock');
+const ContentBlockNode = require('ContentBlockNode');
+const ContentState = require('ContentState');
+const Immutable = require('immutable');
+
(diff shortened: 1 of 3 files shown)


src/model/transaction/moveBlockInContentState.js
```diff
@@ -13,83 +13,272 @@
...
+  blockMap: BlockMap,
+  originalBlockToBeMoved: BlockNodeRecord,
+  originalTargetBlock: BlockNodeRecord,
+  insertionMode: DraftInsertionType,
+  isExperimentalTreeBlock: boolean,
+): BlockMap => {
+  if (!isExperimentalTreeBlock) {
+    return blockMap;
+  }
+  // insertionMode can only either be after or before
```
flarnie: I think this would be more clear if we say
```
// insertionMode can only either be 'after' or 'before'
```
At first I misunderstood and thought that it could only fall after or before some unnamed thing, but we mean that the possible values of the 'insertionMode' variable are 'after' and 'before'.
Maybe even something like this:
```
// possible values of 'insertionMode' are: 'after', 'before'
```

src/model/transaction/moveBlockInContentState.js
```diff
@@ -13,83 +13,272 @@
...
+        newParentBlock.merge({
+          children: newParentChildrenList.insert(
+            insertionIndex,
+            originalBlockKey,
+          ),
+        }),
+      );
+    }
+
+    // update block
```
flarnie: Love the comments in this section. :)

src/model/transaction/moveBlockInContentState.js
```diff
@@ -13,83 +13,272 @@
...
+  const blockKey = blockToBeMoved.getKey();
+
+  invariant(blockKey !== targetKey, 'Block cannot be moved next to itself.');
 
   const blockMap = contentState.getBlockMap();
-  const blockMapWithoutBlockToBeMoved = blockMap.delete(
-    blockToBeMoved.getKey(),
-  );
-  const blocksBefore = blockMapWithoutBlockToBeMoved
+  const isExperimentalTreeBlock = blockToBeMoved instanceof ContentBlockNode;
```
flarnie: I'm curious why we use this check inestd of checking for the 'children' array like we do in other places?

src/model/transaction/__tests__/moveBlockInContentState-test.js
```diff
@@ -0,0 +1,150 @@
...
+
+test('must be able to move block before other block', () => {
+  assertMoveBlockInContentState('C', 'A', 'before');
+});
+
+test('must be able to move block after other block', () => {
+  assertMoveBlockInContentState('A', 'C', 'after');
+});
+
+test('must be able to move nested block before other block', () => {
```
flarnie: I'm missing something here - in the snapshot it looks like it moved D and E too, even though they are not children of C.

src/model/transaction/moveBlockInContentState.js
```diff
@@ -13,83 +13,272 @@
...
+      nextNonDescendantBlock.getParentKey(),
+    );
+  }
+
+  if (!nextNonDescendantBlock) {
+    return null;
+  }
+
+  return nextNonDescendantBlock.getNextSiblingKey();
+};
```
flarnie: cat_its_beautiful

src/model/transaction/__tests__/moveBlockInContentState-test.js
mitermayer: You are correct, just checked on the snapshot and it is a bug! will fix this! thank you for spotting this!

mitermayer: Turns out this is a bug in immutable where they do not recognize `null` as a false value for `takeWhile` just checked with lee and will submit a separate PR for immutable. In the meantime we will just cast that to a boolean with `!!(...)`

src/model/transaction/moveBlockInContentState.js
mitermayer: We check for `children` array only on raw checks, internally we can check as instanceof since they are already record types

PR #1547, opened 2017-12-05
Exploration - Feature tree / Adding support for removing tree nodes ranges from content state

mitermayer:
### Context

**Adding support for removing content state**

This PR adds support for removing range transactions for ContentBlockNodes 

***

**Note:**
production systems.

### DIFF
--- a/src/model/transaction/removeRangeFromContentState.js
+++ b/src/model/transaction/removeRangeFromContentState.js
@@ -298,7 +298,7 @@ const removeRangeFromContentState = (
     .toSeq()
     .skipUntil((_, k) => k === startKey)
     .takeUntil((_, k) => k === endKey)
-    .filterNot((_, k) => parentAncestors.indexOf(k) !== -1)
+    .filter((_, k) => parentAncestors.indexOf(k) === -1)
     .concat(Map([[endKey, null]]))
     .map((_, k) => {
       return k === startKey ? modifiedStart : null;
--- a/src/model/transaction/removeRangeFromContentState.js
+++ b/src/model/transaction/removeRangeFromContentState.js
@@ -13,30 +13,266 @@
 
 'use strict';
 
+import type {BlockMap} from 'BlockMap';
 import type CharacterMetadata from 'CharacterMetadata';
 import type ContentState from 'ContentState';
 import type SelectionState from 'SelectionState';
-import type {List} from 'immutable';
 
-var Immutable = require('immutable');
+const ContentBlockNode = require('ContentBlockNode');
+const Immutable = require('immutable');
 
-function removeRangeFromContentState(
+const getNextDelimiterBlockKey = require('getNextDelimiterBlockKey');
+
+const {List, Map} = Immutable;
+
+const transformBlock = (
+  key: ?string,
+  blockMap: BlockMap,
+  func: (block: ContentBlockNode) => void,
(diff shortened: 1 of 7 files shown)


src/model/transaction/removeRangeFromContentState.js
```diff
@@ -13,30 +13,266 @@
...
+    // we do not need to keep checking all root node siblings, just the first occurance
+    nextDelimiter = block.getParentKey()
+      ? getNextDelimiterBlockKey(block, blockMap)
+      : null;
+  }
+
+  return nextDelimiters;
+};
+
+const getNextValidSibling = (
```
flarnie: I'm curious what 'valid' means - it looks like it means that the block is not part of the section being removed. I'll keep reading and see if I'm right.

src/model/transaction/removeRangeFromContentState.js
```diff
@@ -13,30 +13,266 @@
...
+        prevSibling: startBlock.getPrevSiblingKey(),
+      }),
+    );
+
+    // update start block prev
+    transformBlock(startBlock.getPrevSiblingKey(), blocks, block =>
+      block.merge({
+        nextSibling: getNextValidSibling(startBlock, blocks, originalBlockMap),
+      }),
+    );
```
flarnie: It looks like this is removing the 'start block' from the chain of siblings. Like so:
**Before:**
```
A <-> StartBlock <-> B
```
to
```
A <-> B
StartBlock
```

Will keep reading - might be missing something.

src/model/transaction/removeRangeFromContentState.js
mitermayer: I should maybe rename this. `getNextValidSibling` is used to identify blocks that are still existing on the new modified blockmap. Since blocks are removed by range and we have special rules for retaining blocks if they have children we need to make sure they are still existent on the newly updated blockmap

src/model/transaction/removeRangeFromContentState.js
mitermayer: From this example `A <-> StartBlock <-> B`

startBlock.getPrevSiblingKey() = 'A'

we then update the "A" block nextSibling property to the nearest valid next sibbling (which can be either after the removed range or even sometimes within the range if there was a sibling that got retained for having children)

flarnie: Aaah I see. Thanks, it was helpful to have it written out.

src/model/transaction/removeRangeFromContentState.js
```diff
@@ -13,30 +13,266 @@
...
+
+    // update end block next
+    transformBlock(endBlock.getNextSiblingKey(), blocks, block =>
+      block.merge({
+        prevSibling: getPrevValidSibling(endBlock, blocks, originalBlockMap),
+      }),
+    );
+
+    // update end block prev
```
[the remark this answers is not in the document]
mitermayer: Actuall this coment is correct, we are updating the block prev "nextSibling" property,

```
<blockPrev> <=> endBlock <=> <blockNext>
```

Since we are removing the endBlock we attemp to update `blockPrev` and `blockNext` if they exists. 

`blockPrev` will need to find a valid next block and
`blockNext` will also need to  find a valid previous block

PR #1548, opened 2017-12-05
Exploration - Feature tree / Adding support for splitting transactions

mitermayer:
### Context

**Adding support for splitting transactions**

This PR adds support for tree block nodes splitting transactions

***

**Note:**
production systems.

### DIFF
--- a/src/model/transaction/splitBlockInContentState.js
+++ b/src/model/transaction/splitBlockInContentState.js
@@ -13,54 +13,124 @@
 
 'use strict';
 
+import type {BlockMap} from 'BlockMap';
 import type ContentState from 'ContentState';
 import type SelectionState from 'SelectionState';
 
-var Immutable = require('immutable');
+const ContentBlockNode = require('ContentBlockNode');
+const Immutable = require('immutable');
 
-var generateRandomKey = require('generateRandomKey');
-var invariant = require('invariant');
+const generateRandomKey = require('generateRandomKey');
+const invariant = require('invariant');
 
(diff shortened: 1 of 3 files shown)


src/model/transaction/splitBlockInContentState.js
```diff
@@ -13,55 +13,125 @@
...
-var generateRandomKey = require('generateRandomKey');
-var invariant = require('invariant');
+const generateRandomKey = require('generateRandomKey');
+const invariant = require('invariant');
 
-const {Map} = Immutable;
+const {List, Map} = Immutable;
 
-function splitBlockInContentState(
+const transformBlock = (
```
mitermayer: Worth noting that his is the exact same function used on the remove ranges PR (https://github.com/mitermayer/draft-js/blob/d151ad92d52c0548156a81a7fec399bb2da3d6ea/src/model/transaction/removeRangeFromContentState.js#L28) 

We can either rebase either PR's to which one is merged first or do a follow up PR to remove the duplication

mitermayer: We should also during that process refactor moveBlockInContentState to also use `transformBlock`

flarnie: Doing a follow-up seems like a good idea - since these are scoped, well defined tasks we can also open issues for folks in the community to DRY up the code.

juliankrispel:
@flarnie that is one amazing license plate 🤣

Issue #1562, opened 2017-12-11
CI is failing with segfault


flarnie:
Looking into this - so far we are not sure what is causing it, but it means CI is failing on all new PRs.

flarnie:
Interesting - the failed build logs show that the last thing CI tried to do was build the website.

If I try to follow the steps locally, I do get an error, when I `cd website && npm install`. 
https://travis-ci.org/facebook/draft-js/jobs/309793188#L865-L866

What I am seeing locally is a mismatch between jest versions. Looking into where that is coming from.

flarnie:
Edit - CI is actually failing on the `npm run lint` step.

flarnie:
And for me this does repro locally - 
![screen shot 2017-12-11 at 8 07 58 am](https://user-images.githubusercontent.com/1114467/33840676-760aa112-de4a-11e7-858e-9752b94ba7bf.png)

flarnie:
Checking out the last commit where CI passed does not change this error.

flarnie:
If I jump all the way back to v0.10.4, remove `node_modules`, do a fresh `yarn install && yarn lint`, no error.

flarnie:
This is 100% related to prettier - I found that lint errors on all commits after cd20b99 "Setting up prettier", and passes on commits earlier than that one.

flarnie:
Ok - the failure is due to some change in the Prettier 1.9.* release about 6 days ago.

Here is what I'm going to do:
- Pin Draft dev dependencies to an earlier version of Prettier
- Open an issue with Prettier
- Open a Draft issue to move to Yarn and commit a yarn lockfile (unless we already have an issue for this)

The only way to prevent similar issues in the future is to tightly restrict our dependencies, such that we can't start pulling in an upgraded version without first verifying that all tests pass with that upgrade. That is another follow-up item to consider - the cost is that we have more overhead for dependency management rather than having broken CI once in a while.

flarnie:
Opened an issue with Prettier - https://github.com/prettier/prettier/issues/3457

flarnie:
Opened an issue to use `yarn`.

flarnie:
This was resolved by downgrading the 'prettier' version, but using an older version than the Facebook codebase has opened new problems. We are still seemingly unable to land changes.

PR #1563, opened 2017-12-11
Experiment - does updating Jest fix CI?

flarnie:
**what is the change?:**
Bump the Jest version, undoing the revert commit landed in
https://github.com/facebook/draft-js/commit/e127090a2c23ea4de3e7396fa310a3ecc8c735ce

**why make this change?:**
CI is failing to do a segfault, and we have no strong theories yet about
why. The last commit to land before it started failing was the revert
commit, so let's see if that's what caused the failure.

**test plan:**
push this to Github and see if CI passes

tests are passing locally

**issue:**
#1562

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-uglify": "^1.2.0",
     "gulp-util": "^3.0.6",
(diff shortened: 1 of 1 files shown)


flarnie:
Looks like we can rule this out.

PR #1564, opened 2017-12-11
Trying to fix ci part 2

flarnie:
In order to verify that none of the commits landed recently are related to the CI failure, this PR reverts them and will then run CI.

I'm 90% sure this won't cause CI to pass - it it likely related to a change in a dependency or something in Circle CI itself.


flarnie:
Glad we can rule this out.

PR #1567, opened 2017-12-11
Restrict 'prettier' version to unbreak CI

flarnie:
**what is the change?:**
We have verified that upgrading to Prettier v1.9.0 or v1.9.1 both cause
a Segmentation Fault when running `eslint`. This happens in CI, and I
was able to verify it locally.

**why make this change?:**
We want to use Prettier in CI and not get segmentation faults.

**test plan:**
Locally I did -
`rm -rf node_modules && yarn install && yarn lint` and got no segfault.
Pushing this to github and waiting for CI to pass will further verify.

**issue:**
fixes issue #1562

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "gulp-uglify": "^1.2.0",
     "gulp-util": "^3.0.6",
     "jest": "^21.2.1",
-    "prettier": "^1.7.4",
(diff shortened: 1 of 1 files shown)


flarnie:
Note - I will probably just merge this once CI passes, tagging folks as reviewers for visibility and in case you have recommendations for follow-up work.

PR #1575, opened 2017-12-14
Pinning prettier to 1.9.1 and fixing travis segfault

mitermayer:
Internally we are now using prettier 1.9.1.

This PR pins prettier to `1.9.1` and fixes the segfault builds that were caused by running travis on old node versions


mitermayer:
@flarnie , Looks like this combination fixes our builds

PR #1576, opened 2017-12-14
Lock prettier version to 1.9.1

zpao:
www is now on 1.9.1. Since Draft syncs into and is beholden to the pre-commit hook, we should be using the same version.

Lint is clean, so looks like everything is in sync currently. Upgrading past 1.9.1 does cause a failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-- 6
+- 8
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


mitermayer:
Tested some of the changes from this PR on another PR and managed to get a green build

see on https://github.com/facebook/draft-js/pull/1575/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L84

if we updated `package.json` "node" version to also support `v8` the build then passes with no problems

mitermayer:
travis builds for it (https://travis-ci.org/facebook/draft-js/builds/316814356)

mitermayer:
Once we apply that patch we then should be able to land this PR and solve all problems :)

zpao:
Let's just land yours @mitermayer :)

PR #1580, opened 2017-12-15
Re-enabling linting errors as part of the build

mitermayer:
This PR makes linting part of the CI build once again. While we were out of sync with our internal Prettier version we had to downgrade linting as warnings, this brings prettier linting back to error.

### DIFF
--- a/src/model/transaction/__tests__/moveBlockInContentState-test.js
+++ b/src/model/transaction/__tests__/moveBlockInContentState-test.js
@@ -73,6 +73,15 @@ const contentBlockNodes = [
   }),
 ];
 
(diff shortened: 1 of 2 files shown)


src/model/transaction/__tests__/moveBlockInContentState-test.js
```diff
@@ -96,18 +105,15 @@ const assertMoveBlockInContentState = (
...
-          characterList,
-          depth,
-          type,
-          text,
-          ...other
-        } = filter.toJS();
-        return other;
-      }),
+      .map(filter => filter.toJS())
+      .map(block =>
```
mitermayer: move this out to fix linting warnings

Issue #1454, opened 2017-10-25
Select and drag text in the editor is broken, can recreate this on draftjs.org


d-turner:
Do you want to request a feature or report a bug?
bug

What is the current behavior?
Select and drag text in the editor is broken, causes unpredictable behaviour.
Can recreate this on draftjs.org
Ubuntu:
Chromium Version 61.0.3163.100
Firefox Version 56.0
Windows:
Chrome

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: **

Go to draftjs.org, enter a sentence into the editor, double click the first word to highlight it and drag it to another location in the editor.

The observed result is near different each time and is not what is expected, sometimes the dragged word remains in the original location and in the drop location, sometimes it is placed a few characters after the drop location.

It is tested at https://draftjs.org/

What is the expected behavior?

Move the selected text to the desired caret position.

Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?

It is tested at https://draftjs.org/
Version 0.10.1
Version 0.10.4

flarnie:
Thanks for reporting this!

I just tried to reproduce it quickly in Chrome on MacOSX, and dragging text worked as expected for me.

![dragging_text_ok_in_chrome](https://user-images.githubusercontent.com/1114467/32002961-947f775c-b952-11e7-8422-7da0e43f2d58.gif)

Can you provide more info about how to reproduce this? What browser and operating system are you using?

d-turner:
Thanks for the quick response, I have tried this on Windows Chrome and Ubuntu Chrome and Chromium
Below is the example:

![peek 2017-10-25 15-54](https://user-images.githubusercontent.com/6719400/32006212-08ba3c1a-b99e-11e7-81ed-824b0a2ffb6f.gif)

flarnie:
Aha - I was testing on Firefox. This is indeed easy to reproduce on Chrome, and what an awful bug! Thanks for reporting it!

I'd love to get a PR fixing it. This seems recent - I'm surprised I didn't catch it when manually testing in the past.

You mentioned being able to reproduce it in 0.10.1 - I wonder if it's a problem introduced by a Chrome version update rather than a change in Draft.

Either way, would be great to get a fix. If nobody picks this up I'll look into it when I have time.

itssujan:
I tried this on Mac Chrome and seems to be working fine.

[...]

PR #1582, opened 2017-12-17
Fix dragging text ahead on same line

jackyho112:
**Summary**

Fixes https://github.com/facebook/draft-js/issues/1454 so that dragging text to the back of the same line or the last line of the multiple selected lines puts the text in the correct position and does not produce any error. 

An example with the fix:

mm1lwk8h8j

**Test Plan**

Added unit tests for the fix. 

Let me know if there is any improvement you have in mind. In particular, I took a crack at writing some tests for `moveText` of `DraftModifier` and hope they are appropriate. 😄


src/model/modifier/DraftModifier.js
```diff
@@ -102,6 +102,31 @@ var DraftModifier = {
     targetRange: SelectionState,
   ): ContentState {
     var movedFragment = getContentStateFragment(contentState, removalRange);
+    var removeRangeEndOffset = removalRange.isBackward
```
colinjeanne: Why not use `getStartOffset()`? Similarly, why not use `getEndOffset()` below?

PR #1589, opened 2017-12-25
Add instructions on how to add Draft.css

xixixao:
This will work with create-react-app. Hopefully will save the next person 10 seconds.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -75,6 +75,10 @@ Because Draft.js supports unicode, you must have the following meta tag in the `
 <meta charset="utf-8" />
 ```
 
-`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css).
+`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css). With Babel you can add it like this:
+
+```es6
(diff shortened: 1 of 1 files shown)


docs/Overview.md
```diff
@@ -75,6 +75,10 @@ Because Draft.js supports unicode, you must have the following meta tag in the `
 <meta charset="utf-8" />
 ```
 
-`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css).
+`Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css). With Babel you can add it like this:
```
thibaudcolas: This isn't achieved with Babel but with Webpack (and some loaders). I think you could just mention `create-react-app` instead since the docs mention it in other places.

flarnie: Yea, there a ton of different build systems and infrastructures out there, not sure this would be helpful to folks who may not be using CRA.

I would accept a docs page with tips or instructions for using Draft.js with CRA specifically.

hozefaj: Maybe highlight couple of the most obvious(or most used) build systems. Not a good user experience here where the docs mention that use `Draft.css` but nothing around where to get it from.

Maybe if needed create a new page for it, but some info is definitely needed.

PR #1593, opened 2018-01-03
Exploration - Feature tree / adding internal support for tree data structures

mitermayer:
### Context

**Adding support for tree data structures**

This PR includes the missing parts for adding support for internal tree data structure by providing:
    
* last transaction insertFragmentIntoContentState.js
* last modifier AtomicBlockUtils
* fixing a bug on the randomizeBlockMapKeys to handle orphanated fragments

This is the last PR of the series to add support for internal tree data structures, from here we should then either update `RichTextEditorUtil` to have some `tree` awareness and having some controls for creating nesting content or create a new `NestedRichTextEditorUtil` that can be used for that. Ideally that work would also come with a `demo` example page

***

**Note:**
production systems.

### DIFF
--- a/src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
+++ b/src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
@@ -1,1306 +1,5573 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`must be able to insert atomic block when experimentalTreeDataSupport is enabled 1`] = `
+Array [
+  Object {
+    "characterList": Array [],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "A",
+    "nextSibling": "key0",
+    "parent": null,
+    "prevSibling": null,
+    "text": "",
+    "type": "unstyled",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": "2",
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key0",
+    "nextSibling": "key4",
+    "parent": null,
+    "prevSibling": "A",
+    "text": " ",
+    "type": "atomic",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key4",
+    "nextSibling": null,
+    "parent": null,
+    "prevSibling": "key0",
+    "text": "first block",
+    "type": "unstyled",
+  },
+]
+`;
+
+exports[`must be able to move atomic block when experimentalTreeDataSupport is enabled 1`] = `
+Array [
+  Object {
+    "characterList": Array [],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "A",
+    "nextSibling": "key0",
+    "parent": null,
+    "prevSibling": null,
+    "text": "",
+    "type": "unstyled",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": "2",
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key0",
+    "nextSibling": "key4",
+    "parent": null,
+    "prevSibling": "A",
+    "text": " ",
+    "type": "atomic",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key4",
+    "nextSibling": null,
+    "parent": null,
+    "prevSibling": "key0",
+    "text": "first block",
+    "type": "unstyled",
+  },
+]
+`;
+
+exports[`must be able to move atomic block when experimentalTreeDataSupport is enabled 2`] = `
+Array [
+  Object {
+    "characterList": Array [],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "A",
+    "nextSibling": "key4",
+    "parent": null,
+    "prevSibling": null,
+    "text": "",
+    "type": "unstyled",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key4",
+    "nextSibling": "key0",
+    "parent": null,
+    "prevSibling": "A",
+    "text": "first block",
+    "type": "unstyled",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": "2",
+        "style": Array [],
+      },
+    ],
+    "children": Array [],
+    "data": Object {},
+    "depth": 0,
+    "key": "key0",
+    "nextSibling": null,
+    "parent": null,
+    "prevSibling": "key4",
+    "text": " ",
+    "type": "atomic",
+  },
+]
+`;
+
 exports[`must insert atomic after a block with collapsed selection 1`] = `
-Object {
-  "a": Object {
+Array [
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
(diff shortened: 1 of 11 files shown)


src/model/transaction/__tests__/__snapshots__/getContentStateFragment-test.js.snap
```diff
@@ -261,7 +261,7 @@ Object {
     "data": Object {},
     "depth": 0,
     "key": "key10",
-    "nextSibling": "D",
+    "nextSibling": null,
```
mitermayer: This was a bug that I managed to notice while developing this PR and fixing the randomizeBlockMapKeys.

src/model/transaction/moveBlockInContentState.js
```diff
@@ -26,6 +26,24 @@ const invariant = require('invariant');
 
 const {OrderedMap, List} = Immutable;
 
+const transformBlock = (
```
mitermayer: Once this PR is merged will then start refactoring our codebase for removing duplication and extracting things like this into utils (I believe this function already exists in a few files) could be also good PR's for contributors

src/model/transaction/randomizeBlockMapKeys.js
```diff
@@ -23,7 +23,11 @@ const generateRandomKey = require('generateRandomKey');
 const {OrderedMap} = Immutable;
 
 const randomizeContentBlockNodeKeys = (blockMap: BlockMap): BlockMap => {
-  const newKeys = [];
+  const newKeysRef = {};
+
+  // we keep track of root blocks in order to update subsequent sibling links
+  let lastRootBlock: ContentBlockNode;
```
mitermayer: we need to keep track of previous seen root blocks, since randomizeBlockMapKeys is responsible for dealing with fragments, example is when user select some children of a block for copy and paste without including all siblings and or parent, in that case the fragment need to have the parent links cleaned up

mitermayer:
looks like we have another travis dependencie problem, will fix this on another PR.

flarnie:
Reviewing this today - sorry for the wait. Will look into the build system problems next week.

src/model/transaction/__tests__/insertFragmentIntoContentState-test.js
```diff
@@ -107,3 +116,277 @@ test('must apply multiblock fragments', () => {
...
+        children: List(['D']),
+      },
+      {
+        key: 'D',
+        parent: 'C',
+        text: 'Delta',
+      },
+      {
+        key: 'E',
+        text: 'Delta',
```
flarnie: `'Delta'` -> `'Elephant'` pls

flarnie:
Review progress:

- [x] `AtomicBlockUtils`
- [ ] tests for `AtomicBlockUtils`
- [ ] `moveBlockIntoContentState`
- [ ] tests for `moveBlockIntoContentState`
- [ ] `getContentStateFragment`
- [ ] tests for `getContentStateFragment`
- [x] `insertFragmentIntoContentState`
- [x] tests for `insertFragmentIntoContentState`
- [ ] `randomizeBlockMapKeys`
- [ ] tests for `randomizeBlockMapKeys`

src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
```diff
@@ -1,1306 +1,5573 @@
...
-    "key": "key11",
-    "text": "",
-    "type": "unstyled",
-  },
-  "key8": Object {
-    "key": "key8",
-    "text": " ",
-    "type": "atomic",
-  },
-}
```
flarnie: I am concerned that the ordering of the blocks in the `must insert atomic after a block with collapsed selection 1` snapshot has changed. Is that expected?

It was `a, b, c, d, e, f, key11, key8`.
In the updated snapshot it is `a, key0, key4, b, c, d, e, f`.

src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
```diff
@@ -1,1306 +1,5573 @@
...
+      },
+    ],
+    "data": Object {},
+    "depth": 0,
     "key": "f",
     "text": "Charlie",
     "type": "blockquote",
   },
-  "key0": Object {
-    "key": "key0",
```
flarnie: I'm confused about why the old snapshot had keys `key0` and `key3` but the new snapshot doesn't seem to, for `must insert atomic at start of block with collapsed seletion 1`.

src/model/transaction/moveBlockInContentState.js
```diff
@@ -124,18 +121,14 @@ const updateBlockMapLinks = (
...
+        children: List(newChildrenArray),
+      });
+    });
 
     // update block
-    blocks.mergeIn(
-      originalBlockKey,
-      originalBlockToBeMoved.merge({
+    transformBlock(originalBlockKey, blocks, block =>
+      block.merge({
```
flarnie: Very nice!

src/model/transaction/randomizeBlockMapKeys.js
flarnie: This all looks great! Nice refactor, plus fixing the parent links like you mentioned above.

src/model/transaction/__tests__/__snapshots__/randomizeBlockMapKeys-test.js.snap
```diff
@@ -1,8 +1,130 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
flarnie: This snapshot/test lgtm. 👍

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+  const prependToTail = fragment.last();
+
+  return prependToTail.merge({
+    text: prependToTail.getText() + tailText,
+    characterList: prependToTail.getCharacterList().concat(tailCharacters),
+    data: prependToTail.getData(),
+  });
+};
+
+// TODO fix flow
```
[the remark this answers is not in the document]
mitermayer: Will see if If that comment is still relevant and try to fix in here, if not will create a new task and fix it straight away once this is merged

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+const getRootBlocks = (block: ContentBlockNode, blockMap: BlockMap): any => {
+  const headKey = block.getKey();
+  let rootBlock = block;
+  let rootBlocks = [];
+
+  // we need to account to when the head fragment block is merged with the target block
```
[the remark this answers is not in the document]
mitermayer: Will try to rephrase this to something more meaningful, what that was meant to do is, in some cases we will merge the first block from the fragment with the target block (whenever possible) but when that is not possible we insert all fragments after it. 

example trying to merge a block B with text "Beta" in a block A with text "Alpha" would yeild block A with text AlphaBeta. But if block A or block B had children then we would preserve the normal block A and add block B after it

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+  return blockMap.withMutations(blockMapState => {
+    const targetKey = targetBlock.getKey();
+    const headKey = fragmentHeadBlock.getKey();
+    const targetNextKey = targetBlock.getNextSiblingKey();
+    const targetParentKey = targetBlock.getParentKey();
+    const fragmentRootBlocks = getRootBlocks(fragmentHeadBlock, blockMap);
+    const lastRootFragmentBlockKey =
+      fragmentRootBlocks[fragmentRootBlocks.length - 1];
+
+    if (didNotMergeFirstFragmentBlockWithTargetBlock) {
```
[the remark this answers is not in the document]
mitermayer: Will try to refactor on this PR or depending the amount of work create a follow up task and action on it straight away

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
 
-    newBlockArr.push(modifiedHead);
+      const newChildrenArray = originalTargetParentChildKeys.toArray();
+
+      // insert fragment children
+      newChildrenArray.splice(insertionIndex, 0, ...fragmentRootBlocks);
+
+      blockMapState.setIn(
+        [targetParentKey, 'children'],
+        new List(newChildrenArray),
```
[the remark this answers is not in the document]
mitermayer: Will look into this, and answer in here with my findings. If this is a bug and was not caught on my tests will create a new test covering that scenario!

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+  const head = fragment.first();
+  const tail = fragment.last();
+  const finalOffset = tail.getLength();
+  const finalKey = tail.getKey();
+  const shouldNotUpdateFromFragmentBlock =
+    isTreeBasedBlockMap &&
+    (!target.getChildKeys().isEmpty() || !head.getChildKeys().isEmpty());
+
+  //
+  // TODO - improve the bellow comment
```
mitermayer: will also remove this todos

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+  // the first fragment block's content is used to update the target block
+  // we should not skip the first fragment block since we cannot merge the
+  // target block with the first fragment block contents
+  const startFragmentOffset = shouldNotUpdateFromFragmentBlock ? 0 : 1;
+
+  // check here why the fragment block insertion seems messed up
+  // on the snapshots the 'root' block is not appearing correct order
```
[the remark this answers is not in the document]
mitermayer: I am pretty sure i have addressed this before, I think i may have had a bad master rebase, will look into this, specially with all the todo comments

src/model/transaction/insertFragmentIntoContentState.js
```diff
@@ -13,121 +13,280 @@
...
+    }
+
+    // update head
+    newBlockArr.push(
+      updateHead(
+        block,
+        targetOffset,
+        fragment,
+        shouldNotUpdateFromFragmentBlock,
+      ),
```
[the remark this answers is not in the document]
mitermayer: We could refactor this to simplify that logic, the tricky part is that we can't just add the block since we also need to update its links, will try to think on how to address this without using a boolean flag, if it turns out to be too complex i may address this in follow up PRS

src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
mitermayer: The previous order was incorrect, it was like that due to when we convert ordered map to javascript the order is alphanumeric not order of insertion, when we changed the assertions to use indexed sequence we guaranteed that the ordering of the snapshot was correct.

src/model/transaction/insertFragmentIntoContentState.js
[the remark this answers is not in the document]
mitermayer: will remove the the need to pass boolean flags around, I agree with you and they are not great solutions

src/model/transaction/insertFragmentIntoContentState.js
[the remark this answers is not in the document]
mitermayer: That behaviour is correct, newChildrenArray is an Array<string>, will rename it to newChildrenKeysArray to make sure it reduce confusion

src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
mitermayer: I have just double checked, and if I update the assertion method to how did before

```javascript
      .getBlockMap()
      .map(block => ({
        key: block.getKey(),
        type: block.getType(),
        text: block.getText(),
      }))
      .toJS(),
```
The snapshot is the same.

so the reason why before "key0" and "key4" were in the end is because "k" is after "f" and when we convert from orderedMap with the .toJS() it does not preserve insertion order, that why we update the assertions to use ` .toIndexedSeq()` before converting to JS:

```
      .getBlockMap()
      .toIndexedSeq()
      .toJS(),

```

src/model/modifier/__tests__/__snapshots__/AtomicBlockUtils-test.js.snap
mitermayer: If you look above on the snapshot it is actually there, it is now appearing in the correct place after the block 'a':

```
+Array [
-  "a": Object {
+  Object {
+    "characterList": Array [],
+    "data": Object {},
+    "depth": 0,
     "key": "a",
     "key": "a",
     "text": "",
     "text": "",
     "type": "unstyled",
     "type": "unstyled",
   },
   },
-  "b": Object {
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": "2",
+        "style": Array [],
+      },
+    ],
+    "data": Object {},
+    "depth": 0,
+    "key": "key0",
+    "text": " ",
+    "type": "atomic",
+  },
+  Object {
+    "characterList": Array [
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+      Object {
+        "entity": null,
+        "style": Array [],
+      },
+    ],
+    "data": Object {},
+    "depth": 0,
+    "key": "key4",
+    "text": "Alpha",
+    "type": "unstyled",
+  },
```

flarnie:
dancing_taco

Issue #1383, opened 2017-09-11
After drag-n-drop onChange will be broken


ukrbublik:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
After drag-n-drop of some text (internal, in 1 editor) 'select' events will not be handled by DraftJs: code wil not enter in editOnSelect() and thus onChange will NOT be triggered. UNTIL user clicks editor.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
Reproducing is simple:
1. Take any example, like examples/draft-0-10-0/plaintext/plaintext.html
2. Add `console.log(1);` into start of `function editOnSelect(editor) {..}`
3. Run example. Check console to see that select events are working. 
Then drag-n-drop some text and there will be no '1' in console any more.
Until user clicks editor.

**What is the expected behavior?**
Drag-n-drop should not affect on 'select' events and 'onChange'.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
v0.10.1
Chrome 58.0.3029.110
Ubuntu 16.04

ukrbublik:
Related to issue in React:
https://github.com/facebook/react/issues/11379

PR #1599, opened 2018-01-05
Fixes some drag-n-drop problems

ukrbublik:
**Summary**
Fixes issue #1383. Fixes problem which is: onDrop breaks onSelect/onChange events.
Fixes issue #1454 (partial). Fixes problem which is: after first internal drop dragged text will not be moved to target position, but copied.
The solution is to set `editor._internalDrag` to false not during onDragOver events, but after end of drag-n-drop.

**Test Plan**
Attached reproducable videos (on Chrome for Linux)
Issue #1383 before and after:
1383-before
1383-after
Issue #1454 before and after:
1454-before
1454-after


ukrbublik:
> after first internal drop dragged text will not be moved to target position, but copied
This bug is present on Chrome on Linux, bit not on Windows

niveditc:
@ukrbublik, next steps are that:
* Provided the steps above are successful I'll merge it in tomorrow (Monday). In the slim chance that something breaks it's better to do a risky merge on a weekday than a weekend.

Issue #1591, opened 2017-12-31
Get the inline styles of SelectionState? (and apply them to Modifier.replaceText)


omerdn1:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
I haven't seen a way to get the styles of the SelectionState. This is necessary for `Modifier.replaceText`

**What is the expected behavior?**
Replacing text with `Modifier.replaceText` should have an option to keep the styles of the SelectionState.

koutsenko:
So rude..

Issue #602, opened 2016-08-18
Is there a way to get the inline styles of current selection?


ricardometring:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
I haven't seen a way to get the styles of the current selection

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

**What is the expected behavior?**
Show me a list of the styles from the editorState current selection

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

namKolo:
With current selection you can get `startKey`, `endKey` . Base on those keys, you can get all blocks in that selection.
With each block, you  can use `getCharacterList` function to get all character in that block . Then with each character, use `getStyle` it will return style for you
Concat all of them, you can get style of current selection
Example code
`block.getCharacterList().reduce(
(styles,c) => styles.union(c.getStyle()),
Immutable.OrderedSet()
)`
Hope it will help you

thibaudcolas:
[`draftjs-utils`](https://github.com/jpuri/draftjs-utils) has a `getSelectionCustomInlineStyle` function.

markusenglund:
To anyone coming here from google - there does exist a built in getter method called `getCurrentInlineStyle` on the editorState object.

So if you for example want to know if the current selection is "BOLD" you could do this:

```js
const { editorState } = this.state;
const inlineStyle = editorState.getCurrentInlineStyle();
const isBold = inlineStyle.has("BOLD");
```

I hope that helps someone!

promerorgz:
@palamccc couldnt you get the focus first and pass it into the `getCurrentInlineStyle()` function?

```javascript
const currentFocus = editor.getSelection().getFocusKey()
    const inlineStyle = editor.getCurrentInlineStyle(currentFocus);
    const isBold = inlineStyle.has("BOLD")
```

palamccc:
@promerorgz, thanks for the comment, we have moved our project to https://github.com/ianstormtaylor/slate I can't confirm whether your solution works or now.

jtabone16:
@promerorgz worked like a charm, thanks!

dabretin:
If someone always search for get all the inline styles :
```javascript
const selection 	= editorState.getSelection();
const contentState	= editorState.getCurrentContent();
const styles		= new Set<string>();
[...]

PR #1600, opened 2018-01-05
Feature - get current common and union inline style for selection

ukrbublik:
**Summary**
Added 2 methods to `EditorState` - `getCurrentCommonInlineStyle()` and `getCurrentUnionInlineStyle()` to replace `getCurrentInlineStyle()`.
For non-collapsed selection `getCurrentInlineStyle()` returns just style of first character, which is confusing and probably not correct.
`getCurrentCommonInlineStyle()` will return intersection of all inline styles found in non-collapsed selection.
`getCurrentUnionInlineStyle()` will return union of all inline styles found in non-collapsed selection.

**Test Plan**
Added tests.


flarnie:
Hi @ukrbublik - I see you've opened several PRs, thanks for that!

Right now we likely won't be accepting new features or APIs for Draft.js, since the framework already has a large surface area. I'll look at these in more detail over the coming weeks, but up front I want to say that we likely won't merge them.

It is helpful to know if this is related to any open issue, and what the use case is for these new methods. I think the correctness of the current version of `getCurrentInlineStyle` depends on how you are using it.

ukrbublik:
Hi!
It's not related to specific issue as far as I know (I haven't searched). Related to #1591, #602.

Use case - to show correct info about styles of non-collapsed selection at editor toolbar.
For example, at [draft.js](draft.js) in editor we have text "**11**_22_" (11 - Bold, 22 - Italic). Select this text and you'll see Bold selected at toolbar. It's not correct because "22" is not bold. 
Common behaviour to text processors is to show union styles (in this case - no styles) (but if we have selected text "**_11_**_22_" - there should be just Italic selected at toolbar).

niveditc:
Same comment as @flarnie - I don't think this belongs in the core API & can be implemented a the individual editor level if needed.

PR #1601, opened 2018-01-05
Feature - add Modifier.cutBySelection()

ukrbublik:
**Summary**
Added new `DraftModifier` method - `cutBySelection(contentState, selection)` which cuts part of content state by selection and returns new content state (preserving original block keys).

**Test Plan**
No tests yet.


flarnie:
This looks similar to [getContentStateFragment](https://github.com/facebook/draft-js/blob/ab00a2df7477e44d79a7f4d63941bf969f0a241a/src/model/transaction/getContentStateFragment.js)

That's a helper we use under the hood in the other transactions - I don't think it's documented, so not really fair to expect you or anyone else to know about it.

We are trying to reduce the size of Draft.js, so I don't want to add new transactions now, but I'm open to adding docs about `getContentStateFragment` and making people aware of that helper.

You could also gather some of your new features/helper methods into a separate repo, and we could even link to it in the docs. Also I've noticed a repo called https://github.com/thibaudcolas/draftjs-filters and I wonder if @thibaudcolas would be open to hosting more helper methods for manipulating Draft.js EditorState.

thibaudcolas:
For now the Draft.js filters repo only contains utilities related to copy/pasting of content, and I think I'd rather keep it focused on that – there are other "general Draft.js utilities" packages like https://github.com/jpuri/draftjs-utils.

That said, if there is some agreement it would be helpful, I'd be very happy to contribute to creating another "Draft.js reusable utils" repo. My editor's code has [a few other](https://github.com/springload/draftail/blob/master/lib/api/behavior.js) [functions](https://github.com/springload/draftail/blob/master/lib/api/DraftUtils.js) I think could be useful to others.

Coming back to the filters repo, I invested quite a bit in its tooling, demo site, and testing (there's only ±300 lines of actual code in there, 90% of it is actually the tests), so I think it would be a great base for any future package of the same sort.

PR #1602, opened 2018-01-05
Feature - insert additional atomic block param for block data

bradlocking:
**Summary**
Added a param to `AtomicBlockUtils` - `insertAtomicBlock()`. When creating an Atomic block, there isn't an option to create with block data. This has been added as an optional param, which is then passed to the `new ContentBlock` within `fragmentArray`. 

**Test Plan**
`assertInsertAtomicBlock` has been updated to contain the blockData param, which for test purposes is an empty object.


ddelrio1986:
I see a test and the blockData parameter added but I don't see where the blockData parameter is ever used?

Issue #1434, opened 2017-10-14
Bug: Cannot read property 'getCharacterList' of undefined


seejee:
Hi DraftJS team! Thanks for all your hard work on DraftJS!

I've been able to reliably reproduce an drag/drop exception using DraftJS 0.10.2. It seems that if you dry and drag a selection that spans multiple blocks to the end of last paragraph, and exception will occur. Here's a GIF of the issue:

![draft-drag-drop](https://user-images.githubusercontent.com/553936/31575509-a0581948-b0b6-11e7-8871-3eb2de1dc609.gif)

To reproduce:

- Write two paragraphs of text.
- Select from the end of the first paragraph to somewhere in the second paragraph (in other words, the selection should span the blocks)
- Try and drag that selection to the end of the second paragraph.

What happens:

The text does not move, and this unhandled exception appears in the console:

```Draft.js:5579 Uncaught TypeError: Cannot read property 'getCharacterList' of undefined
    at removeForBlock (Draft.js:5579)
    at removeEntitiesAtEdges (Draft.js:5534)
    at Object.replaceWithFragment (Draft.js:847)
    at Object.moveText (Draft.js:843)
    at moveText (Draft.js:7447)
    at onDrop (Draft.js:7437)
    at Draft.js:6579
    at Object.ReactErrorUtils.invokeGuardedCallback (react.js:10560)
    at executeDispatch (react.js:2862)
    at Object.executeDispatchesInOrder (react.js:2885)
```

What should happen:

The text should be moved to the end of the second paragraph.

ondrejbartas:
@seejee Hi, I am having same problem and it seems that I found why.

DraftModifier.moveText is not looking what should be done first. Basically there are two steps that need to happen,

- remove text from source selection
- insert text to destination selection

Those two steps are done but based on position of source/destination selection those steps needs to reorder.

text `source` something `destination` end of sentence.
then you need to insert first and then remove

but if 

text `destination` something `source` end of sentence.
then you need to remove first and then insert.

PR #1603, opened 2018-01-06
Fixes drag-n-drop problems

ukrbublik:
**Summary**
Fixed issue #1434
Fixed issue #1454
Also added util methods for SelectionsState - compareWithSelection(), updateOnDeletingSelection()

**Test Plan**
No tests. Don't know how to make tests for drag-n-drop events.

### DIFF
--- a/src/component/handlers/drag/DraftEditorDragHandler.js
+++ b/src/component/handlers/drag/DraftEditorDragHandler.js
@@ -133,7 +133,16 @@ function moveText(
     editorState.getSelection(),
     targetSelection,
   );
-  return EditorState.push(editorState, newContentState, 'insert-fragment');
+  editorState = EditorState.push(
+    editorState,
+    newContentState,
+    'insert-fragment',
+  );
+  editorState = EditorState.forceSelection(
+    editorState,
+    newContentState.getSelectionAfter(),
+  );
+  return editorState;
 }
 
 /**
--- a/src/model/immutable/SelectionState.js
+++ b/src/model/immutable/SelectionState.js
@@ -13,6 +13,20 @@
 
 'use strict';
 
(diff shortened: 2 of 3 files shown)


ukrbublik:
> Can you explain how this relates to #1599? Both of them seem to have some part of a fix for #1454.
#1599 fixes this bug:
https://github.com/facebook/draft-js/issues/1454#issuecomment-339360755
#1603 fixes this bugs:
https://github.com/facebook/draft-js/issues/1454#issuecomment-340973151
https://github.com/facebook/draft-js/issues/1454#issuecomment-348872969 (also described at #1434)

> This needs a test plan - can you either add test cases or a video with the bug & fix.
Added unit test. Currently it fails on master (throws exception) and passes after this fix.

I'll explain logic soon..

ukrbublik:
@niveditc
Cleaned up commit, added some comments. Also see unit test  `DraftEditorDragHandler.test.js`.

SelectionsCompareResult is result of comparison of 2 selections. I added general purpose method (helper) SelectionState.compareWithSelection(). 
I use it to compare position in content state with selection to be deleted - at comparePositionWithSelection(), which is used at fixPosOnDeletingSelection(), which is used at SelectionState.updateOnDeletingSelection().
Last one is essential to get correct position in udpated content state after we removed text (dragged off), this position should be used to insert (drop) text then.
Without this DraftModifier.moveText() works incorrect.

sandorvasas:
We are receiving "Cannot read property 'getCharacterList' of undefined".

Why hasn't this been merged since january?

mrkev:
@thread huh, this fell through the cracks. Just confirmed this is still an issue. I need to read more and get familiar with this though, but I'll have a look if I have a spare minute during the holidays.

PR #1605, opened 2018-01-08
[wip] Preserve list indentation from one DraftEditor to another on paste

GordyD:
**Summary**

When pasting from one editor to another we lose list indentation.  By inspecting the classList of each list item contained in HTML pasted we can preserve this state.

**Test Plan**
I validated behavior manually:

- Rebuild
- Open two tabs to rich text example
- In one create indented list
- Select all
- Paste into other
- Styling is preserved

I also added some snapshot testing for regression testing.


src/model/encoding/__tests__/convertFromHTMLToContentBlocks-test.js
```diff
@@ -279,3 +279,33 @@ test('Should preserve entities for whitespace-only content', () => {
...
+      <li class="public-DraftStyleDefault-depth3">depth3</li>
+      <li class="public-DraftStyleDefault-depth4">depth4</li>
+    </ul>
+  `;
+  assertConvertFromHTMLToContentBlocks(html_string, {
+    experimentalTreeDataSupport: false,
+  });
+});
+
+test('Should import recognised draft li depths when nesting enabled', () => {
```
flarnie: This is almost right - but I have one concern.

With "nesting enabled", the new features should create a block map where the "depth" property is not used, and instead we have true nesting set up using 'children' and 'parent' properties on the blocks.

See https://github.com/facebook/draft-js/blob/master/src/model/encoding/__tests__/__snapshots__/convertFromHTMLToContentBlocks-test.js.snap#L303 for an example.

It looks like with this change, we default to using the 'depth' property again for nesting, even with the 'nesting' flag enabled, when those classes are present.

I think we should add a conditional for your check that sets the 'depth' and disable it when the nesting flag is true.

GordyD: Ok - thanks for the pointer. So in the instance that `nesting` is `true` should there be an alternative way to handle converting the li indentation?

src/model/encoding/convertFromHTMLToContentBlocks.js
```diff
@@ -84,6 +85,14 @@ const inlineTags = {
   u: 'UNDERLINE',
 };
 
+const knownListItemDepthClasses = {
+  [cx('public/DraftStyleDefault/depth0')]: 0,
+  [cx('public/DraftStyleDefault/depth1')]: 1,
+  [cx('public/DraftStyleDefault/depth2')]: 2,
+  [cx('public/DraftStyleDefault/depth3')]: 3,
+  [cx('public/DraftStyleDefault/depth4')]: 4,
+};
```
thibaudcolas: From what I understand, list indentations above 4 would not be preserved. Should this be changed to support arbitrary depth?

I imagine the current approach is based on the class names defined in https://github.com/facebook/draft-js/blob/232791a4e92d94a52c869f853f9869367bdabdac/src/component/contents/DraftEditorContents-core.react.js#L58-L62. An alternative would be to generate those class names dynamically with the right depth suffix in `getListItemClasses`, and here in `getListItemDepth` get the depth by looking for a number right after the `cx('public/DraftStyleDefault/depth')` string.

I realise this is a bit more work but it would be nice to support all use cases of depth. While the `getListItemClasses` code does stop at depth 4 at the moment, this is easily extensible via `blockStyleFn`, whereas the conversion from HTML isn't.

flarnie: Hi @thibaudcolas - 
We are working on replacing this approach of using CSS classes with true nested blocks, in a longer term project. So for now we're not adding improvements to the old CSS code, but ultimately will be able to support arbitrary nesting. That is indeed a good idea. :)

I would not block this diff on that change either way.

flarnie:
Created an internal task to have someone enable this in tree-data-structure mode.

PR #1608, opened 2018-01-09
Feature tree/abstract tree wrapping helper

juliankrispel:
### Context

**Extract tree node wrapper methods**

This PR moves the logic for wrapping tree nodes into a separate helper file and replaces a duplicate wrapping effort with said helper (inside `DraftEditorContentsExperimental`)

***

**Note:**
production systems.

- [x] extracted applyWrapperElementToSiblings
- [x] extracted shouldNotAddWrapperElement
- [x] replace custom block rendering in DraftEditorContentsExperimental with applyWrapperElementToSiblings
- [x] move all logic related to wrapping tree nodes into `wrapBlockNodes`
- [x] add snapshots for wrapBlockNodes


docs/APIReference-Editor.md
```diff
@@ -144,25 +144,25 @@ or if you just want to display content for a static use case.
 
 Default is `false`.
 
-#### spellCheck
```
mitermayer: Was this change supposed to be part of this PR or is it a bad merge ?

juliankrispel: whoops, will fix

src/component/utils/exploration/__tests__/__snapshots__/wrapBlockNodes.js.snap
```diff
@@ -0,0 +1,595 @@
...
+          </span>
+        </div>
+      </li>
+    </ol>
+  </div>
+</blockquote>
+`;
+
+exports[`renders block with nested child that have different wrapperElement 1`] = `
+<blockquote
```
flarnie: This output only seems to show 
```
<blockquote (A)> 
  <ol (c)>
  </ol>
</blockquote>
```

What about 'B', the `ul`? I would expect it to look like this:

```
<blockquote (A)> 
  <ul (B)>
    <ol (C)>
    </ol>
  </ul>
</blockquote>
```

src/component/utils/exploration/__tests__/wrapBlockNodes.js
```diff
@@ -0,0 +1,580 @@
...
+    new ContentBlockNode({
+      parent: 'B',
+      key: 'C',
+      text: 'header inside list',
+      type: 'header-one',
+    }),
+  ]);
+
+  const DecoratedComponent = props => {
+    return <span className={'custom-decorated-text'}>{props.children}</span>;
```
flarnie: nit - I think it can be `className='custom-decorated-text'`, no need for the wrapper curly braces if it's a string.

flarnie: Also it appears that the decorator is never applied in the snapshot output - thought maybe I was missing it, but when I search for 'custom-decorated-text' on this page the only result is here.

src/component/utils/exploration/__tests__/__snapshots__/wrapBlockNodes.js.snap
```diff
@@ -0,0 +1,595 @@
...
+        data-block={true}
+        data-editor="editor"
+        data-offset-key="B-0-0"
+      >
+        <div
+          className="public__DraftStyleDefault__block public__DraftStyleDefault__ltr"
+          data-offset-key="B-0-0"
+        >
+          <h1
+            contentEditable={false}
```
flarnie: Do we expect this to be `false`?

src/component/utils/exploration/__tests__/wrapBlockNodes.js
```diff
@@ -0,0 +1,580 @@
...
+    block: rootBlock,
+    blockStyleFn: block => {
+      if (block.getType() === 'header-one') {
+        return 'My-fancy-custom-class';
+      }
+      return null;
+    },
+    tree: BlockTree.generate(contentState, rootBlock, null),
+  });
+});
```
flarnie: Really glad you added all these tests! ^_^

Would it make sense to add one with multiple sibling children at the same level? These are mostly [A -> [B -> [C]]], rather than [A -> [B, C]]
Like a List of children which all have type 'ordered-list-item' would, I expect, get wrapped in the same parent `ol`.

flarnie:
In general I like helping with rebasing, fixing nits, etc. and if you are able to check the 'Allow edits from maintainers' button I think I should be able to push to your PR branches. Not sure how to update that after the PR is created, but I imagine there is a way.

pls_let_me_help_fix_lints_in_your_prs

flarnie:
I'm going to try importing this and updating it internally.

flarnie:
Internally I fixed some lints and removed the mistaken docs update, but there are still flow errors and possibly some problems with the test results.

It's hard to tell whether the test snapshots are correct or not, since this behavior was basically untested before this diff. And that's also a problem - this behavior is untested.

When I have more time, or if someone else has more time, I'd like to see tests similar to the `wrapBlockNodes` test but for `DraftEditorContentsExperimental` and `DraftEditorBlockNode`. We could pass them different props and use `ReactTestRenderer` to verify that the output makes sense.

That should be written and run before and after these changes, and then we can at least see whether this changes the way things work. It also may be surfacing bugs, if the inconsistencies I noted are present in the logic before this refactor.

Sorry it's been slow to get this landed - I think the next time I have a couple of hours I will be able to wrap this up. If @juliankrispel or @mitermayer or anyone else has time to work on it before then, please post here to let me know.

flarnie:
Also for context, since I have to review this every time I look at this PR:

The reason for this change is to dedup some duplicated logic that shows up in both `DraftEditorContentsExperimental` and `DraftEditorBlockNode`.
In `DraftE...Experimental` we are rendering the array of `BlockNodeRecords`, and each one has a `wrapperTemplate` determined by the type of block it is and the corresponding element in the `BlockRenderMap`. If contiguous blocks have the same `wrapperTemplate` they get wrapped in a wrapper element of the type specified in the `BlockRenderMap`.

Inside `DraftEditorBlockNode` we also do this same process for the array of child `BlockNodeRecord`s that are rendered for that BlockNode.

The wrapping logic is being pulled out into `wrapBlockNodes`, and that's why we add that helper module and a unit test for it.

PR #1613, opened 2018-01-14
Fix render crash of DraftEditorTextNode (BR -> text)

### LINKED ISSUE #1612: Bug - DraftEditorTextNode render crash (when BR -> text)
Bug

Can be reproduced pretty easily on draftjs.org.
peek 2018-01-14 03-09
Create 10+ empty lines. Start typying chars at first line and simultaneously try to select by mouse those 
empty lines (by clicking from top to down).
You will notice errors in console: "Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."
React 16 will point on faulty component: 
```The above error occurred in the <br> component:
    in br
    in DraftEditorTextNode```

No rendering crash :)

0.10.4
Chorome 58, Ubuntu 16

### PR DESCRIPTION
**Pull request by ukrbublik**
**Summary**
Fixes #1612
Native rendering must be prevented when <BR> should be replaced by <SPAN>.
By obvious reason. Otherwise React-rendering of DraftEditorTextNode will crash!

Tip: Also this bug can be prevented by using `const useNewlineChar = true` at DraftEditorTextNode (now it's true for IE only)

**Test Plan**
No unit tests. Just did manual ones.

### DIFF
--- a/src/component/handlers/edit/editOnBeforeInput.js
+++ b/src/component/handlers/edit/editOnBeforeInput.js
@@ -185,6 +185,17 @@ function editOnBeforeInput(
         parentNode.firstChild.nodeType === Node.TEXT_NODE &&
         parentNode.firstChild.nodeValue.indexOf('\t') !== -1;
     }
+    // If selected block was empty, DraftEditorTextNode rendered it as <BR>.
+    // But now we typed some chars to it, so <BR> should be removed.
+    // We must let React render this properly.
+    if (
+      nativeSelection.anchorNode &&
+      nativeSelection.anchorNode.nodeName === 'SPAN' &&
+      nativeSelection.anchorNode.firstChild.nodeName == 'BR' &&
+      chars
+    ) {
+      mustPreventNative = true;
+    }
   }
   if (!mustPreventNative) {
     // Check the old and new "fingerprints" of the current block to determine

### CODE REVIEW

**Comment by fazouane-marouane** [comment]
Hello @flarnie  👋
I don't know if you've had the time to review this PR? Does it need something to be mergeable?

Marouane

**Comment by ukrbublik** [comment]
Hi @niveditc
Can you review this PR?

File: src/component/handlers/edit/editOnBeforeInput.js [hunk 1]
```diff
@@ -185,6 +185,17 @@ function editOnBeforeInput(
...
         parentNode.firstChild.nodeValue.indexOf('\t') !== -1;
     }
+    // If selected block was empty, DraftEditorTextNode rendered it as <BR>.
+    // But now we typed some chars to it, so <BR> should be removed.
+    // We must let React render this properly.
+    if (
+      nativeSelection.anchorNode &&
+      nativeSelection.anchorNode.nodeName === 'SPAN' &&
+      nativeSelection.anchorNode.firstChild.nodeName == 'BR' &&
+      chars
```
sophiebits: `chars` check is unnecessary (it's already checked on line 97)

also… does it make sense to check that `.firstChild === .lastChild` too?

src/component/handlers/edit/editOnBeforeInput.js
```diff
@@ -185,6 +185,17 @@ function editOnBeforeInput(
         parentNode.firstChild.nodeType === Node.TEXT_NODE &&
         parentNode.firstChild.nodeValue.indexOf('\t') !== -1;
     }
+    // If selected block was empty, DraftEditorTextNode rendered it as <BR>.
+    // But now we typed some chars to it, so <BR> should be removed.
+    // We must let React render this properly.
+    if (
```
sophiebits: all this added code should be in a !mustPreventNative block for consistency with the others

ukrbublik:
@sophiebits 
> Why isn't the isSelectionAtLeafStart check sufficient? It should already be catching this case.

`nativeSelection` is ahead of selection from editor state used at `isSelectionAtLeafStart()`
That's why it don't catch this case

sophiebits:
How can this be reproduced? I assume you aren't using React's unstable concurrent mode so I would expect _latestCommittedEditorState to always match the latest native selection.

ukrbublik:
Offtop:
`babel-eslint` v7 used curretly gives me error on `lint` command:
`TypeError: Cannot read property 'type' of undefined`
Upgrading to v8 solves problem

ukrbublik:
@sophiebits 
> How can this be reproduced?
I described steps here: https://github.com/facebook/draft-js/issues/1612
The point is to print text from keyboard + click on empty lines simultaneously. 

> I assume you aren't using React's unstable concurrent mode
I don't. 

I tried to print to console info about `nativeSelection` and selection from `_latestCommittedEditorState`. And anchor keys were different
`console.log( anchorNode.getAttribute('data-offset-key'), editor._latestCommittedEditorState.getSelection().getAnchorKey() )`

sophiebits:
When doing what action? Is there anything special about what triggers this? Does it happen every time for you? Only in certain browsers or in multiple browsers?

ukrbublik:
@sophiebits 
> When doing what action? Is there anything special about what triggers this?
During normal use I've seen this error occasionally. 
But I've figured out it can be 100% reproduced by repeatedly clicking mouse on empty lines while typing text.

Try it on Mac: set short "Delay Until Repeat" at Keyboard Settings at System Preferences, goto draftjs.org, create 10+ empty lines, goto first line, hold '=' key for continuous typing, start clicking on empty lines randomly (and quickly), then see errors in console.

> Does it happen every time for you? Only in certain browsers or in multiple browsers?
Definitely happens in Chrome on Ubuntu/OSX, Safari on OSX. Can't reproduce in Firefox. Haven't tried in IE.

einsteinreloaded (approved):

Issue #1621, opened 2018-01-21
ContentBlock. Cannot implement


steida:
v0.10.5

Monitor logs will go to /private/tmp/flow/zSUserszSsteidazSdevzSeste.monitor_log
Error: node_modules/draft-js/lib/ContentBlock.js.flow:54
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
           ^^^^^^^^^^^^ ContentBlock. Cannot implement
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                              ^^^^^^^^^ BlockNode
  Property `findEntityRanges` is incompatible:
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
               ^^^^^^^^^^^^ ContentBlock. Covariant property `findEntityRanges` incompatible with invariant use in
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                                  ^^^^^^^^^ BlockNode

Error: node_modules/draft-js/lib/ContentBlock.js.flow:54
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
           ^^^^^^^^^^^^ ContentBlock. Cannot implement
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                              ^^^^^^^^^ BlockNode
  Property `findStyleRanges` is incompatible:
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
               ^^^^^^^^^^^^ ContentBlock. Covariant property `findStyleRanges` incompatible with invariant use in
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                                  ^^^^^^^^^ BlockNode

Error: node_modules/draft-js/lib/ContentBlock.js.flow:54
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
           ^^^^^^^^^^^^ ContentBlock. Cannot implement
 54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                              ^^^^^^^^^ BlockNode
  Property `getCharacterList` is incompatible:
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
               ^^^^^^^^^^^^ ContentBlock. Covariant property `getCharacterList` incompatible with invariant use in
     54: class ContentBlock extends ContentBlockRecord implements BlockNode {
                                                                  ^^^^^^^^^ BlockNode

Error: node_modules/draft-js/lib/ContentBlock.js.flow:54
[...]

PR #1628, opened 2018-01-26
Bumping internal flow version

mitermayer:
This is first step In order to investigate issues reported by https://github.com/facebook/draft-js/issues/1621, this PR will bump our internal flow dep to latest released version.


mitermayer:
@flarnie , the error we get is actually the same as in master, master is currently broken. submitting a separate PR to fix it.

https://github.com/facebook/draft-js/pull/1631

PR #1637, opened 2018-01-31
Deprecated the coarse `onArrowUp` et al key handler props on `DraftEditor.react` to make it possible to produce editor commands from these keys

steveluscher:
Previously, it was impossible to do this:

```
function mapKeysToEditorCommands(e) {
  if (mentionsAutocompleteIsOpen) {
    switch (e.keyCode) {
      case Keys.UP:
        return 'MentionsAutocomplete/select-previous';
      case Keys.DOWN:
        return 'MentionsAutocomplete/select-next';
    }
  }
  return getDefaultKeyBinding(e);
}

<DraftEditor
  keyBindingFn={mapKeysToEditorCommands}
  ...
/>
```

…because we didn't give TAB, ESCAPE, LEFT, UP, RIGHT, or DOWN a chance to produce a command name.

This pull request:

1. deprecates the prop-based key handlers (`onUpArrow` et al.) and encourages people to move their custom key logic into a `keyBindingFn`, and
2. lets the named keypresses fall through to the `keyBindingFn` whenever a prop-based key handler isn't supplied.
3. Updates the homepage and the rich text example.

tabs

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -262,45 +262,22 @@ Handle other drop operations.
 
 ### Key Handlers (Optional)
 
-These prop functions expose common useful key events. Example: At Facebook, these are
-used to provide keyboard interaction for mention results in inputs.
+Draft lets you supply a custom `keyDown` handler that wraps or overrides its
+default one.
 
-#### onEscape
-```
-onEscape?: (e: SyntheticKeyboardEvent) => void
-```
-
-#### onTab
-```
-onTab?: (e: SyntheticKeyboardEvent) => void
-```
-
-#### onUpArrow
-```
-onUpArrow?: (e: SyntheticKeyboardEvent) => void
-```
-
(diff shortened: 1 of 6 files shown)


steveluscher:
I updated the examples and the homepage, but I wasn't really sure how to update `RichUtils.onTab()`. I tried to split it into `onTabLevelIncrement` and `onTabLevelDecrement` functions that don't rely on the keyboard event, but that makes it really easy to miss a case (eg. you want to `preventDefault` iff you're at a place in the text that even supports tab indenting). I ended up just leaving it the way it was.

flarnie:
It's weird that @steveluscher it looks like you did land this, but the PR was not closed by facebook-github-bot.

I'm closing this, if I'm missing something then please open a new PR and we can land it separately.

I also opened an internal task to look into why the PR wasn't closed when the imported change landed on master. T25937659

PR #1644, opened 2018-02-09
Bundle size stats + Misc changes

altano:
**Summary**

Misc changes that:
1. Make the dev environment work on Windows and Node v9.
1. Upgrade Webpack and UglifyJS.
1. Build now outputs build stats to the `meta/bundle-size-stats directory/` dir.
    1. The build stats are named the current package version, e.g. the current one is `version-0.10.5.json`.
    1. An example of this file can be seen here: https://files.terriblefish.com/version-0.10.5.json
    1. This json file can be visualized in any of the various tools online, e.g. [the official tool](https://webpack.github.io/analyse) which is great for visualizing module dependencies and [this one from Chris Bateman](http://chrisbateman.github.io/webpack-visualizer/) that makes visualizing file sizes really nice.

**Test Plan**

`npm test`, `npm run lint`, and `npm run flow` all pass

**Misc**
* I don't think any of these changes need tests
* No API changes

PR #1653, opened 2018-02-16
Move uglifyjs-webpack-plugin to devDependencies

thibaudcolas:
**Summary**

This dependency got introduced in a separate commit (f8ca29d1a7fa5a8bb89cb9be9a66d12732a0bfa7) from the others in #1644. I imagine this is a mistake since it only provides dev-time tooling.

**Test Plan**

For development, we can try a fresh install of the project:

1. `rm -rf node_modules`
2. `yarn install`. This runs `npm run build` via the `prepublish` script, so if it works it's good to go.

I don't think dependencies and devDependencies are processed any differently in a development environment, so not sure this check is even necessary.

For the published package:

1. `mkdir test`
2. `npm init -y`
3. `npm install git+https://github.com/thibaudcolas/draft-js.git#patch-2`
4. `npm ls --depth=1`

The `npm ls` result should be:

```
├─┬ draft-js@0.10.5 (git+https://github.com/thibaudcolas/draft-js.git#cb06920d0f3dea77692680dddba6ec475663c7b2)
│ ├── fbjs@0.8.16
│ ├── immutable@3.7.6
│ └── object-assign@4.1.1
├── UNMET PEER DEPENDENCY react@^0.14.0 || ^15.0.0-rc || ^16.0.0-rc || ^16.0.0
└── UNMET PEER DEPENDENCY react-dom@^0.14.0 || ^15.0.0-rc || ^16.0.0-rc || ^16.0.0
```

Note `uglifyjs-webpack-plugin` is absent, whereas it would have been there with `npm install git+https://github.com/thibaudcolas/draft-js.git#master`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,7 @@
   "dependencies": {
     "fbjs": "^0.8.15",
     "immutable": "~3.7.4",
-    "object-assign": "^4.1.0",
-    "uglifyjs-webpack-plugin": "^1.1.6"
+    "object-assign": "^4.1.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0-rc || ^16.0.0-rc || ^16.0.0",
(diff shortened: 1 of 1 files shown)


mitermayer:
Thanks for this PR!

PR #1664, opened 2018-02-24
Website: Fixed code highlight

marcelometal:
Before:

screenshot from 2018-02-24 20-40-31

After:

screenshot from 2018-02-24 20-40-42


flarnie:
That looks much nicer! :)

Can you give me a bit more context on why the JS files for the website changed? I didn't build the original version, but it looks like they had a local copy of 'prismjs' and we are replacing that with the npm package for 'prismjs'. For 'marked' it's less clear to me what the change does. ?

marcelometal:
Hi @flarnie, 
the same for marked... I just remove the local copy of marked and replace it with npm package. I've also added marked highlight support.

PR #1677, opened 2018-03-05
Enabling a GK manager for the plaground

mitermayer:
This enables on the playground for us to pass `gk_enable=foo,bar,foobar` or `gk_disable=foo,bar,foobar`, on the URI, 

Example: 

```
http://localhost:3000/?gk_enable=foo&gk_disable=bar,foobar
```

***

TLDR:

- Updating our Gkx stub to treat undefined gks on the stub object as falsy
- This will allow us to update GK flags while interacting with the playground.

### DIFF
--- a/examples/draft-0-10-0/playground/src/GkManager.js
+++ b/examples/draft-0-10-0/playground/src/GkManager.js
@@ -1,0 +1,19 @@
+import querystring from 'querystring';
+
+const QUERY_STRINGS = querystring.parse(window.location.search.substring(1));
+
+// overwrite the feature flag stub object
+const GKManager = (window.__DRAFT_GKX = {});
+
+// enable or disable feature flags
+const flagControls = {
+  gk_disable: flags =>
+    flags.forEach(flag => (window.__DRAFT_GKX[flag] = false)),
+  gk_enable: flags => flags.forEach(flag => (window.__DRAFT_GKX[flag] = true)),
+};
+
+Object.keys(flagControls)
+  .filter(flag => QUERY_STRINGS[flag])
+  .forEach(flag => flagControls[flag](QUERY_STRINGS[flag].split(',')));
+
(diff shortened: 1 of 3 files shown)


examples/draft-0-10-0/playground/src/GkManager.js
```diff
@@ -0,0 +1,19 @@
...
+// enable or disable feature flags
+const flagControls = {
+  gk_disable: flags =>
+    flags.forEach(flag => (window.__DRAFT_GKX[flag] = false)),
+  gk_enable: flags => flags.forEach(flag => (window.__DRAFT_GKX[flag] = true)),
+};
+
+Object.keys(flagControls)
+  .filter(flag => QUERY_STRINGS[flag])
+  .forEach(flag => flagControls[flag](QUERY_STRINGS[flag].split(',')));
```
flarnie: Love the concise implementation :)

examples/draft-0-10-0/playground/src/index.js
```diff
@@ -1,8 +1,11 @@
+import GkManager from './GkManager';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
 import App from './App';
 import registerServiceWorker from './registerServiceWorker';
 
+console.log("Applying feature flag overwrites: ", GkManager);
```
flarnie: ?

mitermayer: Was thinking about logging the GKs that are applied during the playground run for debugging purposes, happy to remove this. What are your thoughts around it ?

mitermayer: The value for GkManager is "window.__DRAFT_GKX", so I guess we could remove this line and relly on users checking that variable instead. What do you think ?

PR #1679, opened 2018-03-05
Added data field to ContentState record.

mzedeler:
**Summary**

This adds a `data` field to `ContentState` making it possible to keep metadata structures in sync with the content state.

**Motivation**
Since draft is built around static transformation functions that generate new ContentState objects, there isn't any easy way to record metadata for a given ContentState instance and ensure that it isn't lost in any of the transformation functions.

If the transformation methods were written to reuse the class of the provided ContentState, it would be possible to extend the ContentState record without any further changes to draft, but this isn't possible as it works now. Adding a `data` field will provide a generic and future proof way of keeping metadata that is in sync with the content state.

**Test Plan**

The editor should be working as before. I've added an automated test to verify that the `data` field is there.


flarnie:
Can you give some specific use cases that are enabled by this change?

My thought is that any 'metadata' related to the editor state can be kept in the state of the parent component, in a separate field from the 'editorState' or 'contentState'. I also have a general preference for small API and strict type definitions, would rather not add a new field that has no type restrictions or specific use.

mzedeler:
We've already headed down this road by adding data fields in various other places and the stuff I am setting up is very tightly coupled to the editor state, so every place the content changes, I'll have to carry around two states that carefully have to be kept in sync. I have thought about the approach and there is a large number of places in my codebase I'll have to change if I do this.

mzedeler:
The specific use case is that I am making it possible to edit XML documents where some elements are rendered as entities including nesting. The tree structure of the underlying XML document needs to be stored somewhere and kept in sync with the rest of the content, since edit operations to the content may also modify the tree. The sync operations goes in both directions, since the user has the option to add new elements to the XML document.

The XML format that I currently work on has elements that represents comments, additions, deletions, quotes, links and footnotes. Many of those elements can be combined, so it is possible to have a footnote inside an addition inside a quote as an example.

mzedeler:
If there is another better way of customizing the contentState so I can add global data some other way, I'd be interested in adding that in stead.

mzedeler:
@flarnie ?

mzedeler:
Ping @flarnie...

thibaudcolas:
@mzedeler it's quite unclear what this field would be for for beyond what *you* use it for, and I don't see why accommodating what you're doing requires changes in Draft.js. Is there anything preventing you from using the following state structure?

```json
const myState = {
    // The Draft.js state.
    editorState: EditorState.createEmpty(),
    // Your own extension.
    data: { /* ... */ },
};
```

I might be misunderstanding your need but if you need this extra state anyway I don't see why you couldn't also write methods to manage it in sync with the editor state (potentially wrapping the Draft.js operations as well to make this easy).

mzedeler:
While it is possible to use the approach, you describe, @thibaudcolas, it isn't very feasible because I am writing a very lange extension for Draft, making it possible to use Draft as an XML editor.

The state of the non-draft contents are kept in a data structure called `treeState` (think XML tree) and I have about 50 places in my code where I need to access the `treeState` for a given `contentState`.

Note that the `treeState` needs to be added to `contentState`, not `editorState`, which makes it even more cumbersome, because I need to keep the `treeState` aside both `editorState` and `contentState` in all the places where they are being passed.

I have 32 locations where I call `getCurrentContent()` that needs to be replaced.

Also, I have plenty of places where I do stuff like this:

    return someEdits
      .reduce(
        (nextContentState, edit) => performEdit(nextContentState, edit),
        contentState
     )

At all those return sites, I need to return a tuple of some kind containing the `treeState` as well, including all the places where I do the same on the `editorState`. It makes the code extremely unwieldy.

A whole different point is that if we can implement this `data` attribute, it'll be much easier for me to split out the XML handling from my application and publish it as Open Source, since the API will be much easier for others to use.

mzedeler:
I have been looking into using Slate and while it seems to have an API much more geared towards the use case I have, it also isn't as mature as Draft.js, so my team decided to stick to Draft.js and build what was missing on top of it.

I have also been looking into storing data on block level, but it'll require some implicit rules regarding the metadata that I have that are far from ideal.

The bottom line is: I have metadata that relates to the document as a whole - not any particular block - that I need to be part of the undo/redo state, just like Draft.js (should) let entities be part of the undo/redo state.

niveditc:
@mzedeler - echoing the comments by @flarnie & @thibaudcolas, I don't think this is a good fit for the current API.

If it might help down the road for your particular use case with a XML tree, we are currently exploring a tree-based data model for Draft.js (you can see some of my recent commits related to that). This is still in very early stages & experimental, so no promises on stability or continued support for that at the moment.

mzedeler:
I have realized that you probably wouldn't accept this change, but do note that the approaches suggested by @thibaudcolas above aren't feasible since I'd have to implement a parallel undo/redo stack, so I've forked Draft using the change in this PR.

mzedeler:
Ping @niveditc - how did it go with the XML editing feature? I've completed mine and it just works, albeit only on a soft fork of Draft where I have a data field on contentState (as the only required API change).

PR #1687, opened 2018-03-12
Integrating experiments on playground

mitermayer:
Integrating experiments on playground
    
- Allowing playground to load experiments

### DIFF
--- a/examples/draft-0-10-0/playground/src/App.css
+++ b/examples/draft-0-10-0/playground/src/App.css
@@ -19,7 +19,7 @@
 }
 
 .playground-main select {
-	width: 100px;
+  width: 100px;
   background: #843131;
 }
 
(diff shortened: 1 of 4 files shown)


examples/draft-0-10-0/playground/src/App.css
```diff
@@ -140,3 +156,14 @@
...
+
+.RichEditor-editor {
+  padding: 10px 20px;
+}
+
+/**
+ * basic style for handling nested lists
+ */
+.RichEditor-editor li[data-block] li[data-block] {
+  margin-left: 20px;
```
flarnie: I'm surprised this doesn't have some default styling that does the indentation. Oh well.

mitermayer: we don't have nested lists support, we use CSS for the depth :(

PR #1688, opened 2018-03-14
Add test for onTab

### LINKED ISSUE #1674: Test for indenting list item more than one level
After we merge https://github.com/facebook/draft-js/pull/762 then Draft should support indentation of a list more than one item deep.

To make sure we don't regress this support in the future, let's add a test.

I would add a test in the suite of unit tests for RichEditorUtil, here -> 
https://github.com/facebook/draft-js/blob/master/src/model/modifier/__tests__/RichTextEditorUtil-test.js

We don't seem to have any test for 'onTab' so let's add one.

testing_i_love_testing

simsim0709:
**Summary**
This PR fixes #1674

### DIFF
--- a/examples/draft-0-10-0/playground/src/App.css
+++ b/examples/draft-0-10-0/playground/src/App.css
@@ -19,6 +19,7 @@
 }
 
 .playground-main select {
+  width: 100px;
   background: #843131;
 }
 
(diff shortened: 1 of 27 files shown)


simsim0709:
Sorry, I will close this and try again.

PR #1712, opened 2018-03-26
Provide an `allowUndo` parameter for `modifyInlineStyle`

lan9:
Provide an `allowUndo` parameter for `modifyInlineStyle`. When `allowUndo` is set to true, the new contentState will not have selectionBefore and selectionAfter updated. This is done so that inline style change can be done without leaving traces in redo/undo stack.


lan9:
Created it by accident. Closed it.

PR #1715, opened 2018-03-29
make readOnly really read-only

mattkrick:
**Summary**

Currently, update gets called even if the readOnly flag is true. I think that's a bug.
This fixes it.

**Test Plan**

Manual :disappointed:


niveditc:
@mattkrick - thanks for submitting this PR!

Next steps are:
* Provided the steps above are successful I'll merge it in tomorrow (Monday). In the slim chance that something breaks it's better to do a risky merge on a weekday than a weekend.

niveditc:
Closing due to no response. Please reopen & resolve comments if you still want to submit this.

mattkrick:
without this onBlur will fire IIRC, which probably isn't what you want. no need to merge, i'm happy using a fork

PR #1717, opened 2018-03-29
Exploration - Feature tree / Fixing range removal for adjcent retained

mitermayer:
### Context

**Fixing case where immediate parent is selected as anchor node and is retained**

 Fixing case where imediate parent is selected as anchor node. Due to how
  we do moveSelectionForward and moveSelectionBackward we highjack the
    natural selection and get in cases where the parent has the anchor or
    focus where it does not have text.
    
    This PR ensures that when the parent is selected and retained that its
    previous block does not lose its links

***

**Note:**
production systems.


mitermayer:
@flarnie this is not a complete fix for "delete" and "backspace" will submit 2 more PR's for it. To further clarify I missed two more files that needed to be tree aware: "moveSelectionBackward.js" and "moveSelectionForward.js". 

Am currently with the fix on my WIP tree playground branch. Will extract them out of it, add some tests and submit as separate PR's.

PR #1725, opened 2018-04-08
Fix drop issue

laysent:
**Summary**

I have created a demo to illustrate this issue, feel free to look at [here](https://github.com/laysent/draft-js-drag-and-drop-issue)

If you look at `src/component/base/DraftEditor.react.js`, there is a problem with `_dragCount`:

When you drag a file into editor and drop it, `onDragEnter` will be triggered to increase `this._dragCount`. However, `onDragLeave` will not be triggered when dropping, thus `this._dragCount` will not be decreased back to 0.

Thus, after dragging and dropping, when user drags another file and NOT drop it, both `onDragEnter` and `onDragLeave` will be triggered. This time, `this._dragCount` will be increased then decreased. However, since it's initial value after dropping is not 0, it will not be decreased back to 0 after `dragLeave`, thus it will not trigger `this.exitCurrentMode()` in `onDragLeave`. Editor remains in "drag" mode.

To resolve this issue, simply set `_dragCount` back to 0 when dropping the file.

**Test Plan**

I don't think one is necessary for this issue. But would be happy to provide one if required.

### DIFF
--- a/src/component/handlers/drag/DraftEditorDragHandler.js
+++ b/src/component/handlers/drag/DraftEditorDragHandler.js
@@ -84,6 +84,7 @@ var DraftEditorDragHandler = {
     );
 
     e.preventDefault();
+    editor._dragCount = 0;
     editor.exitCurrentMode();
 
     if (dropSelection == null) {


laysent:
@flarnie Hi, Flarnie, may I know if we could merge this fix? If you have any question about the fix or repro steps, feel free to let me know.

niveditc:
@laysent - thanks for submitting this PR! It looks safe to merge as far as I can tell. Could you possibly include a video of the issue & fix in the test plan? Basically, this would just be a screen recording in the demo that you linked to, making it easier to quickly look at a test :)

laysent:
@niveditc  Thanks for the feedback.
Following is a screen record of the demo in this repo: <https://github.com/laysent/draft-js-drag-and-drop-issue>
issue
The main problem is, when dropped a file first and not dropping a file second time, it will not come back to "edit" mode and thus will not handle enter correctly anymore.
Following is the code related to draft.js for this demo:
```jsx
<Editor
  onChange={this.onChange /* to update this.state.editorState */}
  handleReturn={() => 'handled'}
  handleDroppedFiles={() => 'handled'}
/>     
```
I have simplified some code above and only kept the necessary for illustration. All code can be found in this demo repo: <https://github.com/laysent/draft-js-drag-and-drop-issue/blob/master/src/App.js>

niveditc:
Next steps are that:
* I'll import this and run FB's test stack
* Do a few sanity checks in the browser
* Provided the steps above are successful, I'll merge it in tomorrow. In the slim chance that something breaks, it's better to do a risky merge on a weekday than a weekend.

laysent:
@niveditc Sounds great. Let me know if there is anything further I could help :-)

Issue #1739, opened 2018-04-24
Improvements for aria-describedby and placeholder usage


afercia:
I've read https://github.com/facebook/draft-js/commit/a6af3e15120e74c8797c5670f5bb63cb45c49a32 and the related issue https://github.com/facebook/draft-js/issues/1214

While I understand the intent to provide a simple implementation, seems to me there's room for a couple improvements.

1
The `aria-describedby` attribute shouldn't be rendered when there's no element with an ID referenced by its value.

When no `placeholder` and no `ariaDescribedBy` props are passed, the `aria-describedby` attribute is still rendered, for example: `aria-describedby="placeholder-6s22n"` but of course there's no element in the page with an id `placeholder-6s22n`. When trying to validate a page where this happen with a couple (appreciated) accessibility evaluation tools, this is reported as an error and for good reasons. Tools used:

WebAIM Wave:
https://wave.webaim.org/extension/

Deque aXe:
https://www.deque.com/axe/
Note: axe-core is available also as package for automated UI testing:
https://github.com/dequelabs/axe-core
https://www.axe-core.org/docs/
 
Based on the spec, the valid value for `aria-describedby` is an "ID reference list" where it's implied the referenced elements must be present in the page.

https://www.w3.org/TR/wai-aria-1.1/#aria-describedby
> Value:	ID reference list
https://www.w3.org/TR/wai-aria-1.1/#valuetype_idref_list
> A list of one or more ID references.

Also, browsers will expose to assistive technologies the information that a description exists, but actually it doesn't. A simple check could avoid to render an `aria-describedby` that points to nothing.

2
When using an `ariaDescribedBy` prop, the placeholder is not announced at all. In this scenario, the placeholder is not referenced by anything and, when landing on the editable field, assistive technologies will just ignore it. Worth noting the placeholder is meant to provide important information:  its recommended usage is for a short hint intended to help users with data entry, e.g. [a sample value or brief description of the expected format](https://www.w3.org/TR/html52/sec-forms.html#the-placeholder-attribute) (designers tend to use it as replacement for a label, but that's a different matter).

The point is that precious information may be completely lost for assistive technology users. When emulating native UI, it would be great to try to replicate all the native functionalities: it's our responsibility, as developers, to give users the same experience they have with native UIs.

[...]

Issue #1214, opened 2017-05-17
ariaDescribedBy prop should render to aria-describedby attribute


basham:
**BUG REPORT (AND POTENTIALLY A FEATURE REQUEST)**

The [`ariaDescribedBy` prop](https://github.com/facebook/draft-js/blob/v0.10.1/src/component/base/DraftEditorProps.js#L95) should render the provided `id` string to the [`aria-describedby` attribute](https://www.w3.org/TR/wai-aria/states_and_properties#aria-describedby).

```jsx
<Editor
  ariaDescribedBy="editorDescription"
  ariaLabel="Usernames or IDs"/>
<div id="editorDescription">Separate values by a comma or space.</div>
```

However, as of Draft `v0.10.1`, it will instead [render the placeholder id as the value for `aria-describedby` and only if the `placeholder` prop is used](https://github.com/facebook/draft-js/blob/v0.10.1/src/component/base/DraftEditor.react.js#L242-L244).

If retaining the current behavior is desired, then at least the `ariaDescribedBy` string should be merged with the `this._placeholderAccessibilityID` string, since the attribute supports multiple id references.

Unfortunately, if the Editor merges these ids on behalf of the developer, the developer has no way to control the ordering of these ids. The order matters, because that's the order in which text is announced by a screen reader. And the order will be different in different contexts.

**Bug Fix Recommendation:** Merge `ariaDescribedBy` prop string before the placeholder id, because that's what the developer is explicitly providing.

**Feature Recommendation:** Consider a future feature in which `ariaDescribedBy` could be a function that provides the placeholder id parameter and returns a string. Such as:

```jsx
<Editor
  ariaDescribedBy={placeholderId => `${placeholderId} editorDescription`}
  ariaLabel="Usernames or IDs"
  placeholder="Try copying and pasting values from a spreadsheet."/>
<div id="editorDescription">Separate values by a comma or space.</div>
```

If `ariaDescribedBy` is a string, then it renders it as-is, regardless of the presence of the `placeholder` prop.

```jsx
<Editor
  ariaDescribedBy="editorDescription"
  ariaLabel="Usernames or IDs"
  placeholder="Try copying and pasting values from a spreadsheet."/>
<div id="editorDescription">Separate values by a comma or space.</div>
```

flarnie:
Thanks so much for reporting this. The "Bug Fix Recommendation" would be good to have asap, since it won't affect existing uses of Draft that may be relying on the current behavior.

[...]

PR #1741, opened 2018-04-25
Make aria-describedby able to reference both the placeholder and the description

afercia:
**Summary**

I've read https://github.com/facebook/draft-js/commit/a6af3e15120e74c8797c5670f5bb63cb45c49a32 and the related issue https://github.com/facebook/draft-js/issues/1214

I'd like to propose an improvement to the way the `placeholder` and `ariaDescribedBy` props work. For more details, please see the related issue #1739.

This PR introduces a new `_getAriaDescribedByIDs()` function to calculate a proper value for the `aria-describedby` attribute.
- it returns `null` when `placeholder` and `ariaDescribedBy` aren't set, to avoid to render an `aria-describedby` attribute that points to nothing 
- when either `placeholder` or `ariaDescribedBy` is set, it just uses its value 
- when both `placeholder` and `ariaDescribedBy` are set, it uses both values, separated by a space 

The third point would allow to replicate the native behavior: when both a `placeholder` and an `aria-describedby` pointing to some description are set in a textarea, assistive technologies use both. 

If you want, you can verify this on a pen I've prepared: https://codepen.io/afercia/full/GdZOmX/

Screen readers announce first the placeholder, then the description referenced by `aria-describedby`. This is the expected behavior, since the `placeholder` and `aria-describedby` have different purposes and `aria-describedby` is not an "override" of the placeholder as mentioned in https://github.com/facebook/draft-js/issues/1214#issuecomment-316835817
Recommended combos to test:
- Safari+VoiceOver
- Firefox+NVDA
- IE11+JAWS

Specifically, the placeholder is meant to help users with data entry, for example providing [a sample value or brief description of the expected format](https://www.w3.org/TR/html52/sec-forms.html#the-placeholder-attribute) while `aria-describedby` is meant for a longer description.

Seems to me allowing both is essential to replicate the native behavior.

Worth noting that currently, when the prop `ariaDescribedBy` is set, the placeholder is not announced at all. As per backwards compatibility concerns, this change shouldn't change so much: everything should work as before, with the added bonus that both props will work.

**Test Plan**

I've tested using the "plaintext" example:
- remove the `placeholder` prop and verify no `aria-describedby` attribute is rendered
- restore the `placeholder` prop and verify the `aria-describedby` attribute value is the same of the placeholder ID
- add an `ariaDescribedBy` with a test ID (you can also add a paragraph with that ID and some content) and verify the `aria-describedby` attribute value takes both IDs
- play with the two props, remove one of them, re-add, etc. and verify the `aria-describedby` attribute value is the expected one

Worth noting `npm run lint` fails for me (also on master, see error below) and `npm run flow` shows a few errors, also on master. Seems to me they're unrelated errors, unless I'm missing something. In that case I'd appreciate any feedback and guidance 🙂 

`npm run lint` error:
```
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at hasTypeOfOperator (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/rules/no-undef.js:19:19)
    at Program:exit.globalScope.through.forEach.ref (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/rules/no-undef.js:59:44)
    at Array.forEach (<anonymous>)
    at Program:exit (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/rules/no-undef.js:56:37)
    at listeners.(anonymous function).forEach.listener (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/util/safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.leaveNode (/Users/andrea/Projects/draft-js/node_modules/eslint/lib/util/node-event-generator.js:303:14)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! draft-js@0.10.5 lint: `eslint .`
npm ERR! Exit status 1
```


afercia:
Seems to me `npm-run-flow` fails because of a change made a few days ago, see https://github.com/facebook/draft-js/commit/8000486ed6890d1f69100379d954a62ac8a4eb08 and now `entityKey` may be of an unexpected type.

Instead, `npm run lint` fails for me for some problem related to the `babel-eslint` version. When I upgrade from `"babel-eslint": "^7.2.3"` to `"babel-eslint": "^8.1.1"`, the error goes away. Maybe something similar to https://github.com/eslint/eslint/issues/9767#issuecomment-353840674 . Not sure why it works on your end. /Cc @flarnie (hello!)

src/component/base/DraftEditor.react.js
```diff
@@ -302,9 +302,21 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
...
       return <DraftEditorPlaceholder {...placeHolderProps} />;
     }
+    this._placeholderAccessibilityID = null;
     return null;
   }
 
+  _getAriaDescribedByIDs(): ?string {
+    const ariaDescribedByIDs = [
+      this._placeholderAccessibilityID,
+      this.props.ariaDescribedBy,
```
jessebeach: Without type safety, I'm concerned that folks might pass in either a string or an Array for this prop value. Is there a way to guarantee a string? Or, if not, you should run this prop through `Array.isArray` and spread it (or join it) before joining with the placeholder id.

src/component/base/DraftEditorPlaceholder.react.js
```diff
@@ -21,7 +21,7 @@ const React = require('React');
 const cx = require('cx');
 
 type Props = {
-  accessibilityID: string,
+  accessibilityID: ?string,
```
jessebeach: Rather than making the value nullable, make the property optional. If the property is defined, it must be a string.

afercia:
@jessebeach thanks for your review! I'll try to address your points as soon as I have some spare time 😄 

On a side note: should the files in `meta/bundle-size-stats/` be committed? I've seen they go stale pretty soon 🙂

src/component/base/DraftEditorPlaceholder.react.js
afercia: Looking better at this, I think it should be nullable. When the placeholder prop is not set, it's necessary to avoid to render the HTML attribute `aria-describedby` that would point to a non-existing element.

src/component/base/DraftEditor.react.js
afercia: I understand the concern. On the other hand, should this apply also to other props, for example `ariaLabelledBy`? I couldn't find in the codebase other examples of checking for a prop value this way, and it would look a bit overkill to me. If you strongly feel it's necessary, please do let me know.

niveditc:
> On a side note: should the files in meta/bundle-size-stats/ be committed? I've seen they go stale pretty soon

I think we should automate the generation of those files and/or add them to `.gitignore`. Feel free to exclude them in your commit for now and I'll look into fixing this process.

niveditc:
@afercia - just wanted to ping this! We'd love to merge this PR when you get a chance to make the changes :)

PR #1755, opened 2018-05-07
keep the fragments block type when inserting a fragment of size 1

MikeKlemarewski:
**Summary**

When calling `insertFragmentIntoContentState` with a fragment of size 1, the original block's type was getting applied instead of the fragment's block type. When inserting larger fragments, each block maintains it's own block type. The behavior should be consistent.


colinjeanne:
The original block's type is applied because the text of the fragment is spliced into the original block. This ensures that inserting an unstyled fragment into the middle of a header block doesn't change the header to be unstyled.

MikeKlemarewski:
I should have tested this a bit more against the example draft editor on draftjs.org  That behavior seems pretty consistent and expected.

I was seeing different behavior in our editor, caused by a different issue.

PR #1607, opened 2018-01-09
Migrate to Docusaurus

aadsm:
This PR updates the website to https://github.com/facebook/docusaurus.

This changes the styling of the website from what we have today in https://draftjs.org/ to the default docusaurus styling for documentation. As of today it's not easy to customize the styling of the documentation, the only way would be to override all selectors and be at the mercy of any docusaurus future styling changes.
The main page is still pretty much the same but with an additional footer.

You can check it out here: https://aadsm.github.io/draft-js/

aadsm github io_draft-js_

aadsm github io_draft-js_docs_getting-started html

PR #1759, opened 2018-05-10
Migrate to Docusaurus - Attempt #2

NoamELB:
This PR continues https://github.com/facebook/draft-js/pull/1607 (by @aadsm) to update the website to https://github.com/facebook/docusaurus.

Changes from the original PR:
* Removing the `./setup.sh` line in .travis.yml (since the file was removed).
* Upgrading to the latest Docusaurus and React.
* Adding Google Analytics to Docusaurus config.
* Adding cname to Docusaurus config.
* Wiring the correct footer links.

Live Example: https://noamelb.github.io/draft-js/

Thanks @flarnie & @yangshun for all the help ☺️ 

screencapture-localhost-3000-2018-05-10-23_18_05
screencapture-localhost-3000-docs-getting-started-html-2018-05-10-23_18_26
ga

Co-authored-by: António Afonso <aadsm@users.noreply.github.com>


flarnie:
^_^ This is fantastic. Thanks @NoamELB 

@Daniel15 do we need any extra steps for this to ship to Draftjs.org? Not sure if it will work automatically to merge it.

.travis.yml
```diff
@@ -43,18 +43,15 @@ script:
...
-    yarn install
-    ./setup.sh
     if [ "$TRAVIS_PULL_REQUEST" = false ] && [ "$TRAVIS_BRANCH" = master ]; then
       # Automatically publish the website
       git config --global user.name "Travis CI"
       git config --global user.email "travis@reactjs.org"
-      echo "machine github.com login reactjs-bot password $GITHUB_TOKEN" >~/.netrc
-      ./publish.sh
+      echo "machine github.com login reactjs-bot password $GITHUB_TOKEN" > ~/.netrc
+      cd website && yarn install && GIT_USER=reactjs-bot npm run publish-gh-pages
```
Daniel15: It's unfortunate that we need these hacks (yes, checking in committed code to Git is a hack). I'd like to eventually move the site over to Netlify to avoid it. That'd also give us preview versions of the site for each PR submitted.

yangshun: Docusaurus uses Netlify to show PR previews. It's pretty easy to set up - https://docusaurus.io/docs/en/1.0.14/publishing.html#hosting-on-netlify

Daniel15: @yangshun Yeah it'd be good to have the DraftJS site totally on Netlify, rather than using both Netlify and GitHub Pages. We shouldn't use Netlify for previews but GitHub Pages for the live site, as the difference between the environments could cause issues. Moving the site across would just need some small changes, like moving the DNS over.

yangshun: Would it cost a lot to use Netlify to host the production site though? If it's free it would be great for Docusaurus to do that too (I think docusaurus.io is also using gh-pages now)

Daniel15: Fre for open-source :) yarnpkg.com and reactjs.org use it.

yangshun: Awesome! @JoelMarcey WDYT?

JoelMarcey: The primary goal of Docusaurus was to make getting sites up and running easily. GitHub/GitHub Pages seemed to be the easiest way to get folks started on that given that they are used heavily and are integrated.

That said, we know that people want to use Netlify for hosting and CircleCI for integration (we should probably document Travis at some point 😉 ) -- which is why we have documented those.

So it's great if Draft.js wants to move to Netlify exclusively. Others will choose different options as well.

Whether Docusaurus should move exclusively to Netlify, I am not convinced. I think us hosting on GitHub Pages adheres to the first goals of Docusaurus and since we use Netlify to preview, we are showcasing other possibilities too.

Not sure I feel too strongly, but as of this point, I would say I wouldn't actively pursue a move to Netlify. But, I can ask others on the Docusaurus team what they think too.

flarnie: Thanks everyone~!

:) Happy to hear that we should be able to merge this - @NoamELB do you want to do the honors? Or would you prefer I find a time to do it?

NoamELB: You can do it 😊

JoelMarcey:
So excited to see this merged soon 🎉 ❤️ 😄

NoamELB:
screencapture-draftjs-org-2018-05-16-20_29_54

JoelMarcey:
Yes! Awesome work all!

Issue #1301, opened 2017-07-23
Draft Entity example may throw exception when input via IME


southerncross:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

If I replace a selection range that contains **\n** and **a half entity** with IME input, Draft will then throw exception：
> Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

See GIF:
![tmp](https://user-images.githubusercontent.com/3861465/28297779-460880d0-6ba2-11e7-8d44-7462d6ee7d1f.gif)

This bug can be reproduced if:
1. selection range contains \n and "half entity".
2. replace with IME input.

I guess the browser mutates the DOM structure while editing through IME input, then it causes React invariant check failed.

Draft will throw exception and the subsequent editing behavior is abnormal.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

Both Draft 0.x and 1.x are the same.

lucasreppewelander:
For those who haven't found the answer yet.
Here is the correct answer, worked for me and many more (no credit)

https://github.com/facebook/draft-js/issues/1320#issuecomment-476509968

PR #1760, opened 2018-05-14
Restore DOM when selection length is greater than 1

### LINKED ISSUE #1320: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
The current behavior:

draftjs

---

macOS + Chrome 60

---

The same issues with #1089, but nobody care.

sarmstrong:
**Summary**

This fixes issue listed in https://github.com/facebook/draft-js/issues/1301, but I believe it could fix #1320 as well. 

The editor state is not updated while an IME is composing text. In it's current state, the IME updates the DOM and then React tries to remove the nodes (entities, style ranges, etc) but the nodes are no longer there. 

If the selection contains an entity or style range, we need to restore the DOM so that React can properly remove the nodes. 

This change adds an extra condition to  the `mustRest` boolean where if the selection is not collapsed, then restore the editor before applying the updates below in lines 180-192.

I could see how one could get more sophisticated about this approach and examine the selection for an entity, style range or multiple blocks, however, I think this might add a lot of undo complexity. 

**Test Plan**

Make sure IME input works when selecting multiple ranges. Should work when selection contains multiple blocks, style ranges or entities. Should also work when the selection contains no entities, style ranges and just a single block.

### DIFF
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -155,6 +155,8 @@ const DraftEditorCompositionHandler = {
       !composedChars ||
       isSelectionAtLeafStart(editorState) ||
       currentStyle.size > 0 ||
+      editorState.getSelection().getAnchorOffset() !==
+        editorState.getSelection().getFocusOffset() ||
       entityKey !== null;
 
     if (mustReset) {


nburt:
I'm also able to reproduce this on Facebook comments:

1. Type some content
2. Start a new paragraph and mention someone
3. Highlight part of the mention and some of the previous text
4. Type using 2-Set Korean input

I would upload a gif but not really sure how to do that without posting a recording of my fb wall and leaking some of my friends' info. Below are screenshots of each step.

image

image

image

image

sarmstrong:
I can report that this solved the bug for us and we haven't had any further issues.

mrkev:
2018 wasn't a particularly active period in draft-js history haha.

I'm looking through old PRs and this one looks very interesting. @thread, is this still an issue? I'd be happy to merge it if so.

benmosher:
I am still seeing the issue in #1320 and was able to resolve with [this snippet](https://github.com/facebook/draft-js/issues/1320#issuecomment-476509968) so yes, still an issue (assuming this does indeed solve #1320 as OP was unclear whether it did or didn't)

PR #1767, opened 2018-05-17
Update Advanced-Topics-Entities.md

haikyuu:
Add a line of code in the example to set the editor state after adding the entity.

For a newcomer, it's not straightforward because it wasn't mentioned anywhere in the docs prior to arriving here. Of course, you'll figure it out once you check the example. But let's save newcomers a roundtrip

**Summary**

Update the docs

### DIFF
--- a/docs/Advanced-Topics-Entities.md
+++ b/docs/Advanced-Topics-Entities.md
@@ -69,6 +69,7 @@ const contentStateWithLink = Modifier.applyEntity(
   selectionState,
   entityKey
 );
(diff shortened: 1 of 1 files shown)


mzedeler:
Good idea :-)

I would prefer if the line used `EditorState.push(...)`, since it would become and undoable operation which is usually what people want.

niveditc:
Thank you for improving our docs, @haikyuu! I like @mzedeler's suggestion above - could you please amend your commit? Once that's done I can merge it in :)

PR #1672, opened 2018-02-28
Fix Android input bugs

mxstbr:
**Summary**

Merges master into #1419 and removes (now) unnecessary feature flags.

I know, I know @flarnie. Android support got kicked from the roadmap. I don't expect this to get merged, but I figured it'd be worth submitting it as a PR anyway now that I've spent the time to clean it up, maybe it helps somebody else.

Feel free to close it!

PR #1774, opened 2018-05-27
Working attempt at fixing Android bugs

aminland:
This is a continuation off of @mxstbr input-bugs fix  #1672 

I found that just calling editOnInput at composition end makes android magically work. There are still a tiny bit of issues like cutting text, but I can't see any reason not to treat a compositionEnd event the same way we treat a spellcheck and read from the dom...

Thoughts?


src/component/handlers/composition/DraftEditorCompositionHandler.js
```diff
@@ -183,16 +249,18 @@ const DraftEditorCompositionHandler = {
         EditorState.push(editorState, contentState, 'insert-characters'),
       );
       return;
+      */
     }
-
+    onInput(editor);
```
aminland: Is there any reason we _shouldn't_ just call the same function that handles spellchecks? This seems to make android just *work*.

mxstbr:
Going to test this in our codebase!

src/component/handlers/composition/DraftEditorCompositionHandler.js
aminland: Also i'm not sure how to test this because the jest test seem to have document.getSelection and global.getSelection = undefined. Not sure how to emulate this.

fabiomcosta:
@mxstbr did you have a chance to test this one?
I'm planning on working on the Android fix, and I'm trying to see what's the best way to get started.
I know SlateJS implemented a fix that seems pretty solid recently, but didn't merge it yet, https://github.com/ianstormtaylor/slate/pull/2565, so I think I'll be using that as the base of my attempt to fix this, but if this PR helps, it would be helpful to know.

src/component/contents/DraftEditorContents-core.react.js
```diff
@@ -269,6 +271,20 @@ class DraftEditorContents extends React.Component<Props> {
...
+    }
+
+    if (ContentWrapper) {
+      return (
+        <div data-contents="true">
+          <ContentWrapper>
+            {outputBlocks}
+          </ContentWrapper>
+        </div>
+      );
```
fabiomcosta: What is this change fixing?

aminland:
closing in favor of PR by @fabiomcosta

PR #1783, opened 2018-06-06
keep the block data when inserting fragment without block data

abel1105:
**Summary**

Here is a demo of the problem: [https://jsfiddle.net/x1d6jr9L/4/](https://jsfiddle.net/x1d6jr9L/4/)

Step:
1. copy the below text "copy here"
2. paste it after "figcaption"
2. image will disappear because the empty block data

It is weird that the block type is still atomic, but block data has been wiped out because of pasting unstyled text.

Solutions:

1. if fragment block data size is 0 then use the original block data (this PR)

https://github.com/abel1105/draft-js/blob/3bdcd9b48aceda0547ca9b0678bfbe2597b8611e/src/model/transaction/insertFragmentIntoContentState.js#L55

2. when detect focus block type is atomic then merge atomic block data into first fragment block.

```js
// editOnPaste.js:122
let htmlMap = BlockMapBuilder.createFromArray(contentBlocks);
const editorState = editor._latestEditorState;
const contentState = editorState.getCurrentContent();
const selectionState = editorState.getSelection();
const focusBlockKey = selectionState.getFocusKey();
const focusBlock = contentState.getBlockForKey(focusBlockKey);
const focusBlockType = focusBlock.getType();
// if detect paste on atomic block copy the atomic data on the first block
// one Fragment: insertFragmentIntoContentState:41
// sereval fragments: insertFragmentIntoContentState:74
if (focusBlockType === 'atomic') {
  const focusBlockData = focusBlock.getData();
  const targetBlock = htmlMap.first();
  const targetBlockKey = targetBlock.getKey();
  htmlMap = htmlMap.set(
    targetBlockKey,
    targetBlock.merge({
      data: focusBlockData
    })
  );
}
```

I think the first solution is more elegant, and implement this fix in this PR.
In order to avoid changing to many source code, I am using the second solution to workaround in my own project `handlePastedText`.

**Test Plan**

Make sure the block data will not be replaced by unstyle text fragment.


niveditc:
@abel1105 - apologies for the delay in reviewing this!

Your analysis & solution look good. I think most folks don't run into this issue because their atomic blocks may not have associated text, but this seems to work.

Next steps:
- I'm going to import this into the Facebook stack to run our test suite
- If all test pass, I will merge this on the next business day.

abel1105:
Hi, I was wondering if there is any updates on this PR ?

allanassis:
Hi guys, any updates? :D

pedrokiefer:
@niveditc any news on this? we are picking this bug on a megadraft plugin.

niveditc:
Hey folks - sorry this fell off my radar earlier. I'll follow-up on this during this week :)

Joaovsa7:
Hey guys, any update? 😄  👀

abel1105:
@niveditc I've already rebased this PR, please check.

Also, there are some error on Travis, seems like I'm not the only one.

Issue #1389, opened 2017-09-18
Extra newlines added to text pasted between two Draft editors.


flarnie:
**Do you want to request a *feature* or report a *bug*?**
This is a bug.

**What is the current behavior?**
Newlines are added to internal copied text.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**

This can  be reproduced with the 'rich' example:
1. Write some text with an empty newline between blocks. For example:
```
hello world

foo bar baz
```
2. Copy and paste all of that text, including the empty newline, into a `rich` text editor in a separate tab.
3. Notice there are extra newlines in the pasted output, like so:
```
hello world

foo bar baz
```

**What is the expected behavior?**

There should be only as many newlines as were present in the copied text.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
This is on the master branch, so 0.10.2+, and I saw it in Chrome, Safari, and Firefox.

thibaudcolas:
#1784 fixes this. If you want to address this in your project right away without forking, I added the details at the end of the PR.

Issue #1163, opened 2017-04-22
Minor issue with copy-pasting rich text styles


flarnie:
**Do you want to request a *feature* or report a *bug*?**
Bug - but it's pretty minor.

**What is the current behavior?**
Nested list styles are not retained when copy-pasting between Draft instances. See screenshot.
![screen shot 2017-04-22 at 3 07 04 pm](https://cloud.githubusercontent.com/assets/1114467/25308920/85a5d458-2774-11e7-9f91-1acf4eead554.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/stopachka/m6z0xn4r/.**
1) Open the rich text editor example
2) Create a nested list
3) Open another tab with the rich text editor example
4) Copy paste between them. 

**What is the expected behavior?**
The nesting should be copied accurately.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
Chrome, and not sure which version this started in. We didn't support nesting until recently also.

thibaudcolas:
I think this got solved in #1605, except for lists deeper than 5 levels (https://github.com/facebook/draft-js/pull/1605#discussion_r160256499).

thibaudcolas:
#1784 fixes this too.

Issue #1154, opened 2017-04-20
Copy/paste between editors strips soft returns


DawnWright:
When copying from one draft editor and then pasting into another, I am seeing soft returns get stripped. The soft returns were '\n' characters added via `RichUtils.insertSoftNewline`.
https://github.com/facebook/draft-js/blob/213cd764f61cc64552bddd672ae1748529d55333/src/model/modifier/RichTextEditorUtil.js#L82

Here is a small test that can be added to `convertFromHTMLToContentBlocks-test.js` which demonstrates the problem:

```
  it(`must not strip soft returns`, () => {
    const contents = 'a\nb'
    const html_string = `
      <div>${contents}</div>
    `;

    const blocks = convertFromHTMLToContentBlocks(html_string);

    expect(blocks.contentBlocks.length).toBe(1);
    expect(blocks.contentBlocks[0].text).toBe(contents);
  });
```
result:
```
    Expected value to be (using ===):
      "a\nb"
    Received:
      "a b"
```

This happens on account of this line of code:
https://github.com/facebook/draft-js/blob/507925ad60638734872dba52efa2d0b66cd94e4b/src/model/encoding/convertFromHTMLToContentBlocks.js#L357
Is there a way we can change this behavior to handle copy paste between draft editors?

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

0.11.0-alpha
Chrome, MacOS

pascalopitz:
Same issue as the one I filed just now? https://github.com/facebook/draft-js/issues/1780

hama0511:
same here

Issue #787, opened 2016-11-16
Copy/paste between editors 


bryanph:
How can I make sure custom blocks and entities are preserved if I copy from one Draft editor to another? Hence, I want the custom block/entity to be recreated in the new editor.

Take for example the **entity** example in the repo, when copying from one draft instance to another, the entities are not copied along, just the raw text. The same happens with custom blocks, for example in the **twitter** example

bryanph:
I think one way to implement this is to set the required data for the blocks and entities in the ClipboardData object when the user triggers a copy event. Then on a paste event you check if the required data is set in the ClipboardData and if so, use that data to paste the entities and custom blocks into the second editor.

What do you think?

StorytellerCZ:
Yes, but won't that interfere with copying into non-draft editors?

bryanph:
@StorytellerCZ It seems like you can just set arbitrary data in the DataTransfer object keyed by a "type", see https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent/clipboardData . No idea if this is an accepted practice though.

In the example below I do that. One thing to be cautious of though is that every time a user copies editor data, you must serialize the content state which is quite costly. No

```
<html>
    <script>
    function addExtraData(e) {
        var clipboardData = e.clipboardData;

        var someObj = {
            "test": { "test": "test" }
        }

        var selection = window.getSelection()

        console.log("called addExtraData")
        console.log(selection)

        e.clipboardData.setData('text/plain', selection)
        e.clipboardData.setData('test', JSON.stringify(someObj))

        e.preventDefault()
    }

    function getExtraData(e) {
        var data = e.clipboardData.getData('text/plain')
        var data2 = e.clipboardData.getData('test')
        console.log("called getExtraData")
        console.log(data)
        console.log(JSON.parse(data2))
    }

    document.addEventListener('copy', addExtraData);
    document.addEventListener('paste', getExtraData);

    </script>
    <body>
        <p>
            Text to Copy
        </p>
        <textarea rows="4" cols="50"></textarea>
    </body>
</html>

```

mitermayer:
We are currently working on a solution for this issue

christophepas:
Hey @mitermayer ! Glad to know this is wip, have you any update on a solution for this ?

Apart from mentions this is also a problem for nested lists, and in general an `onCopy` method added to the Editor API could help solve these issues.

richardson-trevor:
[...]

PR #1784, opened 2018-06-10
Transfer selected content via HTML clipboard on copy-paste between editors

thibaudcolas:
**Summary**

This PR changes how Draft.js handles copy-cut-paste between editors so all content is kept as-is.  I’ve been working on this as an external package at https://github.com/thibaudcolas/draftjs-conductor, but there is no reason for this not to just be fixed in Draft.js directly.

On copy, we store the selected content within the clipboard’s HTML, overriding the browser’s default copy/cut handling, so that the Draft.js paste processor can then retrieve content from the serialised "Raw" content instead of relying on HTML parsing.

This fixes the following issues: #380, #787, #1389, #1163, #1154, https://github.com/facebook/draft-js/pull/1605#pullrequestreview-87340460, and #523 but only for the copy-paste case, https://github.com/facebook/draft-js/issues/523#issuecomment-371098488. Here's what it looks like in practice, copy-pasting between two tabs in the media example:

draftjs-copy-paste-787

As-is, the new copy/cut behavior and HTML clipboard generation doesn't replicate browsers’ behavior of inlining styles into the clipboard content (e.g. if you copy content in a particular font into Google Docs, browsers will try to preserve that). This seems like an advantage to me, but might be worth assessing further. For example, it means that copy-pasting nested lists into Google Docs will no longer add extra indentation to the (otherwise not nested anymore) list items there.

Edit: one last thing – I put the serialised content inside `text/html` instead of using a separate mime type for two reasons. First one is that without this PR, Draft.js only offered access to HTML via  `handlePastedText` for me to hook into. Second reason is that I've heard some browsers didn't support custom mime types. Fixing this issue in Draft.js itself, we could revisit this.

**Test Plan**

- [ ] Add unit tests for `editOnCopy` – I don't think I've seen any yet, would welcome some direction on how to test this.
- [x] Add unit tests for `DraftPasteProcessor`
- [ ] Run the manual [copy-paste test plan](https://github.com/facebook/draft-js/wiki/Manual-Testing#cutcopypaste) in browsers supported by Draft.js (particularly IE11)
- [ ] Go through copy-pasting into multiple word processors, with the following [support matrix](https://github.com/thibaudcolas/draftjs-filters#word-processor-support).

---

Before making this PR, I first fixed this issue for my own editor. If you want to learn more, have a look at https://github.com/wagtail/wagtail/pull/4582.

If you want to use this in your own editor without waiting for the PR to be merged, I made my fix available as a separate package: https://github.com/thibaudcolas/draftjs-conductor#idempotent-copy-paste-between-draftjs-editors. Here's what it looks like to use it:

<details>

```
npm install --save draftjs-conductor
```

Then:

```js
import {
  registerCopySource,
  handleDraftEditorPastedText,
} from "draftjs-conductor";

class MyEditor extends Component {
  constructor(props: Props) {
    super(props);

    this.state = {
      editorState: EditorState.createEmpty(),
    };

    this.onChange = this.onChange.bind(this);
    this.handlePastedText = this.handlePastedText.bind(this);
  }

  componentDidMount() {
    this.copySource = registerCopySource(this.editorRef);
  }

  onChange(nextState: EditorState) {
    this.setState({ editorState: nextState });
  }

  handlePastedText(text: string, html: ?string, editorState: EditorState) {
    let newState = handleDraftEditorPastedText(html, editorState);

    if (newState) {
      this.onChange(newState);
      return true;
    }

    return false;
  }

  componentWillUnmount() {
    if (this.copySource) {
      this.copySource.unregister();
    }
  }

  render() {
    const { editorState } = this.state;

    return (
      <Editor
        ref={(ref) => {
          this.editorRef = ref;
        }}
        editorState={editorState}
        onChange={this.onChange}
        handlePastedText={this.handlePastedText}
      />
    );
  }
}
```

</details>

### DIFF
--- a/src/component/handlers/edit/editOnCopy.js
+++ b/src/component/handlers/edit/editOnCopy.js
@@ -14,6 +14,8 @@
 
 import type DraftEditor from 'DraftEditor.react';
 
+const ContentState = require('ContentState');
+const convertFromDraftStateToRaw = require('convertFromDraftStateToRaw');
 const getFragmentFromSelection = require('getFragmentFromSelection');
 
 /**
@@ -31,7 +33,31 @@ function editOnCopy(editor: DraftEditor, e: SyntheticClipboardEvent<>): void {
     return;
   }
 
-  editor.setClipboard(getFragmentFromSelection(editor._latestEditorState));
+  const fragment = getFragmentFromSelection(editor._latestEditorState);
+
+  editor.setClipboard(fragment);
+
+  // IE11 does not support ClipboardEvent.clipboardData.
+  if (e.clipboardData && fragment) {
+    const content = ContentState.createFromBlockArray(fragment.toArray());
+    const serialisedContent = JSON.stringify(
+      convertFromDraftStateToRaw(content),
+    );
+
+    const fragmentElt = document.createElement('div');
+    const domSelection = window.getSelection();
+    fragmentElt.appendChild(domSelection.getRangeAt(0).cloneContents());
+    fragmentElt.setAttribute('data-editor-content', serialisedContent);
+    // We set the style property to replicate the browser's behavior of inline
+    // styles in rich text copy-paste. This is important for line breaks to be
+    // interpreted correctly when pasted into another word processor.
+    fragmentElt.setAttribute('style', 'white-space: pre-wrap;');
+
+    e.clipboardData.setData('text/plain', domSelection.toString());
+    e.clipboardData.setData('text/html', fragmentElt.outerHTML);
+
+    e.preventDefault();
+  }
 }
 
 module.exports = editOnCopy;
--- a/src/model/paste/DraftPasteProcessor.js
+++ b/src/model/paste/DraftPasteProcessor.js
@@ -24,6 +24,7 @@ const Immutable = require('immutable');
 
 const convertFromHTMLtoContentBlocksClassic = require('convertFromHTMLToContentBlocks');
 const convertFromHTMLtoContentBlocksNew = require('convertFromHTMLToContentBlocks2');
+const convertFromRawToDraftState = require('convertFromRawToDraftState');
 const generateRandomKey = require('generateRandomKey');
 const getSafeBodyFromHTML = require('getSafeBodyFromHTML');
 const gkx = require('gkx');
@@ -46,6 +47,29 @@ const DraftPasteProcessor = {
     html: string,
     blockRenderMap?: DraftBlockRenderMap,
   ): ?{contentBlocks: ?Array<BlockNodeRecord>, entityMap: EntityMap} {
+    const body = getSafeBodyFromHTML(html);
+    const fragmentElt = body && body.querySelector('[data-editor-content]');
+    const fragmentAttr =
+      (fragmentElt && fragmentElt.getAttribute('data-editor-content')) || null;
+
+    // Handle the paste without converting the HTML if it comes from another Draft.js editor.
+    if (fragmentAttr) {
+      let rawContent;
+
+      try {
+        // If JSON parsing fails, handle the paste as normal HTML.
+        rawContent = JSON.parse(fragmentAttr);
+      } catch (error) {}
+
+      if (rawContent) {
+        const content = convertFromRawToDraftState(rawContent);
+        return {
+          contentBlocks: content.getBlocksAsArray(),
+          entityMap: content.getEntityMap(),
+        };
+      }
+    }
+
     return convertFromHTMLtoContentBlocks(
       html,
       getSafeBodyFromHTML,
--- a/src/model/paste/__tests__/DraftPasteProcessor-test.js
+++ b/src/model/paste/__tests__/DraftPasteProcessor-test.js
@@ -342,3 +342,37 @@ test('must create ContentBlocks when experimentalTreeDataSupport is disabled whi
 test('must create ContentBlockNodes when experimentalTreeDataSupport is enabled while processing text', () => {
   assertDraftPasteProcessorProcessText(['Alpha', 'Beta', 'Charlie'], true);
 });
+
+test('must use serialised Draft.js fragment when available', () => {
+  assertDraftPasteProcessorProcessHTML(`
+    <div data-editor-content='${JSON.stringify({
+      entityMap: {},
+      blocks: [
+        {
+          key: 'a',
+          text: 'line\nbreak',
+          type: 'unstyled',
+        },
+      ],
+    })}'>
+        <p>line\nbreak</p>
+    </div>
+  `);
+});
+
+test('must fallback to HTML if serialised Draft.js fragment is invalid JSON', () => {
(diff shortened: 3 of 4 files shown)


src/component/handlers/edit/editOnCopy.js
```diff
@@ -31,7 +33,31 @@ function editOnCopy(editor: DraftEditor, e: SyntheticClipboardEvent<>): void {
     return;
   }
 
-  editor.setClipboard(getFragmentFromSelection(editor._latestEditorState));
+  const fragment = getFragmentFromSelection(editor._latestEditorState);
+
+  editor.setClipboard(fragment);
+
+  // IE11 does not support ClipboardEvent.clipboardData.
+  if (e.clipboardData && fragment) {
```
thibaudcolas: I think skipping IE11 here is fair, since it doesn't handle HTML in the clipboard to start with.

src/component/handlers/edit/editOnCopy.js
```diff
@@ -31,7 +33,31 @@ function editOnCopy(editor: DraftEditor, e: SyntheticClipboardEvent<>): void {
...
+    const content = ContentState.createFromBlockArray(fragment.toArray());
+    const serialisedContent = JSON.stringify(
+      convertFromDraftStateToRaw(content),
+    );
+
+    const fragmentElt = document.createElement('div');
+    const domSelection = window.getSelection();
+    fragmentElt.appendChild(domSelection.getRangeAt(0).cloneContents());
```
thibaudcolas: This is inspired from Rangy: https://github.com/timdown/rangy/blob/1e55169d2e4d1d9458c2a87119addf47a8265276/src/core/domrange.js#L515-L520.

Also from what I understand modern browsers only support one range in the selection.

src/model/paste/DraftPasteProcessor.js
```diff
@@ -46,6 +47,29 @@ const DraftPasteProcessor = {
...
+    const fragmentAttr =
+      (fragmentElt && fragmentElt.getAttribute('data-editor-content')) || null;
+
+    // Handle the paste without converting the HTML if it comes from another Draft.js editor.
+    if (fragmentAttr) {
+      let rawContent;
+
+      try {
+        // If JSON parsing fails, handle the paste as normal HTML.
+        rawContent = JSON.parse(fragmentAttr);
```
thibaudcolas: I'm not sure in what scenario that would occur – perhaps if the clipboard gets truncated somehow.

src/component/handlers/edit/editOnCopy.js
colinjeanne: We handle IE11 in our fork in a rather different way. We have two hidden divs, one for copy and one for paste, that exist to accept or receive incoming content. In the copy case we copy the HTML of the component into this div and temporarily move selection so that the browser is able to copy that content in the way it wants to. For paste we temporarily move focus to the other div and then grab the content from there and convert it into a fragment.

The trick is ensuring those hidden divs are styled the same as the editor and that the focus games play out correctly but it works extremely well.

thibaudcolas: @colinjeanne could you share your code's fork, at least the part that's relevant to this? I think I'd be keen to add this to my editor as well, depending on how much code it takes.

colinjeanne: Yeah, there are a few different changes, the first was https://github.com/textioHQ/draft-js/pull/21 followed later by https://github.com/textioHQ/draft-js/pull/26. We originally wrote this solution in order to get HTML into the clipboard even for browsers that didn't support adding it explicitly.

pablopunk:
Any update on this? I think I'm having an issue that could be solved with this PR

thibaudcolas:
@pablopunk this is working really well in my editor via the separate package I made available (https://github.com/thibaudcolas/draftjs-conductor#idempotent-copy-paste-between-draftjs-editors).    If you have time to try it out, I'm sure more feedback would help the PR.

rchasman:
Thanks for this @thibaudcolas, and packaging it so nicely! We are having the same issues. Facebook, any plans to merge this?

niveditc:
Thanks @thibaudcolas for submitting this & other folks for confirming that this is working for you :) 

For this change, I want to make sure that it doesn't break any existing copy-paste behavior on facebook.com or our internal editors. Will import into our stack and test it soon - thanks for your patience!

thibaudcolas:
That's great to hear @niveditc! I haven't received any bug reports from the users of my editor since putting this in production in august 2018.

A user of `draftjs-conductor` has opened https://github.com/thibaudcolas/draftjs-conductor/issues/12 as a potential bug, however I wouldn't be surprised if it was due to the Draft.js code rather than my overrides. That said, there are a few other sources of bugs I can easily think of:

- The lack of "inlining of styles" that browsers do by default for rich copy-paste. This has been an advantage in my testing (less useless styles, see top comment, and https://github.com/thibaudcolas/draftjs-conductor/blob/d0c8d8a7b3729d5f843da0b692707a91eb210da7/src/lib/api/copypaste.js#L39-L42), but I'm not super familiar with how this inlining of styles in the HTML clipboard works, and whether/how to replicate it well.
- Draft.js’s lack of ability to define *what* should be retained on paste (#166, see also https://github.com/facebook/draft-js/pull/1057#pullrequestreview-191954277). I'm sure things would break if different editors support different styles/entities/block types, and users try to copy/paste content between them without it being filtered.

For the latter issue, this is already happening when pasting links or images from rich text sources in an editor that doesn't support them, so in a sense that’s nothing new. It will just become more likely if the copy/paste retains more content. I haven't had this issue with my editor because I also made a filtering library (https://github.com/thibaudcolas/draftjs-filters) so copy/pasting only keeps what the editor actually supports.

yasuf:
wow @thibaudcolas thanks a lot for the detailed explanation I'll try using your package for this!

thibaudcolas:
My pleasure @yasuf 👌 Let me know if you run into any hiccups.

thibaudcolas:
I’ve updated the user-land implementation in [draftjs-conductor](https://github.com/thibaudcolas/draftjs-conductor) to use the new `onCopy` and `onCut` methods added in Draft.js v0.11.0 (https://github.com/facebook/draft-js/commit/d09ef3e416beb766b5a227ea84e8d735ff11823b). This makes the API a bit nicer, but doesn’t change implementation much. Unfortunately those new event handlers don’t allow falling back to the built-in Draft.js implementation like other methods do (#721), so under the hood there’s more hooking into Draft.js internals needed.

---

The next step would be to switch to using a custom MIME type to transfer the pasted content (`application/x-draft-js`), and use the new `onPaste` to read from this (https://github.com/thibaudcolas/draftjs-conductor/issues/269). If I get to this I might update this PR to also change its implementation, but 2 years on it would be nice to get some signal from maintainers to make sure I’m not doing this for nothing.

thibaudcolas:
Closing because of #3136. I’m glad I didn’t waste any further time on this PR!

PR #1811, opened 2018-07-06
Fix SelectionState’s `hasEdgeWithin`

### LINKED ISSUE #1700: SelectionState `hasEdgeWithin` has a false positive case
Bug

_Terminology: “target range” is the range described by the arguments to `hasEdgeWithin`._

When a SelectionState is confined to a single block, `hasEdgeWithin` returns true when one edge is before the target range and the other edge is after the target range. (I.e., a special case of neither edge being within the target range.)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

https://runkit.com/andrewbranch/5aac76e3315c160012a07b84

```js
const selection = SelectionState.createEmpty('a').merge({
  anchorOffset: 0,
  focusOffset: 20
});

// currently returns true, even though neither 0 nor 20 is between 5 and 10
selection.hasEdgeWithin('a', 5, 10);
```

`hasEdgeWithin` returns false for all cases where neither selection edge is within the target range.

The `hasEdgeWithin` algorithm [hasn’t changed since Draft JS was open sourced.](https://github.com/facebook/draft-js/blame/master/src/model/immutable/SelectionState.js#L87)

andrewbranch:
Fixes #1700, which includes a detailed description and reproduction of the issue.

Hopefully it's ok that I refactored the true/false assertions not to use snapshots; I found them confusing and tedious and figured it was probably just an artifact of migrating to Jest a long time ago.

### DIFF
--- a/src/model/immutable/SelectionState.js
+++ b/src/model/immutable/SelectionState.js
@@ -91,7 +91,10 @@ class SelectionState extends SelectionStateRecord {
     if (anchorKey === focusKey && anchorKey === blockKey) {
       const selectionStart = this.getStartOffset();
       const selectionEnd = this.getEndOffset();
-      return start <= selectionEnd && selectionStart <= end;
+      return (
+        (start <= selectionStart && selectionStart <= end) || // selectionStart is between start and end, or
+        (start <= selectionEnd && selectionEnd <= end) // selectionEnd is between start and end
+      );
     }
 
     if (blockKey !== anchorKey && blockKey !== focusKey) {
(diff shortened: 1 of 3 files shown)


niveditc:
@andrewbranch - thanks for putting up this fix. I won't get a chance to review until next weekend, but if you could rebase on master in the meanwhile, that would be a big help!

andrewbranch:
@niveditc done 👍

niveditc:
Next steps are that:
* I'll import this and run FB's test stack
* Do a few sanity checks in the browser
* Provided the steps above are successful, I'll merge it in tomorrow. In the slim chance that something breaks, it's better to do a risky merge on a weekday than a weekend.

Issue #1822, opened 2018-07-25
convertFromHTML breaks after converting \n string


yury-sannikov:
**Do you want to request a *feature* or report a *bug*?**
I want to report a bug

**What is the current behavior?**
After a call to the `convertFromHTML('\n')` another call to the `convertFromHTML`, for instance `convertFromHTML('<h1>H</h1>')` return unstyled block type.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
https://jsfiddle.net/m6z0xn4r/1190/

**What is the expected behavior?**
Call to the `convertFromHTML` should be idempotent.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
0.10.5 (latest)

yury-sannikov:
The core issue is here:
https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L376

A call to the `let chunk = {...EMPTY_CHUNK};` create a shallow copy of the `EMPTY_CHUNK` object. It means that `chunk.blocks` array (and others) share the same reference to the `EMPTY_CHUNK.blocks`.

If code hit this line https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L633 `EMPTY_CHUNK` object will end up having `block` array with one element of `{type: 'unstyled', depth: 0}` which caused the aforementioned issue.

yury-sannikov:
PR has been created: https://github.com/facebook/draft-js/pull/1823

MikeKoval:
I also faced the same issue, thank you @yury-sannikov

claudiopro:
Issue already fixed with switch to the new HTML converter, test case was added with #1823

PR #1823, opened 2018-07-25
convertFromHTML breaks after converting \n string, issue #1822

yury-sannikov:
**Summary**
The core issue is here:
https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L376

A call to the `let chunk = {...EMPTY_CHUNK};` create a shallow copy of the `EMPTY_CHUNK` object. It means that `chunk.blocks` array (and others) share the same reference to the `EMPTY_CHUNK.blocks`.

If code hit this line https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js#L633 `EMPTY_CHUNK` object will end up having `block` array with one element of `{type: 'unstyled', depth: 0}` which caused the aforementioned issue.

Next call to the `convertFromHTML` will start with polluted `EMPTY_CHUNK` object. :-(

**Test Plan**
Open a https://jsfiddle.net/m6z0xn4r/1190/
Put a breakpoint on `convertFromHTML` and see 2 different results while converting headers

Another way:
Open chrome dev tools and:
- execute `convertFromHTML('\n')`
- execute `convertFromHTML('<h1>H</h1>')`
Observe that second call return a block with unstyled type.


nitayneeman:
Thanks for the PR @yury-sannikov, I'm facing the same issue.

I'm just wondering what's the status of this PR? Are you planning to fix the conflicts soon? 🤔

claudiopro:
Thanks for spotting the issue and submitting a fix @yury-sannikov. Please note the [convertFromHTMLToContentBlocks.js](https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks.js) converter is deprecated and will be replaced by [convertFromHTMLToContentBlocks2.js](https://github.com/facebook/draft-js/blob/master/src/model/encoding/convertFromHTMLToContentBlocks2.js).

May I ask you to update this PR by rebasing the commits on latest master, and write a test case for convertFromHTMLToContentBlocks2.js as well?

PR #1835, opened 2018-08-13
Disable auto-mocking for tree data invariants test

niveditc:
**Summary**
Travis build was failing because Jest was auto-mocking all modules.

**Test Plan**
```
yarn run test
```


niveditc:
Creating a different PR with fix

PR #1837, opened 2018-08-15
Fixes incorrect docs, see #1497

### LINKED ISSUE #1497: Element Type Invalid Error rendering Custom block type
Hi, I'm wondering if this is a bug or if it's something I'm doing wrong. I'm following the instructions on the draft-js site for custom block rendering:

https://draftjs.org/docs/advanced-topics-custom-block-render-map.html#content

When I try to swap the block type to the custom block type I configure, it gives the error:

```
Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it’s defined in.
```

Here's a github I made to show the problem:
https://github.com/albertlockett/draftjs-help/

If you click just below the button, (in the editor), and the click the button to change the block type, you'll get that error in the console.

I would expect that it renders the editor to change ot my custom block type.

I experienced this in Chrome
"draft-js": "0.10.0",
"react": "15.4.2",

mattgreenfield:
**Summary**

Updates the docs to fix #1497. The current suggested method results in an `Element type is invalid:` error.

I can confirm that using the component itself rather than its name as a string works as expected.

Credit goes to @albertlockett

### DIFF
--- a/docs/Advanced-Topics-Custom-Block-Render.md
+++ b/docs/Advanced-Topics-Custom-Block-Render.md
@@ -142,7 +142,7 @@ const blockRenderMap = Immutable.Map({
     // element is used during paste or html conversion to auto match your component;
     // it is also retained as part of this.props.children and not stripped out
     element: 'section',
-    wrapper: MyCustomBlock,
+    wrapper: <MyCustomBlock />,
   }
 });
 


mattgreenfield:
@facebook-github-bot - I've now signed the CLA

albertlockett:
Thanks for the shoutout @mattgreenfield!

PR #1838, opened 2018-08-16
Fix incorrect link to docs in README.md

pataar:
Link was pointing to a 404 page.


niveditc:
Thanks for putting up a PR to update this error in the docs! We pulled in #1802 instead, so closing this one :)

PR #1841, opened 2018-08-21
Export RawDraftContentState publicly

cedricmessiant:
**Summary**

When I used v0.10.1 I was able to import `RawDraftContentState`. Since I upgraded to version 0.10.5 I can't.

Example of line that does not work anymore:

`import { RawDraftContentState } from 'draft-js';`


niveditc:
Seems reasonable! Thank you for submitting this PR :) 

I'm going to import this in & will merge it in if all tests pass!

Issue #1845, opened 2018-08-27
Edge/Win10 Swipe keyboard corrupt editor state


lan9:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Using [Windows 10's swipe keyboard](https://www.theverge.com/2018/6/14/17464736/microsoft-windows-10-swiftkey-keyboard-update)in Edge can corrupt editorstate. Console log display: "Unable to get property 'replaceChild' of undefined or null reference'.
Note:
This only happens if you use the swipe feature at the beginning of a new line.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. **
Can be reproduced on draftjs.org or https://jsfiddle.net/stopachka/m6z0xn4r/.
Steps:
- Enable Windows 10 Touch Keyboard. This requires a windows device with keyboard detached (tablet mode). It's possible to simulate this environment with VMWare on Mac if you tweak the config to simulate touch support, and force enable tablet mode in settings.
- Use Edge to open draftjs.org, swipe hello on the Touch keyboard. 
- Observe error in console. At this moment editorstate can be in a bad state. (can't type / delete / error in console).

This is a demo of the steps: https://twitter.com/DevYuze/status/1034207906206429184

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
Edge with Windows 10 Swipe keyboard. Doesn't work in previous version of Draftjs.

lan9:
We have a fix in https://github.com/twitter-forks/draft-js/pull/8 --- going to post a pull request here soon.

PR #1848, opened 2018-08-28
Always use \ n for new line detection

lan9:
**Summary**

Draft.js uses < br > for new line detection on all browsers except for IE. This can sometimes be problematic because most browsers[1] removes the < br > as user type in a new line. When react wants to remove the < br >, it may already be gone.

In https://github.com/facebook/draft-js/issues/1845 , a react exception is thrown when user tries to input with Windows 10 touch keyboard. This PR propose to always use \ n (rather than < br >) for new line detection, which fixes the issue. 

[1] Latest version of Edge, Chrome, Safari, but not Firefox. https://lan9.github.io/keyevent.html is a minimal test case --- use different browsers and focus on the second line, type something, noticed that the < br > is removed by browser.

**Test Plan**

Steps:
- Enable Windows 10 Touch Keyboard. This requires a windows device with keyboard detached (tablet mode). It's possible to simulate this environment with VMWare on Mac if you tweak the config to simulate touch support, and force enable tablet mode in settings.
- Use Edge to open draftjs.org, swipe hello on the Touch keyboard.
Observe error in console. At this moment editorstate can be in a bad state. (can't type / delete / error in console).
This is a demo of the steps: https://twitter.com/DevYuze/status/1034207906206429184

Follow the same step to verify the fix on: https://jsfiddle.net/lang1z/ygmzjf31/14/

**Discussion**
1) If most browsers remove the < br >, why is it only a problem on Edge with Touch keyboard?
There are two hypothesis, but they are not verified:
Hypothesis 1: because both browser and react try to remove the < br >, it's a race condition. If React remove it before browser does, it's fine. Otherwise it's not.
Hypothesis 2: React's maker knows some browsers remove the < br >, so it has logic to adapt to it. Because Touch keyboard triggers wrong key events (you can test it on https://jsfiddle.net/lang1z/ygmzjf31/14/), React may not be able to adapt.

2) Why did the code base used  < br > in the first place?
It's very obvious the original author deliberately choose to use < br > on most browsers. The motivation is unknown to me. I tested  https://jsfiddle.net/lang1z/ygmzjf31/14/ on reasonably old browser: Chrome 55, Firefox 45 and Safari 10, new line rendering and typing all seem fine.
An safer alternative is to only switch Edge to use \ n, as the issue really only surfaced on Edge, like what we've done in twitter-forked draft-js: https://github.com/twitter-forks/draft-js/pull/8. However, I feel such patches may not be healthy for the codebase in long term.


lan9:
@flarnie we found this issue and have it fixed in twitter-forks -- @comp615 and I thought it's worth to have a broader fix discussed with the maintainers. 

I'm happy to clarify and iterate if you have any comments!

niveditc:
Hey @lan9 - thanks for submitting this PR :) 

We're currently in this process of onboarding new internal maintainers for Draft.js, so the new folks (including myself) are still gaining context. Please give us a few more days to review this! :) 

Also, we're planning to hold a sync meeting with some of our external stakeholders tomorrow at 9:30 am PDT. Would love to have you, @comp615 and/or anyone else from Twitter join us - please let me know!

Issue #1775, opened 2018-05-29
DocSearch ready for your documentation website


s-pace:
👋 team:

Congratulations, your search is now ready! You [can try a showcase here](https://draftjs-docsearch.surge.sh/):
![Preview here](https://cl.ly/2X413T2r2O0X/download/Screen%20Recording%202018-05-29%20at%2011.46%20AM.gif)

I've successfully configured the underlying crawler and it will now run every 24h.

You're now a few steps away from having it working on your website:

- Copy the following CSS/JS snippets and add them to your page

```html
<!-- at the end of the HEAD -->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />

<!-- at the end of the BODY -->
<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
<script type="text/javascript"> docsearch({
  apiKey: 'ae94c9e3ee00ea8edddd484adafc37cd',
  indexName: 'draft-js',
  inputSelector: '### REPLACE ME ####',
  debug: false // Set debug to true if you want to inspect the dropdown
});
</script>
```
- Add a search input in your page if you don't have any yet. Then update the inputSelector value in JS snippet to a CSS selector that targets your search input field.
- Optionally customize the look and feel by following [the DocSearch documentation](https://community.algolia.com/docsearch/documentation/)
- You can also check your configuration [in our GitHub repo](https://github.com/algolia/docsearch-configs/blob/master/configs/draft-js.json).

The integration within Docusaurus is [even faster](https://docusaurus.io/docs/en/search.html#docsNav). We do provide analytics on the search. Feel free to ping if you want me to grant you access.

Please open a pull request if want to leverage your configuration!

Feel free to get back to us if you have any issues or questions regarding the integration.

We'd also be happy to get your feedback and thoughts about DocSearch - so we can continue to improve it.

Have a nice day :)

niveditc:
Seeking input from some folks who have helped with setting up Docusaurus for Draft.js - @yangshun, @NoamELB, @aadsm. What do you think about adding this and is it being used in other FB open-source projects?

yangshun:
@niveditc Yes! Algolia docSearch is frequently being used for Docusaurus projects. It's the de facto way we integrate search into a Docusaurus site now.

niveditc:
@yangshun - awesome! Seems exciting :D

[...]

PR #1864, opened 2018-09-12
Add Algolia search

yangshun:
**Summary**

Add Algolia search as per #1775.

**Test Plan**

screen shot 2018-09-12 at 10 07 59 am

Clicking on the results brings me to the correct location in the docs. It's magic!

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -47,6 +47,12 @@ const siteConfig = {
 
   // No .html in the path.
   cleanUrl: true,
+
+  // Algolia search.
+  algolia: {
+    apiKey: 'ae94c9e3ee00ea8edddd484adafc37cd',
+    indexName: 'draft-js',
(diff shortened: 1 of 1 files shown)


yangshun:
@niveditc I'm quite sure it has to do with Travis or `node_modules/fbjs-scripts/node/check-dev-engines.js` file.

When I set the Travis node version in `.travis.yml` to 8.9.0, the job passed - https://travis-ci.org/facebook/draft-js/jobs/427571646

However, using the default of 8, it installed 8.12.0 but the error message suggested that the Node version is 6.4.1 - https://travis-ci.org/facebook/draft-js/jobs/427490443

This is also failing in master as #1865 which is just a small typo fix is failing because of the same reason. I'll debug by trying to dig into `fbjs-scripts/node/check-dev-engines.js`.

Issue #1868, opened 2018-09-12
Missing Block Type in `insertFragmentIntoContentState`


marconi1992:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Pasting only an HTML element the block type is not preserved.

Steps to reproduce:

- Copy an heading element such as (h1, h2) from any page.
- Paste in the demo editor here https://draftjs.org/

**What is the expected behavior?**
Keep the block type.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

_version:_ 0.10.5
_browser:_ It's not brower issue, the bug is happening in the helper.

colinjeanne:
This is a dupe of https://github.com/facebook/draft-js/issues/1511

Issue #1511, opened 2017-11-16
Modifier.replaceWithFragment() loses block type if fragment size === 1


tomkelsey:
**Do you want to request a *feature* or report a *bug*?**

*Bug*

**What is the current behavior?**

`Modifier.replaceWithFragment()` loses block type if `fragment size === 1`

**What is the expected behavior?**

`Modifier.replaceWithFragment()` persists block type

**Other**

I don't know if there's a wider reason for this that I haven't understood yet but

https://github.com/facebook/draft-js/blob/c1150a729080d04014551d1c78292cb05da7cf1b/src/model/transaction/insertFragmentIntoContentState.js#L61

I believe there should be an extra line after `data: pastedBlock.getData()` with `type: pastedBlock.getType()` as this would seemingly resolve the issue?

colinjeanne:
Consider this: suppose I have a block whose type is `unordered-list-item` and whose text is `Hello meh world`. I select `meh` and paste `great` over it. If I copied `great` from somewhere outside of Draft then the fragment's block type is likely going to be `unstyled`.

If we use the fragment's block type in this case then the result of the paste is that the block's text is changed to `Hello great world` and the block is changed so that it's no longer a list.

That, I think, would be an unpleasant experience.

tomkelsey:
Thanks for the explanation @colinjeanne - that makes sense as to why it is the way it is then!

I'm afraid I'm not familiar enough with DraftJS' workings to even begin to suggest something that would work in both instances. By that, I mean I am currently attempting to copy and paste an image. I parse the HTML in `handlePastedText` and set the block type of the image to being `atomic` but when replacing the current selection with it using `replaceWithFragment` it loses it's block type and doesn't function as intended.

kaylee42:
I recognize that the current draft behavior is desirable in the majority of circumstances; however, it would be nice to be able to manually override the current behavior of `replaceWithFragment` to insert only a single block to handle cases like the one @tomkelsey describes above. 

I have a custom block insertion util based on draft's `AtomicBlockUtils` (with variations to be able to insert multiple custom block types). The `AtomicBlockUtils` insertion function always [includes an empty, unstyled divider block](https://github.com/facebook/draft-js/blob/07eb9c4e2e2771d284c3c8f0c0d78d6b488619fd/src/model/modifier/AtomicBlockUtils.js#L73), and right now that's why my insertion function does as well. 

[...]

PR #1869, opened 2018-09-12
Fix issue #1868

marconi1992:
**Summary**
This PR fix the issue reported in #1868, the function `updateExistingBlock` into `src/model/transaction/insertFragmentIntoContentState.js` wasn't keeping the original block type.

**Test Plan**
I updated the test, so if the block type was unstyled and you paste a heading html element the new block should be heading instead unstyled.


yangshun:
Thanks for the PR @marconi1992! Please use a more meaningful PR title and a commit message that is clearer about what the PR is trying to achieve. 

The tests are failing on master now probably because of an update to Node/npm. I believe I fixed them in #1866 but you'll have to rebase first to get the fix.

marconi1992:
Thanks, @yangshun, the PR is updated and the related issue #1868

colinjeanne:
Also note that this is a breaking change to the API. Please see https://github.com/facebook/draft-js/issues/1511 for a case in which copying the fragment's block type is actually quite bad.

PR #1871, opened 2018-09-14
added favicon

### LINKED ISSUE #1857: Favicon is missing for website
The migration to Docusaurus broke the website's favicon.

This can probably be fixed by [adding the favicon in `siteConfig.js`](https://docusaurus.io/docs/en/site-config#mandatory-fields).

saranshkataria:
Fixed #1857 by adding a favicon. The color scheme of the icon needs to be thought about though since it is not visible when the tab is active because of the white color.

On chrome:
image

Looks okay when the tab is inactive, but when it is active, the logo is not visible:
image


claudiopro:
Hi @saranshkataria, thanks for contributing this fix! It does address #1857 indeed.

I'd rather use a black color for the icon, as white on transparent background is too faint. Or even a burgundy backdrop for the white icon. I'll chat with our designers to check which is the right approach!

PR #1875, opened 2018-09-17
add path to include Draft.css

hozefaj:
**Summary**
The documentation mentions that `Draft.css` needs to be included but no mention on how to do it. 

This update gives a way and also gives the path of the CSS to if folks are using `webpack` they can include it either style code directly rather than JS.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -73,4 +73,9 @@ Because Draft.js supports unicode, you must have the following meta tag in the `
 
 `Draft.css` should be included when rendering the editor. Learn more about [why](/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css).
 
+```js
+// It's located in node_modules/draft-js/dist/Draft.css.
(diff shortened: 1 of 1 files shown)


thibaudcolas:
Relevant: https://github.com/facebook/draft-js/pull/1589#discussion_r162636429

cojennin:
@hozefaj In #1589, @flarnie stated that a docs page would be accepted with tips or instructions for using Draft.js with [Create React App](https://github.com/facebook/create-react-app). Can you modify this PR to meet those requirements? I [recognize your concern](https://github.com/facebook/draft-js/pull/1875#issuecomment-423832533) and I think having a docs page for [Create React App](https://github.com/facebook/create-react-app) would help.

hozefaj:
@cojennin Yeah sure I can create a new docs page for that.

So you want the docs page to have example of create react app and draft js being used. I also propose maybe using codesandbox for example.

PR #765, opened 2016-11-05
Support rendering Draft.js into iframes (nested browsing contexts)

acusti:
### Summary

Removes Draft’s dependency on the global `window` and `document` objects and makes it work across nested browsing contexts (iframes) by using the document and window relative to the browsing context in which Draft is being used. React itself makes similar assumptions about being able to use the global `window` and `document` objects, as documented in facebook/react#427, so this PR depends on facebook/react#9184. All together, these changes make it possible to render Draft.js inside an iframe, which in turn makes it possible to have a UX where the page includes multiple visible selections, like [in this demo](http://codepen.io/acusti/pen/RGEJZE?editors=0011). In that demo, I used built versions of React and Draft.js which include all the changes necessary to make it work across nested browsing contexts.

Fixes #527

### Test Plan

All of the unit tests are passing. For manual testing, I’ve focused on verifying the behavior of the editor when dealing with selections. That includes:

- Select a portion of text and start typing to replace it
- Select a portion of text and hit delete to remove it
- Select a portion of text and paste in content to replace it
- Select a portion of text and copy it, then paste it somewhere
- Select a portion of text and cut it, then paste it somewhere

I then:

- Repeated the steps above with a collapsed selection (for the insertion step)
- Repeated the steps above with a selection that spans multiple blocks

Finally, I manually tested all of those behaviors with:

- A regular editor implementation where the editor is inside the same browsing context as the rest of the React app
- An implementation where the editor is rendered into an iframe within the parent browsing context where React is running, like with the codepen demo I linked to above

In order to make it easy to try and test these behaviors, I created installable branches of fbjs and React that include the `lib/` files to make it installable and usable via npm as well as the built files to be usable directly in the browser.

### Try it out via npm

In your package.json, use:

```js
    "draft-js": "Brandcast/draft-js#f9affa3",
    "fbjs": "Brandcast/fbjs-built#97b8e54",
    "react": "Brandcast/react#9a81d1a",
```

### Try it out in the browser

Include:
```html
    <script src="https://rawgit.com/Brandcast/react/nested-browsing-contexts-2-built/build/react.js"></script>
    <script src="https://rawgit.com/Brandcast/react/nested-browsing-contexts-2-built/build/react-dom.js"></script>
    <script src="https://cdn.jsdelivr.net/immutable.js/latest/immutable.min.js"></script>
    <script src="https://cdn.jsdelivr.net/es6.shim/0.35.1/es6-shim.min.js"></script>
[...]

PR #1877, opened 2018-09-21
Support rendering Draft-js into iframe

haikyuu:
**Summary**
This pull request adds support for draft-js usage inside iframes. There is another WIP PR  which tries to achieve the same thing #765. 

Note that this was mainly broken because `instanceof` does not check objects across iframes.
In other words (code) `iframeElement instanceof Element === false` while `iframeElement instanceof iframeDocument.Element === true`.

And the other reason is `getSelection` depends on the window object as well.

## Checking if a node is instance of `Element`, `HTMLElement` ...

Depending on the context, we need a way to access the right iframe window object to check using `instanceof`.

| ﻿Solutions                                                                             | D | Pros  👍                                                                 | Cons 👎                                                                                                                         | Shortcomings                                                                                |
|---------------------------------------------------------------------------------------|---|-----------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
| attach the iframeWindow to global object, and use it to check whenever it's available | 🛑 | - Easy access to iframe document                                            | - breaks when there is an instance of Draft-js inside the iframe and another outside.- breaks when there are +2 iframes as well |                                                                                             |
| get the document from the node itself and check                                       | ✅ | - Works fine even when there are many iframes or editors in and out iframes |                                                                                                                                 | - you can only access the global document when you have access to a node. (not needed here) |
| check the `nodeType`  property instead (used in [acusti's PR ](https://github.com/facebook/draft-js/pull/765))                         | 🛑 |                                                                             | - can't differentiate between Element and HTMLElement                                                                           |                                                                                             |

### Challenges

- When pasting something from a different context. We get a body element which document object doesn't have a corresponding  `window` / `defaultView` . It was fixed by adding another check.

## Accessing the window or document object in other cases `getSelection` ...
Whenever the `event` is accessible, we use it to access the `ownerDocument`

## Places skipped (they use the document and global objects)

- `getSafeBodyFromHTML` : it creates a new document anyway
- `getTextContentFromFiles` : it uses the `global.fileReader`

## Other challenges
In order to fix some flow errors, i had to force casting node types in some places. Note that i only did this after `instanceof` checks, so it's  safe. Those checks were not detected by flow because we use `instanceof` to check in iframe window or in the main window. One solution to fix this all together is to check nodes using a utility like [`fbjs`' `isNode` ](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/dom/isNode.js) but i prefer to keep the changes minimal.

## Known issues
- Inside an editor **in an iframe**: selecting some text and hitting `cmd+b` would apply the inline style but the selection will collapse. It happens only once, and **doesn't happen** outside an iframe. You can check it here and reproduce it in the [sandbox](https://codesandbox.io/s/y0q1q281kz)
draft-js-gif

**Test Plan**

I created [a sandbox](https://codesandbox.io/s/y0q1q281kz) to test `draft-js` editor. It has 2 editors inside separate iframes, and one in the main window.


claudiopro:
I also found an issue when the iframed Editor has a `placeholder` prop, the placeholder text is rendered as a text block and the caret is placed on the next line.

Repro: https://codesandbox.io/s/nk90w2jv9j

claudiopro:
@haikyuu do you need any help to address the review feedback? I'd like to merge this since it's in a good shape and there are many interested consumers!

claudiopro:
Hi @haikyuu, would you have time to look into the finishing touch to this PR?

Several Draft.js users who are eager to embed editors in an iframe would be thankful for your efforts 😄

marek-baranowski:
Hey @haikyuu, I'm also blocked by the iframe issue and I would love to see this PR merged :)

haikyuu:
@marek-baranowski @claudiopro continued the work on #1938 👍

niveditc:
Closing this since it's continued on #1938 instead.

PR #1883, opened 2018-09-30
Bring back the ariaOwneeID prop.

### LINKED ISSUE #1736: Bring back support for aria-owns
I've read https://github.com/facebook/draft-js/commit/7f0cab28386ec4bde8ec6289377bff9e53cd019b

I think there's probably a misunderstanding in what a "parent-child relationship between the owner and ownee" means in ARIA terms as opposed to DOM terms. ARIA doesn't do anything in the browser, nor changes the DOM of course. It changes the way browsers expose information to assistive technologies through the Accessibility APIs and the accessibility tree.

The combobox ARIA widget is a bit complicated. There are now two different valid patterns, the "traditional" one from ARIA 1.0 and the new one from ARIA 1.1. The differences between the two patterns are very well described in the [ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox).

Regardless, based on the specifications, `aria-owns` is a required property for the combobox ARIA widgets.

References:
ARIA 1.0
https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-owns
> The reason for adding aria-owns is to expose a parent/child **contextual** relationship to assistive technologies that is otherwise impossible to infer from the DOM.

ARIA 1.1
https://www.w3.org/TR/wai-aria-1.1/#aria-owns
> Identifies an element (or elements) in order to define a **visual, functional, or contextual parent/child relationship** between DOM elements where the DOM hierarchy cannot be used to represent the relationship.

Regardless of the specs, the ARIA authoring practices describe very well the implementation details. See under the combobox "WAI-ARIA Roles, States, and Properties"
https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-5 they describe both patterms (1.0 and 1.1)

1.1:
> The element with role combobox contains **or owns** a textbox element that has either role textbox or role searchbox.

Please notice the 1.1 pattern is made of a container with role=combobox and aria-owns, while the textbox element is a child and uses aria-controls, so this pattern is very different from the current implementation, for example, in the Mentions plugin.

1.0
- When the combobox popup is visible, the element with role combobox has aria-owns set to a value that refers to an element with role listbox.

Please also refer to the 5 W3C examples linked there: https://www.w3.org/TR/wai-aria-practices-1.1/#examples-1 they all use `aria-owns`.

The practical consequence of the missing aria-owns attribute is that some browser / screen reader combinations don't read out the suggestions at all. For example, Safari and VoiceOver are not able to announce the options. See also https://github.com/draft-js-plugins/draft-js-plugins/issues/1073

afercia:
**Summary**

Reverts https://github.com/facebook/draft-js/commit/7f0cab28386ec4bde8ec6289377bff9e53cd019b to bring back the `ariaOwneeID` prop. The ARIA property `aria-owns` is required for the combobox ARIA widget, in both ARIA 1.0 and 1.1. Some plugins, for example the Mentions and Emoji plugins, still pass an `ariaOwneeID` prop which at the moment doesn't do anything. Also, without `aria-owns`, some browser / screen reader combinations don't read out the suggestions at all. For more details please refer to the related issue #1736.

Fixes #1736.


niveditc:
This looks good to me! @jessebeach, can you confirm since you're more familiar with this area? :)

cdelahousse:
Sorry for the noise. Just testing some stuff on our end with GitHub integration.

PR #1890, opened 2018-10-08
Use flow-typed for flow checks in jest

niveditc:
**Summary**

I enabled strict flow checks in bbd3ef102b5191ef8a58edd04997a3c345e291b7 and this broke the flow check in Travis CI because jest mocks are unable to correctly use flow. I referenced this [Stack Overflow thread](https://stackoverflow.com/questions/35898251/whats-the-right-way-to-write-jest-tests-verified-with-flow) for Jest which says that importing `flow-typed` is the way to go for this.

The only other thing we could do is not flow strict this mock file, though I feel that we will probably run into the issue again.

**Test Plan**

```
yarn run flow
yarn run lint
yarn run test
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -9,3 +9,4 @@ scripts/jest/preprocessor.js
 scripts/jest/hasteImpl.js
 examples/draft-0-10-0/universal/static/bundle.js
 examples/draft-0-9-1/universal/static/bundle.js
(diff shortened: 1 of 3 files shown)


niveditc:
Actually, I think this bloats the project quite a bit. Let's hold off for now & I'll unstrict the file instead

Issue #1843, opened 2018-08-23
ContentBlock Keys with "-" in them cause the editor to crash


nayomitchell:
**Do you want to request a *feature* or report a *bug*?** 
Report a bug

**What is the current behavior?**
Passing a string with a "-" character within it causes the editor to fail when getting the new selection state. It looks like the editor adds a suffix to the content block key (Ex. "key-0-0"), then removes the suffix by splitting on the first "-" character.
That leads to this error when the malformed key is used to retrieve the BlockTree:
![image](https://user-images.githubusercontent.com/34354914/44541514-0ead4180-a6bf-11e8-9163-5a9a83d30d50.png)

**What is the expected behavior?**
The editor doesn't crash due to specific characters in the key.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
I'm able to repro this in the most recent versions of Chrome and Firefox. We're using v0.10.5 of draft.js.

We're currently getting around this by replacing any "-" characters with another before creating ContentBlocks.

niveditc:
@nayomitchell - thanks for reporting this issue. I think the block keys are intended/assumed to be alphanumeric, though I don't see that documented anywhere.

The recommended way to generate a new block key is using the `generateRandomKey` method defined in https://github.com/facebook/draft-js/blob/master/src/model/keys/generateRandomKey.js (it converts a random number into a string in base 32, resulting in only characters 0-9 & a-v).

I'll leave this issue open. **My recommended fix is adding to the docs to clarify that keys are alphanumeric & that `generateRandomKey` is recommended.**

erick2014:
Hey guys, I would like to work on this :)

cojennin:
@erick2014 given what @niveditc recommended above it doesn't look like anything needs to be validated. It seems like adding a sentence or two to [this section of the documentation](https://draftjs.org/docs/api-reference-content-block#getkey) that explicitly states block keys are alphanumeric and recommends using `generateRandomKey` to create block keys would be a good approach

niveditc:
#1892 fixed this issue. Thanks for the PR, @flowkraD 👍

PR #1892, opened 2018-10-10
Added information about nature of block keys

flowkraD:
**Summary**

Added a note letting users know the block keys are strictly alphanumeric and using the `generateRandomKey` function to generate block keys is a good approach.

Any changes needed can be mentioned and will be implemented in a jiffy!

Issue in focus: #1843


flowkraD:
My pleasure (^_^)

PR #1938, opened 2018-11-19
Take over enhancements to render Draft.js in an iframe

claudiopro:
**Summary**

Takes over and applies requested PR feedback to #1877 by @haikyuu.

Refer to original PR for context and discussion.

**Test Plan**

Run some manual smoke tests on the editor running in an iframe as follows:

```
yarn
python3 -m http.server 8080 .
open http://localhost:8080/examples/draft-0-10-0/iframe/iframe.html
```


haikyuu:
Thanks for taking it on @claudiopro.
I've been quite busy these days. 👍

claudiopro:
Thanks to you @haikyuu  for working out most of the changes! I just took it from there 😄

odusseys:
I'm pretty sure there's is a problem with the way you destructure your event targets (eg. I'm having errors in `DraftEditorDragHandler.js`) to get the `ownerDocument`. For instance, if your target is the document itself, `ownerDocument` will be null, in which case you need to fallback to the target itself.

claudiopro:
Thanks for the heads up @odusseys, do you mind sharing steps to reproduce the error?

odusseys:
Eh I'm not a hundred percent sure anymore since I fixed it on my fork and don't really have time to reproduce it; but I think I just added the drag plugin from draft-js-plugins together with the image plugin, and dragging threw an error (ownerDocument was null).
Either way this is the normal behavior (https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) so null should be checked either way.

haikyuu:
Hey @claudiopro what's the status on this one?

cosmophile:
Any updates on this y'all?

cosmophile:
@claudiopro @haikyuu thank you for your work on this. I was able to make it work pretty seamlessly with the latest version of Draft by replicating your commits. (Just needed to add an isHTMLBRElement check in the style of isHTMLAnchorElement etc.)

Issue #1940, opened 2018-11-20
Remove references to React lifecycle methods deprecated in 16.3


claudiopro:
React components in Draft.js still use or reference lifecycle methods that were deprecated in React 16.3 because if misused, they can get in the way of asynchronous rendering.

The deprecated methods are:
```
componentWillMount
componentWillReceiveProps
componentWillUpdate
```

[This post on the React blog](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#migrating-from-legacy-lifecycles) provides an excellent migration guide and offers common scenarios with recipes to replace the legacy lifecycle methods with the new API.

It seems we only have references in examples or in code comments, so this makes it an easy first bug!

danedavid:
Can I take this one up?

claudiopro:
Hi @danedavid, thanks for taking a look! I am afraid this issue is closed already with #1952. Please check back again for issues tagged with "good first issue".

PR #1952, opened 2018-12-06
Remove componentWillReceiveProps usages in examples

chaitanya-deep:
**Summary**
Resolves [1940](https://github.com/facebook/draft-js/issues/1940)
Change usage of `componentWillReceiveProps` in Tex examples to `componentDidUpdate`

### DIFF
--- a/examples/draft-0-9-1/tex/js/components/TeXBlock.js
+++ b/examples/draft-0-9-1/tex/js/components/TeXBlock.js
@@ -35,7 +35,7 @@ class KatexOutput extends React.Component {
     this._update();
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     if (prevProps.content !== this.props.content) {
       this._update();
     }
--- a/examples/draft-0-9-1/tex/js/components/TeXBlock.js
+++ b/examples/draft-0-9-1/tex/js/components/TeXBlock.js
@@ -21,36 +21,24 @@ import {Entity} from 'draft-js';
 class KatexOutput extends React.Component {
   constructor(props) {
     super(props);
-    this._timer = null;
   }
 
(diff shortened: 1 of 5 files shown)


claudiopro:
Hi @chaitanya-deep, welcome to Draft.js and thanks for taking the time to look into this issue! This change looks good to me 👍🏼, sounds good to remove the sketchy `setTimeout`s also.

examples/draft-0-9-1/tex/js/components/TeXBlock.js
```diff
@@ -21,38 +21,26 @@ import {Entity} from 'draft-js';
...
+    );
   }
 
   componentDidMount() {
     this._update();
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.content !== this.props.content) {
+  componentDidUpdate(prevProps) {
```
claudiopro: Maybe worth calling out there's an unused `prevState` second argument here.

chaitanya-deep: Made the change but is that necessary?
The third `snapshot` argument is also unused.

And the docs for typical usage of componentDidUpdate don't include it either
https://reactjs.org/docs/react-component.html#componentdidupdate

package.json
```diff
@@ -85,7 +85,7 @@
     "webpack-stream": "^4.0.0"
   },
   "resolutions": {
-    "gulp/**/natives": "1.1.3"
+    "gulp/**/natives": "1.1.6"
```
niveditc: @chaitanya-deep - can we split this part into a separate PR? Just so we don't conflate two different changes `componentWillReceiveProps` & dependency updates. 

Also for the dependency update, I think updating the `yarn.lock` file is a good idea.

PR #1978, opened 2019-01-16
Update fbjs to 1.0.0 to fix ReDos Vulnerability

anthonymuau:
**Summary**
draft-js was using older version of fbjs that uses older version ua-parser-js what has a vulnerability.

https://app.snyk.io/vuln/npm:ua-parser-js:20180227

**Test Plan**

1. Run npm test and it passes
2. Run npm build and it passes


anthonymuau:
> Thank you for upgrading this!
> 
> * Can you please make sure the Travis CI is passing. Not sure why this is breaking - maybe it's unrelated to this PR?
> * I think these changes need to be propagated to the `yarn.lock` file too & running `yarn` will update this for you.

I updated the yarn lock file but it looks like flow type for joinClasses in fbjs is incorrect as it's typed as joinClasses(className: mixed): string but the correct type would be joinClasses(...className: Array<string>) as joinClasses function take an array of args and pass them inside of variable called className. because of this i had to add a flow fix me.

PR #1980, opened 2019-01-21
Adds missing onMouseDown callback -- closes #1972

### LINKED ISSUE #1972: _onMouseDown not used
Hi,

I was looking through the code and saw that the `DraftEditor.react.js` file defined a `_onMouseDown` function [here](https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditor.react.js#L224), though that function is not actually being registered with the component. This seemed strange to me since every other handler was registered except this one. It's not being used anywhere. I'm sure this is not causing any bugs at the moment, but it seems like an oversight, so I wanted to create an issue so it's on the radar.

I'd be happy to submit a quick pull request if this is a bug.

brenmcnamara:
**Summary**

Minor fix: the _onMouseDown callback that was created was never attached to the react component.

**Test Plan**

Make sure everything runs and compiles as before. Because onMouseDown was never being used, this should not cause any changes to Draft JS behavior.


fabiomcosta:
@brenmcnamara @niveditc I think the right fix here would be to to remove the `onMouseDown` creation altogether, since it's never used anywhere.

PR #1995, opened 2019-02-06
Fix bad destructuring when content block key has a `-`

jankdc:
**Summary**

If a content block was created with a key that contains the `-`, it would fail to decode the `decoratorKey` and the `leafKey` because the current implementation relies on `-` to be a parsing delimiter. Here's an example:

```js
class SomeComponent extends React.Component {
  constructor(props) {
    super(props);

    const block = new ContentBlock({
      type: 'paragraph',
      text: 'some-text',
      key: 'some-key'
    });

    const contentState = ContentBlock.createFromBlockArray([block]);

    this.state = {editorState: EditorState.createWithContent(contentState)};

    this.onChange = editorState => this.setState({editorState});
  }

  render() {
    return (
      <Editor
        onChange={this.onChange}
      />
    );
  }
}
```

When you select anywhere on the editor, you'd get this error:

```
    TypeError: Cannot read property 'getIn' of undefined

      45 |   const focusBlockKey = focusPath.blockKey;
      46 |   const focusLeaf = editorState
    > 47 |     .getBlockTree(focusBlockKey)
         |                              ^
      48 |     .getIn([focusPath.decoratorKey, 'leaves', focusPath.leafKey]);
      49 |
      50 |   const anchorLeafStart: number = anchorLeaf.get('start');

      at getUpdatedSelectionState (src/component/selection/getUpdatedSelectionState.js:47:30)
      at getDraftEditorSelectionWithNodes (src/component/selection/getDraftEditorSelectionWithNodes.js:48:58)
      at getDraftEditorSelection (src/component/selection/getDraftEditorSelection.js:37:53)
      at assertGetDraftEditorSelection (src/component/selection/__tests__/getDraftEditorSelection-test.js:55:53)
      at Object.<anonymous> (src/component/selection/__tests__/getDraftEditorSelection-test.js:211:3)
```

**Test Plan**

Added a unit test for `DraftOffsetKey.js` to check for all delimiter cases.

### DIFF
--- a/src/component/selection/DraftOffsetKey.js
+++ b/src/component/selection/DraftOffsetKey.js
@@ -25,7 +25,15 @@ const DraftOffsetKey = {
   },
 
   decode: function(offsetKey: string): DraftOffsetKeyPath {
-    const [blockKey, decoratorKey, leafKey] = offsetKey.split(KEY_DELIMITER);
+    const keyComponents = offsetKey.split(KEY_DELIMITER);
+    const blockKey = keyComponents
+      .slice(0, keyComponents.length - 2)
+      .join(KEY_DELIMITER);
+
+    const [decoratorKey, leafKey] = keyComponents.slice(
+      keyComponents.length - 2,
+    );
+
     return {
       blockKey,
       decoratorKey: parseInt(decoratorKey, 10),
--- a/src/component/selection/DraftOffsetKey.js
+++ b/src/component/selection/DraftOffsetKey.js
@@ -25,17 +25,14 @@ const DraftOffsetKey = {
   },
 
   decode: function(offsetKey: string): DraftOffsetKeyPath {
-    const keyComponents = offsetKey.split(KEY_DELIMITER);
-    const blockKey = keyComponents
-      .slice(0, keyComponents.length - 2)
-      .join(KEY_DELIMITER);
-
-    const [decoratorKey, leafKey] = keyComponents.slice(
-      keyComponents.length - 2,
-    );
+    // Extracts the last two parts of offsetKey and captures the rest in blockKeyParts
+    const [leafKey, decoratorKey, ...blockKeyParts] = offsetKey
+      .split(KEY_DELIMITER)
+      .reverse();
 
     return {
-      blockKey,
+      // Recomposes the parts of blockKey after reversing them
+      blockKey: blockKeyParts.reverse().join(KEY_DELIMITER),
       decoratorKey: parseInt(decoratorKey, 10),
       leafKey: parseInt(leafKey, 10),
     };
--- a/src/component/selection/__tests__/DraftOffsetKey-test.js
+++ b/src/component/selection/__tests__/DraftOffsetKey-test.js
@@ -6,6 +6,7 @@
  *
  * @emails oncall+draft_js
  * @format
+ * @flow strict-local
  */
 
 jest.disableAutomock();
--- a/src/component/selection/__tests__/DraftOffsetKey-test.js
+++ b/src/component/selection/__tests__/DraftOffsetKey-test.js
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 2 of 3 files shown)


claudiopro:
Good catch @jankdc, thanks for contributing this fix to Draft.js! I can't find an issue associated with this PR, I'm curious how did you encounter this problem?

Please take a look at my comments inline, I think we can make the fix slightly easier to reason about.

src/component/selection/DraftOffsetKey.js
```diff
@@ -25,7 +25,15 @@ const DraftOffsetKey = {
...
-    const [blockKey, decoratorKey, leafKey] = offsetKey.split(KEY_DELIMITER);
+    const keyComponents = offsetKey.split(KEY_DELIMITER);
+    const blockKey = keyComponents
+      .slice(0, keyComponents.length - 2)
+      .join(KEY_DELIMITER);
+
+    const [decoratorKey, leafKey] = keyComponents.slice(
+      keyComponents.length - 2,
+    );
+
```
claudiopro: I personally find this logic a bit convoluted and difficult to reason about. What about the following instead?

```js
    // Extracts the last two parts of offsetKey and captures the rest in blockKeyParts
    const [leafKey, decoratorKey, ...blockKeyParts] = offsetKey
      .split(KEY_DELIMITER)
      .reverse();

    return {
      // Recomposes the parts of blockKey after reversing them
      blockKey: blockKeyParts.reverse().join(KEY_DELIMITER),
      decoratorKey: parseInt(decoratorKey, 10),
      leafKey: parseInt(leafKey, 10),
    };
```

jankdc:
Oh sorry, I didn't really create an issue for it. We encountered this problem because we were manually mapping our tree data structure to a list of `ContentBlock`s , where each branch or leaf, had a GUID id, like so:

```js
const data = {
  id: "0dca5e9d-2ebd-4574-921e-2b68793c9e8c",
  sections: [
    {
      id: "4b35e274-81e6-4eb5-8964-ef566517051d",
      text: "Some-text",
      meta: { ... },
      sections: [ ... ]
    }
  ]
}; 

function reduceTree(data): ContentBlock[] { ... }
```

The reason we're going this route is because we want to keep some metadata in the blocks instead of using `convertFromHTML` which strips it all away. Anyway, that's where we came across it.

src/component/selection/DraftOffsetKey.js
jankdc: Yeah, that's much nicer. Thank you!

claudiopro:
Ah, that makes sense. Maybe we should make the separator token configurable. Thanks again for contributing this fix!

src/component/selection/__tests__/DraftOffsetKey-test.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @emails oncall+draft_js
+ * @format
+ */
+
+jest.disableAutomock();
+
+const DraftOffsetKey = require('DraftOffsetKey');
+
+test('decodes offset key with no delimeter', () => {
```
claudiopro: Small typo:
```suggestion
test('decodes offset key with no delimiter', () => {
```

src/component/selection/__tests__/DraftOffsetKey-test.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails oncall+draft_js
+ * @format
```
claudiopro: Do you mind adding `@flow strict-local` to this file? Thanks!

jankdc:
> Or even better, allow arbitrary payload to be associated with blocks e.g. via `data-` attributes. 

Yeah, this would be a really handy feature.

SeanSilke:
Hello when this fix will be released?   Do you Have "release circle", there is near three months past from last release.

PR #2002, opened 2019-02-13
[Docs] Rename DraftEditorBlock to EditorBlock

umanghome:
**Summary**

The `DraftEditorBlock` component is exposed to the public API as `EditorBlock`, while the [documentation](https://draftjs.org/docs/advanced-topics-block-components) still mentions `DraftEditorBlock`.
https://github.com/facebook/draft-js/blob/b4183b1519edcd075325adae0cb208551dc80f50/src/Draft.js#L45


claudiopro:
Good catch @umanghome, thanks for contributing a fix! 💯

Issue #2005, opened 2019-02-18
typing Korean in linux, chrome/electron is buggy


palindrom615:
**Do you want to request a *feature* or report a *bug*?**

bug.

**What is the current behavior?**

the first korean letter after `compositionend` event is ignored.

![draft-js-kor](https://user-images.githubusercontent.com/10936496/52938946-4a8f3e00-33a6-11e9-9157-73981a1bd521.gif)

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

after commit 07b1eb847fb801d61b41a7a03c7c26c80cca5478. I have tested windows and fedora 29 and the bug happened only on chrome or electron in linux fedora 29.

I think I found the reason why bug happens.

![left one: jsfiddle showing order of occuring event in chrome, windows. right one: same one in linux.](https://user-images.githubusercontent.com/10936496/52939323-3861cf80-33a7-11e9-9b8c-a554101ca325.png)

at chrome in linux, `beforeinput` event occurs between `compositionend` and following `compositionstart`, unlike that of windows. So when typing new Korean character, `beforeinput` event handler of mode `edit`
https://github.com/facebook/draft-js/blob/3400cdafb465640c41d3728d52ed59e8f13ad78a/src/component/handlers/edit/editOnBeforeInput.js#L80
is called instead of mode `composite`.
https://github.com/facebook/draft-js/blob/3400cdafb465640c41d3728d52ed59e8f13ad78a/src/component/handlers/composition/DraftEditorCompositionHandler.js#L48

yeah that is chromium's own bug, but we can bypass the bug by *not to* change mode from `composite` to `edit`.

https://github.com/facebook/draft-js/blob/3400cdafb465640c41d3728d52ed59e8f13ad78a/src/component/handlers/composition/DraftEditorCompositionHandler.js#L90

The mode is changed because of invalid `stillComposing` flag, so by removing the if statement above, the bug fixed.

Actually, I cannot catch logic of the statement or even comments. Can @sophiebits  or anyone help me understand the code? At least I cannot find any bug without that part. Maybe because I don't have any macOS...

palindrom615:
I think this issue is not so trivial because I've found some people complaining about writing facebook post or messenger with linux and chrome...

sophiebits:
IIRC Safari on Mac needs that for typing to work in Korean. Unfortunately every browser sends different events and it is really hard to know what will work on all of them.

palindrom615:
[...]

PR #2007, opened 2019-02-18
Fiexes issue #2005

palindrom615:
**Summary**

This PR solves issue #2005 which is bug occurs when typing Korean text in Linux and chrome.

**Test Plan**


fabiomcosta:
I think this PR requires an investigation on when/why those lines were added in the first place.
After that it would also require a clear test plan on how we could reproduce the issue.
I'd also recommend adding some unit tests to back this PR up even further.
@palindrom615 can you take a look at these items?

After those items are covered, someone at Facebook might also want to do some manual regression tests before merging.

PR #2008, opened 2019-02-19
docs: remove --save flag

pvsmounish:
--save is no longer needed.


claudiopro:
Good catch, as [per the npm CLI docs](https://docs.npmjs.com/cli/install), `npm install` saves packages to dependencies by default. Thanks for this fix!

PR #1176, opened 2017-04-25
Prevent clobbering inline style overrides if the selection position is the same

tjaneczko:
This PR is meant to address #696 and to mitigate some of the issues of #726 .  The issue stems from the fact that `updateSelection` will automatically clear the `inlineStyleOverride` whenever it's called.  `updateSelection` is called both on blur and on focus of the editor (and in various other places), which causes any button presses for setting an inline style to completely clear the inline style overrides.  This PR checks to see if the position of the selection changes and will not clear the `inlineStyleOverride` if the positions have not changed.

Please let me know if this needs any adjustment or if more tests should be added to check for this functionality.

PR #2022, opened 2019-03-04
Fix inline style

mmissey:

4. Ensure that:

-

**Summary**

This is a small addition to https://github.com/facebook/draft-js/pull/1176, adding tests and updating `var` to `const`.  The original PR is nearly 2 years old and my company desperately needs this change, however we'd prefer not to keep our own fork of Draft maintained forever. 

This addresses an issue in which `EditorState.updateSelection` discarded inlineStyleOverride on any update, including a change in `hasFocus`. This caused a bug in which you could not have a user input a hex color in one text field, and then type in that color in DraftJS. The change in focus would remove the previously set inlineStyleOverride. In some cases, developers could work around this by using `onMouseDown` rather than `onClick` to toggle styles, however this doesn't work with text inputs. 

Here is a JSFiddle showing the problem with a bold toggle: https://jsfiddle.net/mmissey/3mq62gu8/

**Test Plan**

A snapshot test has been added to ensure that inlineStyleOverrides are not discarded when only focus changes, and another test ensures that inlineStyleOverrides *are* discarded when the actual position of the selection has been changed.

### DIFF
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -505,7 +505,7 @@ function updateSelection(
   forceSelection: boolean,
 ): EditorState {
   // Only unset the nativelyRenderedContent and inlineStyleOverride if the
-  // position is different
+  // position is different, not on blur/focus.
   const editorSelection = editorState.getSelection();
   if (
     editorSelection.getAnchorKey() === selection.getAnchorKey() &&
--- a/src/model/immutable/EditorState.js
+++ b/src/model/immutable/EditorState.js
@@ -504,6 +504,21 @@ function updateSelection(
   selection: SelectionState,
   forceSelection: boolean,
 ): EditorState {
+  // Only unset the nativelyRenderedContent and inlineStyleOverride if the
+  // position is different
+  const editorSelection = editorState.getSelection();
+  if (
+    editorSelection.getAnchorKey() === selection.getAnchorKey() &&
+    editorSelection.getAnchorOffset() === selection.getAnchorOffset() &&
+    editorSelection.getFocusKey() === selection.getFocusKey() &&
+    editorSelection.getFocusOffset() === selection.getFocusOffset() &&
+    editorSelection.getIsBackward() === selection.getIsBackward()
+  ) {
+    return EditorState.set(editorState, {
+      selection,
+      forceSelection,
+    });
+  }
   return EditorState.set(editorState, {
     selection,
     forceSelection,
--- a/src/model/immutable/__tests__/EditorState-test.js
+++ b/src/model/immutable/__tests__/EditorState-test.js
@@ -181,6 +181,22 @@ test('does not discard style override when splitting block', () => {
   expect(editor.getCurrentInlineStyle().toJS()).toMatchSnapshot();
 });
 
+test('does not discard style override when gaining/losing focus', () => {
+  let editor = UNDECORATED_STATE;
+  editor = EditorState.moveFocusToEnd(editor);
+  editor = RichTextEditorUtil.toggleInlineStyle(editor, 'BOLD');
+  editor = EditorState.moveFocusToEnd(editor);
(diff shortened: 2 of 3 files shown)


mmissey:
I'm not sure how my changes could have possibly caused Flow errors. Is there a problem with CI?

fabiomcosta:
Hi @mmissey the issue is still not clear to me and I don't know how to reproduce it on your example.

Can you please provide:
* Clear steps on how I can reproduce the issue on your example and what would be the expected behavior vs what currently happens
* What devices/browsers this issue happens

mmissey:
Hi @fabiomcosta,

Steps to reproduce in this fiddle: https://jsfiddle.net/mmissey/3mq62gu8/

- Begin typing in the editor, InlineStyleOverride is set to `null` so the flag shows "bold: off"
- Click the "Toggle Bold" button, notice that the flag shows "Bold: on" since the inlinestyle override is set to `BOLD`
- The flag will switch from  "bold: on" back to "bold: off" as soon as focus is placed back into the editor. (300ms timeout in order to see the change happening)

The expected behavior is that I can select a new style (bold), and continue to type in that style. 

This is not currently possible as any change in selection state, including a change of `hasFocus` will completely discard any inlineStyleOverride. Discarding inlineStyleOverride makes sense if the cursor is in a new position or the select, but when the cursor remains in the same position I would expect to have  keep my inlineStyleOverride.

fabiomcosta:
@mmissey I see the issue and could reproduce.
I was trying to compare your example with the `rich/rich.html` example included on the draft-js repo, that can be see on the main page: https://draftjs.org/
There I don't have this issue.
I see that you are using draft 0.7.0 on your JSFiddle, but even when I tried using 0.10.5 I still get the same issue.
Do you see any difference between these 2 examples?

fabiomcosta:
@mmissey I found the differences and made your example work: https://jsfiddle.net/Lzxt0v4g/

Here is what I did:
1- removed the setTimeout
2- on `BoldToggler` used onMouseDown instead of onClick and the handler is`this.toggleBold` instead of `this.props.toggleBold`

That being said, the fact that `inlineStyleOverride` is always set to null there does feel like a bug to me. I think I'm +1, but I'm not from the core team so I can't really merge this, hopefully someone from their team will help get this merged.

mmissey:
Hey @fabiomcosta,

Thanks for looking closer. The `onMouseDown` workaround was mentioned in a few other issues, and we were able to hack around it a bit, however this does not work with `<input type='text'>`. We need the ability for the user to enter their own hex code to set an inlineStyleOverride that sets a color. 

With the current DraftJS version, when the user clicks on the input field to type, focus is lost. After the user types the hex code they must click on the editor again and that re-focus clears the inlineStyleOverride.

We are currently using this modified version of DraftJS to fix our problem. 

Also, if you look at the PR that I linked to you will see that this issue and fix has already been approved, it just needed some test and a small update. That's all that I did here.

mmissey:
Can an authorized contributor please take a look at this one? The original PR is 2 years old. This one, with cleanup and tests is just 3 weeks.

kamenminkovcom:
I believe [this](https://github.com/facebook/draft-js/issues/2079) is the same issue. We will also like to see it merged instead of working around the issue. As @mmissey said, `onMouseDown` with a `preventDefault` is a kind of solution but it won't work with more complex types of input.

Sigurdlex:
Hello, everyone! First of all, thank you for awesome text editor!
Could some authorized contributor please merge this PR? I'm starting to work with draft.js and having the same issue with discarding InlineStyleOverride after I return focus to editor. Workaround with `preverntDefault` doesn't work in my case, because I need to implement font change via dropdown, it's much more complex than simple button.

richardtagger:
Hey @claudiopro I saw you recently active in this repo and wondered if you'd be able to take a look. Would be good to get this fix into the next release :)

mrkev:
hmmm I think this makes sense.

PR #2030, opened 2019-03-08
Fix npm run dev

fabiomcosta:
**Summary**

Fixes the `npm run dev` command, which currently fails when run.

**Test Plan**

`npm run dev`
`yarn run dev`


fabiomcosta:
I sent an email based on the feedback from the cla-bot, waiting.

gulpfile.js
```diff
@@ -269,14 +271,14 @@ gulp.task(
...
+    gulp.watch(paths.src, gulp.parallel('modules'));
   }),
 );
 
 gulp.task(
   'dev',
   gulp.series(function() {
-    gulp.watch(paths.src, ['modules', 'dist']);
+    gulp.watch(paths.src, gulp.parallel('modules', 'dist'));
   }),
```
claudiopro: Isn't this redundant? The `dist` task has a dependency on `modules`, and it should be run in series, not in parallel anyway.

fabiomcosta: done

PR #2045, opened 2019-03-27
Exports isSoftNewlineEvent as static method of KeyBindingUtil

### LINKED ISSUE #2044: isSoftNewlineEvent is not available in exports
Bug

Currently I have to do the following to use isSoftNewlineEvent:
```javascript
import isSoftNewlineEvent from 'draft-js/lib/isSoftNewlineEvent';
```
I should be able to use this function like so:
```javascript
import { isSoftNewlineEvent } from 'draft-js';
```
or 
it should be attached to RichUtils,
```javascript
import { RichUtils } from 'draft-js';
...
RichUtils.isSoftNewlineEvent(...);
```
or KeyBindingUtil
```javascript
import { KeyBindingUtil } from 'draft-js'
...
KeyBindingUtils.isSoftNewlineEvent(...);
```

khawkinson:
Exports isSoftNewlineEvent as static method of KeyBindingUtil, so it can be imported as 

```js
import { KeyBindingUtil } from 'draft-js';
// ...
KeyBindingUtil.isSoftNewlineEvent(...);
```

Fixes #2044

### DIFF
--- a/src/component/utils/KeyBindingUtil.js
+++ b/src/component/utils/KeyBindingUtil.js
@@ -13,9 +13,9 @@
 
 const UserAgent = require('UserAgent');
 
-const isOSX = UserAgent.isPlatform('Mac OS X');
+const isSoftNewlineEvent = require('isSoftNewlineEvent');
 
-const isSoftNewlineEvent = require('./isSoftNewlineEvent');
+const isOSX = UserAgent.isPlatform('Mac OS X');
 
 const KeyBindingUtil = {
   /**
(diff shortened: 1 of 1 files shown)


src/component/utils/KeyBindingUtil.js
```diff
@@ -15,6 +15,8 @@ const UserAgent = require('UserAgent');
 
 const isOSX = UserAgent.isPlatform('Mac OS X');
 
+const isSoftNewlineEvent = require('./isSoftNewlineEvent');
```
thibaudcolas: Based on how imports are done in other files, I think this would be better right underneath `const UserAgent = require('UserAgent');` above, and using the custom `require` format with only the module name, 

```suggestion
const isSoftNewlineEvent = require('isSoftNewlineEvent');
```

PR #2049, opened 2019-03-30
Correct warning condition

unshift:
**Summary**

To my knowledge, ```warning``` works the way ```invariant``` does; it warns if the first argument is false not true.

Prior to this PR, ```moveSelectionBackward``` warns when the selection is collapsed when it should do the opposite.

### DIFF
--- a/src/component/handlers/edit/commands/moveSelectionBackward.js
+++ b/src/component/handlers/edit/commands/moveSelectionBackward.js
@@ -31,7 +31,7 @@ function moveSelectionBackward(
   const selection = editorState.getSelection();
   // Should eventually make this an invariant
   warning(
-    !selection.isCollapsed(),
+    selection.isCollapsed(),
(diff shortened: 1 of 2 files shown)


claudiopro:
Hey @unshift, welcome to Draft.js and thanks for spending the time to submit a PR!

You're absolutely right, the `warning` function in fbjs prints a warning if the condition is false:

https://github.com/facebook/fbjs/blob/11f270b715ec7b39cba6a6ba8a72dcc3d0462e57/packages/fbjs/src/__forks__/warning.js#L44

Therefore, in order to warn when the selection is not collapsed, the first argument must be `selection.isCollapsed()`.

claudiopro:
I reproduced the problem in the examples. This is good to merge.
Screen Shot 2019-03-31 at 12 20 00 PM

fabiomcosta:
Good to know those were incorrect! Thank you!

Issue #494, opened 2016-06-27
Error if there is no property "text" in the DataTransfer object while dropping


delijah:
I am dragging and dropping image-, video and slideshow-objects into the draft-js editor. To give the drop handler the information he needs to insert that object i am serializing a data object into the DataTransfer object. That's why i do not need the text-property of the DataTransfer object. If there is no text-property set on the DataTransfer object, draft-js raises an error:

```
Uncaught TypeError: Cannot read property 'length' of null
  insertTextIntoContentState
  replaceText
  insertText
  insertTextAtSelection
  onDrop
  ...
```

Is there a way to avoid this error if i do not want to insert any text and handle the drop action by myself? Or is this a bug?

spicyj:
I think it would be reasonable to only call this.update() here if data.getText() returns non-null:

https://github.com/facebook/draft-js/blob/9ce1bab81cd006666a1a47b724e7c06f97c4ea1f/src/component/handlers/drag/DraftEditorDragHandler.js#L123

butchmarshall:
I have also encountered this while using Draft+[react-dnd](https://github.com/gaearon/react-dnd)

![image](https://cloud.githubusercontent.com/assets/339763/18869579/ed10f3b8-847a-11e6-9261-25e8865609b6.png)

delijah:
Any progress on this?

goodhyun:
@delijah this quick fix solves the problem for a while.
https://github.com/facebook/draft-js/pull/864/commits/62ff707f73bb92716292df60e2d34cdb1ab656b0 as @spicyj recommended.

delijah:
Closing because of inactivity.

mattkrick:
does this need a PR? still running into this...

icd2k3:
Still running into this as well with react-dnd-enabled blocks

icd2k3:
NOTE: if you're using the `<Editor />` component from draft-js-plugins-editor you'll have to explicitly return `'handled'` instead of `true` for this to work

kaylee42:
I am still seeing this error - is there a timeline for that commit being merged?

delijah:
Since people still get this error and would like to see it fixed, i reopen this issue hereby.

efjacobson:
I am still seeing this error and it's a pain in the neck - I would love to see the fix implemented. Need a PR?

claudiopro:
This should be fixed by #2117. Please reopen if you believe this is not addressed @delijah

PR #2054, opened 2019-04-11
OnDrop don't insert if getText return null

quanbrew:
Fix https://github.com/facebook/draft-js/issues/494
-

**Summary**

When drop a custom draggable object to DraftEditor, draft will try get text from object.

It can be fixed by stop insert if getText return null.

**Test Plan**

Drag and Drop behaviors difficult to write a test case.


quanbrew:
signed

mrkev:
This gets addressed by https://github.com/facebook/draft-js/pull/2117, but I think it's a good idea to keep this check too

Issue #950, opened 2017-01-17
Upgrade Immutable to 3.8.1


djforth:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Current versions of Draft use Immutable 2.7.4

**What is the expected behavior?**
Would be useful to upgrade to latest version (3.8.1) to reduce dependancy mismatch (Working on project that needs 3.8.1, which means I need 2 versions of immutable).  Also is worth setting Immutable as a peer dependancy rather than a dependancy so people are aware if there is a mismatch.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
Had a quick look and 0.91, 0.10-stable & 0.10-alpha all depend on 2.7.4

Bleeky:
Completely agree, would be nice to have such an update of the dependencies. Plus, the #607 is still open, and this can cause some serious problems on some browsers.

djforth:
I did have a quick look at upgrading it fork, but looks like it isn't a trivial task.  Would be happy to have a go if the maintainers think it would be a useful thing?

quicksnap:
Related to #879

chrisui:
Could we declare Immutable as a `^` version rather than `~` so that we can be on *any* later minor version rather than a specific one? This would allow us to upgrade minor versions in our apps without having to get a new version of draft-js every time.

flarnie:
This would be great! If anyone wants to work on it, we'll make time to review the PR.

RomanLysogor:
Hey folks, any updates on this one? I found this issue as a performance critical. The logging message appears on each user typing, thus is slowing performance significantly.

mc-funk:
Seconded that an update on this would be helpful. The logging messages are really detrimental.

octatone:
Hi there, I just discovered that my bundled code contains two copies of immutable because of draft-js' old version dependency.  We are using immutable@3.8.1 in our code, draft-js@0.10.1 is immutable@3.7.6.

There are fixes for immutable Records between these two versions that we assert in tests in our code and that which would improve performance of draft-js as a whole.

shakdoesgithub:
Hi, wanted to see if there are any plans on doing this? As others have said, the console.warn is on every input changed and causes serious slowdown.

mitermayer:
Hi @shakdoesgithub ,

[...]

PR #2072, opened 2019-05-09
update immutable to the latest stable version (3.8.2)

raphamorim:
**Summary**

Update immutable to latest stable versions. The proposed version already have more than 2y since the release (https://www.npmjs.com/package/immutable/v/3.8.2). 

Also, `immutale@3.8.2` is an important version since this patch release relicenses it with the MIT license + contains many fixes over the years.

Screen Shot 2019-05-09 at 15 10 45

https://github.com/immutable-js/immutable-js/releases


JLarky:
@raphamorim https://www.npmjs.com/package/immutable/v/4.0.0-rc.12 is marked as `latest`, so it probably makes more sense to update to it.

thibaudcolas:
See #950. It’s not enough to just update the version number, you also need to update the Draft.js type annotations / API so type checking passes.

claudiopro:
This is a bit of a contentious topic. Internally to FB, we're moving away from immutable data types (citation needed) and towards lightweight data structures with immutability enforced via type annotations. However, we may move forward with this PR because it would require minimal changes to internal callsites in the www codebase.

mrkev:
I think we should merge this, but it's a little more complicated than it seems given how NPM deps work internally and how our repo is set up. I'll DM you @claudiopro.

If it were as easy as bumping the package.json I would like imagine @leebyron or someone else would've already done it 😛In any case, not an issue with the PR, just with how things are set up internally.

bj00rn:
any progress on this issue?

PR #2076, opened 2019-05-20
Adds support for nullish coalescing operator

claudiopro:
**Summary**

Adds support for nullish coalescing operator which was missing from the babel config of the OSS repo.

**Test Plan**

`yarn`
Verify Travis CI tests pass


package.json
```diff
@@ -32,6 +32,7 @@
     "test-ci": "cross-env NODE_ENV=test npm run lint && npm run flow && npm run test"
   },
   "dependencies": {
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
```
kedromelon: should this be in `devDependencies` since it should only be needed for dev/build? that's where the other babel stuff is pulled in

claudiopro: Good point, I'll change it.

claudiopro:
Thanks @NoamELB!

PR #2094, opened 2019-05-29
Force the selection when composition is resolved

### LINKED ISSUE #2093: Selection breaks when Chinese composition ends without changes in IE11
To reproduce:

* Use the latest version of Draft (i.e., one that includes https://github.com/facebook/draft-js/commit/634bd297f1646934087051b1149164b01a09a136)
* In IE11, open the rich text example
* Switch to Chinese (Simplified, PRC), and use the IME "Microsoft Pinyin New Experience Input Style". This one comes standard with Windows 7.
* Type `ni` and then press backspace twice.
* Type `ni` again.

Result: the cursor stays on the first (blank) line, whereas the letters `ni` and the Chinese completions appear on the next line. This is incorrect: there shouldn't be a blank line

* Type `1` and press enter.

Result: the editor now contains a blank line (with the cursor), followed by a line with the character `你`, followed by a line with the character `n`. This is incorrect: the editor should only contain `你` and nothing else.

* Type `wo1` and press enter.

Result: the editor now contains a line with `n`, followed by a line with the character `我`. This is incorrect: the character `你` has somehow disappeared and the `n` suddenly moved to the top of the editor.

The screencast below shows more clearly what happens:

screencast 2019-05-28 14-01-19

I believe that the editorstate breaks immediately after the step "Type `ni` and then press backspace twice", which is the first time a composition is resolved.

This might be related to #2080, but it's not the same. The fix that I proposed for that issue does *not* solve this one.

robbertbrak:
Fixes #2093.

### DIFF
--- a/src/component/handlers/composition/DraftEditorCompositionHandler.js
+++ b/src/component/handlers/composition/DraftEditorCompositionHandler.js
@@ -225,7 +225,7 @@ const DraftEditorCompositionHandler = {
 
     editor.restoreEditorDOM();
 
-    const editorStateWithUpdatedSelection = EditorState.acceptSelection(
+    const editorStateWithUpdatedSelection = EditorState.forceSelection(
       editorState,
       compositionEndSelectionState,
     );


fabiomcosta:
@robbertbrak for this one, we might want to do the same logic that we do at: https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnSelect.js#L41-L58
I'll see if I'm able to repro the reported issue.

fabiomcosta:
@robbertbrak I actually just double checked and even on the scenario that you specified needsRecovery is `false`, so that logic wont work.
I made sure this is all fine by also testing it on Android.
Your change LGTM.
cc @claudiopro

PR #2102, opened 2019-06-10
Block style fn

RomanLeonov:
**Summary**

I have use-case when need styling content block by set style object.
 E.g. I have possibility styling text content in numbering list, so need styling numbers also. Using class names aren't enough for this case.
I've extended props.blockStyleFn, in my version this function returns object  => {style: {}, className: ''}

I didn't find possibility to do this without fork.

May be to have other way to do this?


mrkev:
Closing due to inactivity. Feel free to comment for re-opening if there's interest in revisiting this PR.

PR #2106, opened 2019-06-13
Fixes #2104 issue when immutable entity is delete on compositionEnd

### LINKED ISSUE #2104: Entity editing broke with recent DraftEditorCompositionHandler updates
Recent updates to [DraftEditorCompositionHandler](https://github.com/facebook/draft-js/blob/master/src/component/handlers/composition/DraftEditorCompositionHandler.js) seems to have broken how entities are updated. 

Take this example from https://github.com/facebook/draft-js/blob/master/examples/draft-0-10-0/entity/entity.html

This screenshot demonstrates current behavior. The entity is not correctly updated when using 2-Set Korean and throws an error.
df498672882d65f74c595a85325085e1

Here is a screenshot of the behavior before the update
d752df6d747f1b7a8970726f720823f1

Let me know if you need anything more. For recreation, I use the 2-Set Korean on Mac for testing. Typing "tkfk" + "enter" produces the content you see in the screenshot. 

Recent updates to DraftEditorCompositionHandler do seem to solve a lot of issues, so hopefully this isn't discouraging.
 
+ @fabiomcosta

fabiomcosta:
When `getDraftEditorSelection` is called at https://github.com/facebook/draft-js/blob/master/src/component/handlers/composition/DraftEditorCompositionHandler.js#L220 `editorState` is already properly updated and without the removed immutable "Superman" entity (see reported issue for more context https://github.com/facebook/draft-js/issues/2104), but the DOM is still on the old state, so when `global.getSelection()` is called, the immutable entity is detected to be the leaf node of the selection, but since it doesn't exist on `editorState` the error ends up happening.

This suggested fix is possible because of this change https://github.com/facebook/draft-js/pull/2035/commits/997edf7d1adb7bb0b78d9a527add68bf0f785af2, which makes sure that the select state is always up to date, meaning we can just save the current state, apply the DOM mutations and re-apply the state, without having to read it from the DOM, as we are doing currently, and which caused the issue.

### Test plan 1
Described on Issue https://github.com/facebook/draft-js/issues/2104

### Test plan 2
Tested on Android to make sure the selection state works as it used to.


fabiomcosta:
@sarmstrong do you mind trying out this fix to confirm it fixes the issue?
Ir worked for me, thank you again for the report!

sarmstrong:
Thanks @fabiomcosta for working on a fix and addressing it so quickly.

The JavaScript error is fixed but the behavior is still not entirely correct. 

With this change the Superman entity is removed, however it should stay intact. Unless the content of the Superman entity is altered, it should always remain. 

Current behavior: 
c0af667d5f3611f8b4848d89d20fb543

Previous behavior:
d752df6d747f1b7a8970726f720823f1

sarmstrong:
One problem I noticed is that `getEntityKeyForSelection` doesn't work with `IMMUTABLE` entities as `filterKey` will return null for anything that is not `MUTABLE`

PR #2107, opened 2019-06-14
fixed a typo in `docs/Overview.md`

Haobo-Zhao:
**Summary**

This PR fixed a typo in file `docs/Overview.md`, where better options for word `use` in the changed place are  `we` and `users`, and `we` sounds better to me and thus the PR.

**Test Plan**

No test needed.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -66,7 +66,7 @@ ReactDOM.render(
 );
 ```
 
-Since the release of React 16.8, use can use [Hooks](https://reactjs.org/docs/hooks-intro.html) as a way to work with `EditorState` without using a class.
+Since the release of React 16.8, we can use [Hooks](https://reactjs.org/docs/hooks-intro.html) as a way to work with `EditorState` without using a class.
 
 ```js
 import React from 'react';


docs/Overview.md
```diff
@@ -66,7 +66,7 @@ ReactDOM.render(
 );
 ```
 
-Since the release of React 16.8, use can use [Hooks](https://reactjs.org/docs/hooks-intro.html) as a way to work with `EditorState` without using a class.
+Since the release of React 16.8, we can use [Hooks](https://reactjs.org/docs/hooks-intro.html) as a way to work with `EditorState` without using a class.
```
robbertbrak: How about "you" instead of "we"?

Haobo-Zhao: thank you for the response Robbert, `you` is a good option too, you can definitely open a PR and see who gets picked 😉 , to me `we` sounds better personally.

robbertbrak: Looks like this was already addressed in #2089.

claudiopro: Hi @Haobo-Zhao @robbertbrak, thanks for proposing a fix for this typo! I prioritized the earlier PR as I think it's more fair to accept a fix from a contributor who spotted and addressed it first.

claudiopro:
Already addressed with #2089, but thanks for proposing a fix @Haobo-Zhao!

PR #2113, opened 2019-06-21
Update Overview.md

danielo515:
**Summary**

There is no need to create an extra function just for updating the state on the hooks example

**Test Plan**

Just a doc update, no need to test


danielo515:
Just signed, not sure if I have to do something else

claudiopro:
Manually merged to sync www repo.

Issue #2123, opened 2019-07-08
DOM Selections cleared when <select>ing outside editor


holloway:
**Do you want to request a *feature* or report a *bug*?**

Feature.

I'm happy to open a PR with this change however I first wanted to see if there was interest in this from the Draft team.

**What is the current behavior?**

DOM Selections are cleared on blur.

**What is the expected behavior?**

DOM Selections are maintained.

I hoped to find a config option to disable this behaviour but there was no none.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**

All versions.

-------------------------------

In the Draft source there's this function and code comment,
```javascript
function editOnBlur(editor, e) {
  // In a contentEditable element, when you select a range and then click
  // another active element, this does trigger a `blur` event but will not
  // remove the DOM selection from the contenteditable.
  // This is consistent across all browsers, but we prefer that the editor
  // behave like a textarea, where a `blur` event clears the DOM selection.
  // We therefore force the issue to be certain, checking whether the active
  // element is `body` to force it when blurring occurs within the window (as
  // opposed to clicking to another tab or window).
  if (getActiveElement() === document.body) {
    var _selection = global.getSelection();
    var editorNode = editor.editor;
    if (_selection.rangeCount === 1 && containsNode(editorNode, _selection.anchorNode) && containsNode(editorNode, _selection.focusNode)) {
      _selection.removeAllRanges();
    }
  }
```

In our web app we have a toolbar outside the `<Editor>` for formatting, and this toolbar has both `<button>`s and `<select>`s.

The `<select>`s are the problem in Draft.js because clicking them visibly deselects any text. Have a look at https://jsfiddle.net/y4Lszjv5/ and type some text, select it (while making sure that the selection would be visible with the native `<select>` expanded) and the selection will disappear while the `<select>` is expanded. This makes it harder to see which text would be formatted.

Note that Quill and most other editors don't have this problem as they leave the consistent browser behaviour as-is (as your code comment notes _"This is consistent across all browsers"_).

So I guess there are two potential solutions,

1. Offer a config prop to disable `_selection.removeAllRanges();`, or
2. Handle `<select>` specifically and don't run `_selection.removeAllRanges();` when focus has moved to that type of element.

mrkev:
[...]

PR #2128, opened 2019-07-12
Adding 'preserveSelectionOnBlur' prop

holloway:
**Summary**

This adds a `preserveSelectionOnBlur` prop as per #2123 

**Test Plan**

Tests were added for the two possible scenarios. I'm not sure how much to mock in this test because it is just testing a specific branch of code, so I've faked certain details like the rangeCount.

What do you think @mrkev @claudiopro ?

### DIFF
--- a/src/component/base/DraftEditorProps.js
+++ b/src/component/base/DraftEditorProps.js
@@ -180,6 +180,11 @@ export type DraftEditorProps = {
   // is used for both rendering and paste processing.
   blockRenderMap: DraftBlockRenderMap,
 
+  // When the Editor loses focus (blurs) text selections are cleared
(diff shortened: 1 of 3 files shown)


mrkev:
I like this! Thanks for building it out, I'll import it (:

holloway:
@mrkev ah great, thanks!

So what's the process from here? It gets imported to your internal repo and then it'll make its way into the next release?

mrkev:
@holloway yup— PRs get merged from the internal repo so we can keep the code we have inside and outside in sync, and we can test it against our internal codebases before merging. It'll be merged and available on master right away, and then be available with the release of 0.11.1 through NPM.

PR #2153, opened 2019-08-05
Use entitykey instead of decoratorOffsetKey as DecoratorComponent's key

lintuming:
**Summary**

React don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.
If the `<DecoratorComponent/>` is stateful component ,it may receive wrong state value because the decoratorOffsetKey consist of blockKey and tree's index.
gif2

The demo is here:https://codesandbox.io/s/draftjs-pill-entity-e1689

**Test Plan**

[...]


src/component/contents/DraftEditorBlock.react.js
```diff
@@ -200,12 +200,11 @@ class DraftEditorBlock extends React.Component<Props> {
           UnicodeBidi.getDirection(decoratedText),
           this.props.direction,
         );
-
         const commonProps: DraftDecoratorComponentProps = {
           contentState: this.props.contentState,
           decoratedText,
           dir: dir,
-          key: decoratorOffsetKey,
+          key: entityKey || decoratorOffsetKey,
```
fraziermork: I think this assumes that there's a 1:1 mapping between entities and decorations. We had the same problem and  ended up switching to `decoratorKey` instead in our fork (which has diverged quite a bit jfyi).

lintuming: Thanks for the review,it seems that using `decoratorKey` is more suitable and reasonable.
Should i add a commit that using `decoratorKey` or close this PR?

PR #2209, opened 2019-10-11
fix lineHeight on rtl languages

themzed:
Hi, line-height for RTL languages wasn't correct on website and this PR fixes it.


mrkev:
Hey @themzed, thanks for the contribution! `website/static/css/Draft.css` actually gets generated from files inside `src/`. Could you make your change to one of the files there (and then build Draft to generate the new css)? If I merge this it could get replaced when Draft is built again.

`yarn run build` will build everything. You can learn more about how css is built from the `gulpfile.js` in the root directory.

mrkev:
Doing some triaging so I'll close this, but feel free to comment and re-open if you do get to this. Would love to see this fix happen!

PR #2220, opened 2019-10-18
Fix playground example import

### LINKED ISSUE #2219: Playground Example is broken
Bug

Playground example doesn't compile

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/gmertk/e61z7nfa/.**
Just run it

It should work.

The error seems to be coming from `App.js` because the import refers to a non-existent file.

```ts
import convertFromHTMLModern from 'draft-js/lib/convertFromHTMLToContentBlocks2';
```

it should be

```ts
import convertFromHTMLModern from 'draft-js/lib/convertFromHTMLToContentBlocks';
```

https://github.com/facebook/draft-js/commit/42a9d8b04315cbe7bd7a2fa01e29387c6fae09e1

drdgvhbh:
Fixes  #2219

### DIFF
--- a/examples/draft-0-10-0/playground/src/App.js
+++ b/examples/draft-0-10-0/playground/src/App.js
@@ -27,7 +27,7 @@ import JSONTree from 'react-json-tree';
 import {convertToHTML} from 'draft-convert';
 import PanelGroup from 'react-panelgroup';
 import gkx from 'draft-js/lib/gkx';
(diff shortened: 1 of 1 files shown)


claudiopro:
Good catch @drdgvhbh, thanks for taking the time to contribute a fix!

PR #2230, opened 2019-10-29
Pass correct click listener to editor wrapper

necinc:
**Summary**

As I understand the `onClick` listener on the wrapper `div` element is suppose to replicate native `textarea/input` behavior. But the actual value of `onClick` property was `undefined` instead of a real method `domEditor`s method.

**Test Plan**

It's an example file, no test plan is required.


mrkev:
lgtm

Issue #2226, opened 2019-10-24
convertFromHTML returns 0 contentBlocks for an empty string or an empty tag


mstoltenburg:
**Do you want to request a *feature* or report a *bug*?**
Report a 🐛 

**What is the current behavior?**
`convertFromHTML` returns zero contentBlocks for an empty string or an empty tag like `<p></p>` or `<em></em>`. When I use that result to create a ContentState with `ContentState.createFromBlockArray()`, the ContentState can't be used in e.g.`EditorState.createWithContent()` or `ContentState.hasText()`.

**Please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. convert empty tag like e.g. `convertFromHTML('<p></p>')`
2. use the result in `ContentState.createFromBlockArray()`
3. Try to use that content state in e.g. `EditorState.createWithContent(content)` or `content.hasText()`

```javascript
import { ContentState, convertFromHTML } from 'draft-js';

const { contentBlocks, entityMap } = convertFromHTML('<p></p>');
const content = ContentState.createFromBlockArray(contentBlocks, entityMap);
const hasText = content.hasText();

// throws TypeError: blockMap.first(...) is undefined
```
https://jsfiddle.net/xygf095s/

**What is the expected behavior?**
Return a usable ContentState for empty strings or empty tags converted with convertFromHTML or maybe that the blockMap size is equal for ContentState.createFromText and ContentState.createFromBlockArray if I converted an empty string or tag with `convertFromHTML`.
```javascript
const emptyString = '':
const { contentBlocks, entityMap } = convertFromHTML(emptyString);
ContentState.createFromText(emptyString).blockMap.size === 1
ContentState.createFromBlockArray(contentBlocks, entityMap).blockMap.size === 0
```

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
draft-js 0.11.1
Chrome 78

davoscript:
Hi @mstoltenburg, I was having the same issue, I sent this pull request (https://github.com/facebook/draft-js/pull/2240) yesterday which fixes the problem in case you want to take a look at it.

mstoltenburg:
Very nice. My fix at the moment is quite similar to yours 😃 
```js
export function createEditorStateWithContent(html) {
    const content = createContentStateFromHtml(html);

    return content.getBlockMap().size
        ? EditorState.createWithContent(content)
        : EditorState.createEmpty();
}

[...]

Issue #61, opened 2016-02-24
Context-oriented behaviour


DenisIzmaylov:
Hello and thank you for your yet another awesome solution.

Just watched the demo on [project page](https://facebook.github.io/draft-js/) and noticed it doesn't have context-oriented behaviour. What does it mean?

+ When I switch to H1-H5 mode I'd like to turn off this mode by Enter (new line) key.
+ When I switch to list mode I'd like to create a new level by Tab key.

It seems the good example for improving UX is focusing on Apple Pages to get good ideas and inherit user expectations.

tasti:
You're absolutely right. The context-orientation behavior is really useful, which hasn't been implement in the `rich` demo.

We actually do this in Facebook Notes. It seems like this will be a really common question and could be  provided through the `RichTextEditorUtil`.

hellendag:
For resetting block type, you can use https://github.com/facebook/draft-js/blob/master/src/model/modifier/RichTextEditorUtil.js#L406 within a `handleReturn` prop function.

We could create a `RichUtils` function to bake this in, though my thinking was that the expected rules for return and backspace behavior might vary for different use cases. Apple Pages might be a good UX model, but people have a broad range of expectations from their editors -- it can be hard to prescribe one for all.

For nested lists, see http://facebook.github.io/draft-js/docs/advanced-topics-nested-lists.html#content.

hellendag:
Going to close this because I think we can allow clients to define this behavior themselves. If it comes up more and it seems there is an obvious default, then we can reassess. :)

PR #2240, opened 2019-10-30
Defaulting to createEmpty if block map is empty in createWithContent (Fixes issue #2226)

davoscript:
**Summary**

When passing `createWithContent` a `state` parameter created using `convertFromHTML` from an empty string and `createFromBlockArray`, as shown in the example:

```
const sampleMarkup =
  '<b>Bold text</b>, <i>Italic text</i><br/ ><br />' +
  '<a href="http://www.facebook.com">Example link</a><br /><br/ >' +
  '<img src="image.png" height="112" width="200" />';

const blocksFromHTML = convertFromHTML(sampleMarkup);
  blocksFromHTML.contentBlocks,
  blocksFromHTML.entityMap,
);

this.state = {
  editorState: EditorState.createWithContent(
    state,
    decorator,
  ),
};
```

The following error is thrown:

```
TypeError: Cannot read property 'getKey' of undefined

  84 | EditorState.createWithContent = function createWithContent(contentState, decorator) {
> 85 |   var firstKey = contentState.getBlockMap().first().getKey();
     | ^  86 |   return EditorState.create({
  87 |     currentContent: contentState,
  88 |     undoStack: Stack(),
```
The previous error is generated because `createWithContent` asumes that `contentState.getBlockMap().first()` will return an element, which is wrong for scenarios where the block map is empty. As a consecuence the `getKey();` function in `contentState.getBlockMap().first().getKey();` is executed on `undefined`, thus throwing the error.

The previous scenario is very common if you are planning to use the editor to allow a user to write HTML content which could also be blank, especially as a default value.

**Solution**

In order to make the less amount of modifications to the code, I've added a validation that will run `createEmpty` in case the block map is empty.

**Test Plan**

Open `/draft-js/examples/draft-0-10-0/convertFromHTML/convert.html` and change line #61 from:

`const blocksFromHTML = convertFromHTML(sampleMarkup);`

to:

`const blocksFromHTML = convertFromHTML('');`

No error will be shown.


mrkev:
Sweet, thanks for the fix and the through explanation @davoscript! I'll try to merge it in the coming days.

davoscript:
That's great @mrkev, I see the Linter testing warning here, is there anything I can do to fix it? I'm happy to review it, but I can't see the details.

mrkev:
Ah yeah @davoscript, we expect all `if`'s to have curly braces (`if (...) return ...` vs `if (...) { return ... }`). Any chance you could fix this real quick?

mrkev:
We should sync the lint rules on the ESLint config in this repo with the internal one to avoid these issues to be honest. For future reference we need to add:

```
Lint code: ESLINT
Lint name: curly
```

davoscript:
Thanks a lot for your feedback @mrkev, I've just pushed the commit and see all tests passed this time ;)

PR #2267, opened 2019-12-01
Improve docs syntax and formatting

yangshun:
**Summary**

Reformat the docs in preparation for Docusaurus 2:

- Add syntax highlighting for the code blocks
- Remove `.html` from the Markdown link extensions. Docusaurus 1 allows clean URL but Docusaurus 2 doesn't, so this is necessary
- Wrap API headings in code blocks
- Change the level of some API headings so that they appear in the right table of contents

**Test Plan**

Run Docusaurus website locally and click through.

Try it out - https://draft-js-pc0wv8qnv.now.sh

### DIFF
--- a/docs/APIReference-CharacterMetadata.md
+++ b/docs/APIReference-CharacterMetadata.md
@@ -21,7 +21,7 @@ operations are already implemented and available via utility modules. The getter
 methods, however, may come in handy at render time.
 
 See the API reference on
-[ContentBlock](/docs/api-reference-content-block.html#representing-styles-and-entities)
+[ContentBlock](/docs/api-reference-content-block#representing-styles-and-entities)
 for information on how `CharacterMetadata` is used within `ContentBlock`.
 
 ## Overview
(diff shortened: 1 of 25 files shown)


docs/APIReference-CharacterMetadata.md
```diff
@@ -21,7 +21,7 @@ operations are already implemented and available via utility modules. The getter
 methods, however, may come in handy at render time.
 
 See the API reference on
-[ContentBlock](/docs/api-reference-content-block.html#representing-styles-and-entities)
```
claudiopro: Nice catch!

docs/APIReference-ContentState.md
```diff
@@ -170,36 +170,39 @@ objects.
 
 ## Static Methods
 
-### createFromText
```
claudiopro: Unlike previous headings e.g. `getDepth()`, this and later method names are missing parens `()`. While you're at it, do you mind adding them here and in the other occurrences? Thanks!

yangshun: I wasn't sure whether the omission of the parens was intentional so I left them out. Have added them to all the functions/methods in the API section!

PR #1571, opened 2017-12-12
fixed a bug related to draft losing its selection

pferdone:
**Summary**

Sometimes rangeCount <= 0 and Draft tries to access an index that's out of range. See [stackoverflow](https://stackoverflow.com/questions/22935320/uncaught-indexsizeerror-failed-to-execute-getrangeat-on-selection-0-is-not) for a simplified example. Chrome throws the following exception:

    react-dom.development.js:10994 Uncaught DOMException: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index.
        at addFocusToSelection (http://localhost:3002/vendor.bundle.js:47834:27)
        at setDraftEditorSelection (http://localhost:3002/vendor.bundle.js:47728:5)
        at DraftEditorLeaf._setSelection (http://localhost:3002/vendor.bundle.js:45761:5)
        at DraftEditorLeaf.componentDidUpdate (http://localhost:3002/vendor.bundle.js:45771:10)
        at commitLifeCycles (http://localhost:3002/vendor.bundle.js:34699:24)
        at commitAllLifeCycles (http://localhost:3002/vendor.bundle.js:35867:9)
        at HTMLUnknownElement.callCallback (http://localhost:3002/vendor.bundle.js:26463:14)
        at Object.invokeGuardedCallbackDev (http://localhost:3002/vendor.bundle.js:26502:16)
        at invokeGuardedCallback (http://localhost:3002/vendor.bundle.js:26359:27)
        at commitRoot (http://localhost:3002/vendor.bundle.js:35971:9)

&nbsp;

**Fix**

I added a rangeCount check.

PR #2271, opened 2019-12-03
Added tests and fixed IE IndexSizeError trying to get a range from a selection when there is not one

lrnwytt:
**Summary**
Various browsers will throw errors if `selection.getRangeAt(0)` is called when there is no range. To fix this, a check was added to make sure a range exists.

This change was originally suggested [here](https://github.com/facebook/draft-js/pull/1571) a few years ago. I opted for opening a new PR and adding tests here since that PR seemed stale.

Selection and Range did not work in the test environment, so I had to stub them. I only added functionality for what was needed in the tests I added, but more can be added in the future very easily.

### DIFF
--- a/src/component/contents/DraftEditorLeaf.react.js
+++ b/src/component/contents/DraftEditorLeaf.react.js
@@ -20,7 +20,8 @@ const React = require('React');
 
 const invariant = require('invariant');
 const isHTMLBRElement = require('isHTMLBRElement');
-const setDraftEditorSelection = require('setDraftEditorSelection');
+const setDraftEditorSelection = require('setDraftEditorSelection')
+  .setDraftEditorSelection;
 
 type Props = {
   // The block that contains this leaf.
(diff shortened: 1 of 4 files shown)


mrkev:
This is awesome! Thanks for adding the tests too. I'll try to get this merged.

gdtrice:
Nice!

lrnwytt:
@mrkev Is there anything else I need to do (resolving conflicts and making changes based on whatever the internal test failures/warnings are)? Or is that handed by Facebook now since this PR has been "imported"

mrkev:
Hey @lrnwytt, holiday season has lots of things fighting for my attention, thanks for bringing this up again, I want to get this in 👍

Facebook web exists as a monorepo. Had a quick look at the test failure— it seems that the way this repo is set up to sync with our www monorepo means that adding files on Github doesn't add them to our dependency resolution system, so the tests can't actually find `RangeApi` and `SelectionApi` when we run them internally.

Let's just embed the `RangeApi` and `SelectionApi` classes in the file that uses them. It will make it a little more verbose but it will save a lot of hoop-jumping, and honestly if they're just used here I think it's the most worth approach.

The lint failure is actually a warning because the test is not `@flow strict-local` and the requires are not sorted alphabetically— this won't block landing and I'm fine landing it as is.

mrkev:
Sweet thanks!

lrnwytt:
Thanks for working through this with me! We just got this logged as a bug and it's so great to see how easy the outsider PR process is.
Have a great holiday!

mrkev:
@lrnwytt thanks, you too! (:

PR #2275, opened 2019-12-04
Prettify docs files

yangshun:
**Summary**

Prettier can also be used to format more than just JS files, it can format Markdown files too. Instead of using Flow parser for all files in `prettier.config.js`, we can just limit Flow parser for JS files and let Prettier find the default parsers for the various file formats.

Hence I tweaked the Prettier config and added overrides for:

- JS format - Use Flow parser
- Markdown format - Restrict to 80 characters so that it's easier to read

I then ran Prettier on all the docs. I've added a `prettier:diff` command into the `lint-docs` step, ran alongside with Alex that will fail when there are unformatted docs. It's not part of the CI step as I think that might be a bit too strict given that some people (including myself) edit docs in the GitHub UI and that might cause formatting issues which can be fixed by running the `format-docs` command every now and then.

The formatting is only restricted to the docs now and not code as it's potentially dangerous to format code that gets synced with internal FB repos. 

**Test Plan**

Load website.


mrkev:
Oh this is cool. Ideally yeah, let's enforce it in `lint-docs`. Can prettier automatically apply formatting to markdown files too? Adding that in `format` or a new script would be nice.

prettier.config.js
```diff
@@ -10,5 +10,18 @@ module.exports = {
   trailingComma: 'all',
   bracketSpacing: false,
   jsxBracketSameLine: true,
-  parser: 'flow',
+  overrides: [
+    {
+      files: '*.js',
+      options: {
+        parser: 'flow',
```
yangshun: Use Flow only for JS so that other filetypes can be formatted by Prettier using their respective parsers.

package.json
```diff
@@ -26,8 +26,10 @@
     "dev": "gulp dev",
     "postbuild": "node node_modules/fbjs-scripts/node/check-lib-requires.js lib",
     "lint": "eslint .",
-    "lint-docs": "alex .",
+    "lint-docs": "alex . && yarn format-docs:diff",
```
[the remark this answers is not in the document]
claudiopro: Yep that's legit, the `&&` enforces a clean Alex pass before Prettier

website/README.md
```diff
@@ -16,6 +16,14 @@ $ yarn start
 
 This command starts a local development server and open up a browser window. Most changes are reflected live without having to restart the server.
 
+### Formatting Docs
+
+If you have changed any documentation, before submitting your PR, please run the following command from the root of the directory to use Prettier to format the documentation.
+
+```
+$ yarn format-docs
```
claudiopro: 🎉

prettier.config.js
claudiopro: Are there other files that should be formatted by flow? e.g. `.json` or inlined into `.html` files?

PR #2279, opened 2019-12-07
Remove Jest auto mocking

yangshun:
**Summary**

Saw that the Travis CI is failing because of a new test (`isHTMLBRElement`) which did not explicitly disable auto mocking. In Sandcastle, Jest tests aren't automocked hence the tests passed internally but is failing on GitHub.

Since all the tests in the repo have `jest.disableAutomock()`, there's no value in having `automock: true` within `package.json`, we can remove it and save ourselves of this config discrepancy and further pain in future.

If we decide to go with this approach then #2278 can be abandoned/closed.

**Test Plan**

Travis CI should pass. Also check that Sandcastle passes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "npm": "2.x || 3.x || 5.x || 6.x"
   },
   "jest": {
-    "automock": true,
     "globals": {
       "__DEV__": true
     },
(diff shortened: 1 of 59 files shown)


yangshun:
The internal linting which is failing is `flow strict` warnings which can't be added now (too many things to change.

PR #2287, opened 2019-12-18
Re-write `DraftEditorExample` in the website using react hooks

mrkev:
Also took the oportunity to clean up the component- it had unused imports and `_handleKeyCommand` was returning a boolean instead of `"handled"` and `"not-handled"`.

`yarn && yarn start`, used the editor.

### DIFF
--- a/website/src/components/DraftEditorExample/index.js
+++ b/website/src/components/DraftEditorExample/index.js
@@ -6,111 +6,102 @@
  */
 
 import React from 'react';
-import Link from '@docusaurus/Link';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
-import useBaseUrl from '@docusaurus/useBaseUrl';
 import {Editor, EditorState, RichUtils, getDefaultKeyBinding} from 'draft-js';
-import classnames from 'classnames';
-
-import Layout from '@theme/Layout';
 
 import './css/example.css';
 import './css/draft.css';
 import './css/rich-editor.css';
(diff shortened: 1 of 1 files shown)


claudiopro:
Good job! What about integrating the repo with something like Zeit to deploy the PR for easier testing?

Hey @yangshun how did you do that for your Docusaurus 2 PR, did you deploy manually to now.sh?

website/src/components/DraftEditorExample/index.js
```diff
@@ -6,111 +6,102 @@
...
-    return getDefaultKeyBinding(e);
-  }
+function RichEditorExample(props) {
+  const [editorState, setEditorState] = useState(EditorState.createEmpty());
+  const editor = useRef(null);
 
-  _toggleBlockType(blockType) {
-    this.onChange(RichUtils.toggleBlockType(this.state.editorState, blockType));
-  }
+  const focus = () => {
```
claudiopro: Nit: can you rename to `focusEditor`? This is shadowing the `window.focus()` method

yangshun:
> Hey @yangshun how did you do that for your Docusaurus 2 PR, did you deploy manually to now.sh?

We use Netlify previews. However, Netlify is recently tighten its free tier limits so it might not be that viable as an option. You could check out Netlify or Now.

claudiopro:
@mrkev I don't want to block on this, let's land as-is and fix forward

PR #2288, opened 2019-12-20
Add rules to .flowconfig for flow strict rules

mrkev:
Internally at facebook, we use `@flow strict` to define certain rules that make our codebase more reliable. I just noticed the .flowconfig for this repo doesn't include rules for `@flow strict`. We do have files in the repo annotated with strict flow, so lets configure it to get the benefits.

Unfortunatley, there's 3 rules we have enabled internally which we can't use in the repo as it exists today. I put the necessary info as a comment on the .flowconfig itself.

npm run flow -> no errors.


claudiopro:
This is great! It will eliminate a lot of inconsistencies between OSS repo and www 👍 This will also cover us with `strict-local` files https://flow.org/en/docs/strict/#toc-strict-local

PR #2294, opened 2020-01-02
Add catch argument to fix build

mrkev:
## Summary

Turns out optional argument for catch statements are not universally supported— some versions of node and some browsers don't allow them. [Whoops](https://github.com/facebook/draft-js/commit/64b51dff37dbc385532a56ed19e23a9da1437385).

## Test Plan

`npm run build`


mrkev:
Oh, nvm. Fix already in https://github.com/facebook/draft-js/pull/2291

PR #2297, opened 2020-01-03
Making sure we don't throw when selecting text on multiple fields

fabiomcosta:
**Summary**

When selecting text between multiple draft-js Editor fields rendered on a page, an unexpected error happens.
This comment on the code explains why the bug happens: https://github.com/facebook/draft-js/pull/2297/files#diff-92f7a619f202063fc7fd29fcfa73c3e4R46-R50

**Test Plan**

I created this sandbox to more easily show the issue this PR is solving.
Follow the instructions on the sandbox to repro the issue and see the error: https://codesandbox.io/s/affectionate-rgb-64zsc

Here is what you should see:
draftjs_select_bug


fabiomcosta:
This is the same fix merged at https://github.com/facebook/draft-js/commit/a9fcbb201de604e040dacb66e67407989a0e2d9b

Closing

PR #2300, opened 2020-01-07
chore(travis): bumps node versions used in Travis CI to 10 and 12

claudiopro:
**Summary**

As of January 2020, Node 8 is past Active LTS end of life. This PR removes Node 8 from the Travis CI environment config and replaces it with Node 12.

https://nodejs.org/en/about/releases/

This PR also updates supported Node versions in `package.json` to 10, 12 and 13.

**Test Plan**

Tested locally building all targets on Node 12 with [nvm](http://nvm.sh/)
```
nvm use 12
yarn
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 
 node_js:
-- 8
 - 10
+- 12
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


fabiomcosta:
nice

Issue #2304, opened 2020-01-14
Untracked website/static/css/ when running yarn


lukyth:
**Do you want to request a *feature* or report a *bug*?**
Probably *feature* (improvement)

**What is the current behavior?**
Run `yarn` on the current master will produce this untracked file `website/static/css/`.

**What is the expected behavior?**
`website/static/css/` should be either ignored or commited.

**Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?**
v0.11.4 (https://github.com/facebook/draft-js/tree/ad8374d278507ae73dceffd9bbbcede398f849e3)

claudiopro:
Good catch @lukyth ! This has been nagging me for a few days. It's a remainder from the migration to Docusaurus 2 and should be deleted. Would you be willing to contribute a PR? That file is generated by a gulp task:

https://github.com/facebook/draft-js/blob/master/gulpfile.js#L258-L262

lukyth:
Sure I'd love to try! :D

PR #2308, opened 2020-01-18
Remove unused gulp task remained from docusaurus 2 migration

lukyth:
**Summary**

Run yarn on the current master will produce an untracked file `website/static/css/Draft.css`. It's a remainder from the migration to Docusaurus 2 and should be deleted.

This will fix https://github.com/facebook/draft-js/issues/2304

**Test Plan**

I ran `cd website && yarn && yarn start` and the doc website works fine locally.


claudiopro:
Thank you for contributing this fix @lukyth! 🎉

PR #2310, opened 2020-01-22
Update webpack deps

mrkev:
**Summary**

This diff update build-related deps. We were creating more than one instance of `webpack-stream`, which is an error on the latest one, so I fix that too.

**Test Plan**

build && playground example

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -71,7 +71,12 @@ var COPYRIGHT_HEADER = `/**
  */
 `;
 
+var wpack = null;
+
 var buildDist = function(opts) {
+  if (wpack !== null) {
+    return wpack;
+  }
   var webpackOpts = {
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -71,7 +71,12 @@ var COPYRIGHT_HEADER = `/**
  */
 `;
 
+var wpack = null;
```
claudiopro: Nit: can we rename this `wpStream`? I think it would be easier to reason about

claudiopro:
Thanks for taking a look at this issue @mrkev ! Please take a look at my comment inline, not blocking on it though.

mrkev:
Sound good! I'll apply the changes internally through Phabricator and land it right away to keep things going.

PR #2320, opened 2020-02-05
Remove usage of String.prototype.endsWith API as it's unavailable in …

Wobbly-Wibbly:
Remove usage of String.prototype.endsWith API as it's unavailable in **IE 11** :sob:
https://caniuse.com/#search=endsWith

Mostly positive, that you won't merge that ))
As you mentioned before that you don't support old browsers, but there is always hope ))
I had to try ))

Thanks for your library and have a nice day 😄


robbertbrak:
You are right that this will most likely not be merged... 🙂 

We are also running DraftJS in an application that still supports IE11. A common solution is to add polyfills for the functions that are lacking in IE11, such as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Polyfill.

PR #2321, opened 2020-02-05
Remove usage of endsWith API as it's unavailable in IE 11

Wobbly-Wibbly:
Remove usage of String.prototype.endsWith API as it's unavailable in IE 11 😭
https://caniuse.com/#search=endsWith

Mostly positive, that you won't merge that ))
As you mentioned before that you don't support old browsers, but there is always hope ))
I had to try ))

Thanks for your library and have a nice day 😄


Wobbly-Wibbly:
sry for duplicate pr, was confused with the master restriction on forked repo

PR #2356, opened 2020-03-19
Don't SSR example editor on website.

mrkev:
There's suspicion the example editor fails in prod because of an SSR error, so lets not SSR-it. Also upgrades draft to the latest version on npm.

## Test Plan

`yarn start`, played around with the editor.


yangshun:
Good one, thanks!

PR #2369, opened 2020-03-26
Fix flow by allowing optional chaining

mrkev:
**Summary**

Turns out the optional chaining PR (#2330) also broke flow. We should run a flow check automatically before we build, not just strip the types.

**Test Plan**

`flow .`


fabiomcosta:
+1

PR #2387, opened 2020-04-17
Add keepPastedSoftLineBreaks option

luism-s:
**Summary**

I was struggling with a problem that whatever text with line breaks I would paste would produce multiple blocks.

The same way that we are able to strip every style and also add soft line breaks instead of new blocks on return with
```
    function handleReturn(e, editorState) {
      const newEditorState = RichUtils.insertSoftNewline(editorState));
      ...
      
      return 'handled';
    }
```

Maybe there are more use cases (like mine) where the developer wants to interpret pasted line breaks as 'soft line breaks'.

For this I added a new property named `keepPastedSoftLineBreaks` (default: `false`), which if set to `true` the text will not be splitted.

I was trying to make it work alongside `stripPastedStyles=false`, but it seems that the HTML parser takes multiple divs as hard line breaks and I couldn't figure out how to stop it.

### DIFF
--- a/docs/APIReference-Editor.md
+++ b/docs/APIReference-Editor.md
@@ -176,6 +176,22 @@ This should be used if your editor does not support rich styles.
 
 Default is `false`.
 
+### `keepPastedSoftLineBreaks`
+
+```js
+keepPastedSoftLineBreaks?: boolean
+```
+
+Set wether to interpret line breaks as distinct blocks or as soft line breaks.
+
+It means that if set to `true`, line breaks will not produce multiple blocks,
+but one with all the text. It will still habve line breaks, as if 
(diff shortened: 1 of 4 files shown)


mrkev:
Hey @luism-s, thanks for writing this! Can you rebase it so that the CI passes? I think it was broken on master when you did the PR, an hence it's failing now.

luism-s:
Hi @mrkev, I'm not sure of the name either.

I wanted something a long the lines: interpret pasted line breaks as soft line breaks. Do you have any suggestion? I thought about stripPastedLineBreaks, but we are not actually removing them..

About stripPastedStyles=false I guess it would have to go deeper into the level of html parsing, as div as p elements would need to be treated as both soft new line

PR #2429, opened 2020-05-14
[Docs]: Add Null return value type in code example

jacobdcastro:
**Summary**

TypeScript threw this error when I attempted to copy and paste in my code:
```
Type 'null' is not assignable to type 'string'.ts(2322)
```

The `getDefaultKeyBinding()` method may return null. So I added `| null` to the return type declaration. Just thought it would help others who may also be copying/pasting the code!

**Test Plan**

No testing required.


mrkev:
Yup, can confirm the type is an optional `DraftEditorCommand` 👍

https://github.com/facebook/draft-js/blob/bc716b279299748d955bbdc398454bff45ea0191/src/component/utils/getDefaultKeyBinding.js#L63

PR #2434, opened 2020-05-19
Update Overview.md

abdellani:
**Summary**

Import `Draft.css` in the getting started example.


claudiopro:
Thank you for this, it is clearly not obvious that the editor needs its `Draft.css` stylesheet to be imported for base styles.

While the intent of an `import` with a CSS file is quite clear, it may not work for everyone. What about calling out that the user should import styles in `Draft.css` either directly or using their style framework of choice?

PR #2452, opened 2020-06-02
Update media.html

love-fay:
I'm just adding two lines of comments, because I made this hard to see difference mistake, and I'm worried that others are the same as me


mrkev:
@love-fay this looks useful, but I think you need to sign the CLA before I can import this.

PR #2497, opened 2020-07-16
Fixes links accessibility by adding underlines

### LINKED ISSUE #2494: Because of BLM banner change, the site became inaccessible
Because of this change, the site became inaccessible

What happened to links
image

It is hard to see links.

_Originally posted by @RAYDENFilipp in https://github.com/facebook/draft-js/pull/2454#issuecomment-658213125_

rachelnabors:
Closes #2494

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -31,3 +31,7 @@
 .container {
   filter: grayscale(100%);
 }
+
+.main-wrapper a:link:not(.button) {
+  text-decoration: underline;
+}


website/src/css/custom.css
```diff
@@ -30,4 +30,8 @@
 
 .container {
   filter: grayscale(100%);
-}
\ No newline at end of file
+}
+
+.main-wrapper a:link:not(.button) {
```
claudiopro, line 35: Excellent, good catch on excluding links styled as `.button`.

RAYDENFilipp:
This doesn't solve the issue of accessibility. Links are just one thing, but there are code-blocks, videos, etc. The whole site causes a headache when reading it.

PR #2523, opened 2020-07-23
feat(docs): Update README.md

iamdarshshah:
**Summary**

#### Changes
- [x] Added logo to the top
- [x] Aligned header elements to the center

> This change will make it look symmetrical and aligned in a proper format.

> Screenshot
Draft JS

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="http://draftjs.org/">
-    <img src="https://draftjs.org/img/draftjs-logo.svg" alt="draftjs-logo" width="15%" />
+    <img src="https://draftjs.org/img/draftjs-logo.svg" alt="draftjs-logo" width="8%" />
   </a>
 </p>
 <h1 align="center">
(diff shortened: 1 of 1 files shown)


mrkev:
Can you share a screenshot?

iamdarshshah:
Sure @mrkev

iamdarshshah:
Sure @mrkev , Summary updated! We do have other options to look for i.e., we can place the gif after the introductory content. I would like to know your thoughts on this.

Thanks 👍

mrkev:
@iamdarshshah I think keeping the gif there is ok— it's a quick way of seeing what draft is and we should open with it.

The logo looks a little big though. I know it's a current design trend to have the big, centered logo, but this always felt weird. I did a little research and I think it's probably because other Facebook JS projects don't even show their logo on their readme: [react-native](https://github.com/facebook/react-native), [react](https://github.com/facebook/react), [relay](https://github.com/facebook/relay).

It's a cool logo though, so I'm down to try it out. Can we try putting it inline before the title (ie, "[logo] Draft.js")? That'd be my first option, though I'm open to see how it looks on top but maybe like around 50% smaller (:

iamdarshshah:
> I think keeping the gif there is ok— it's a quick way of seeing what draft is and we should open with it.

Okay, agreed! @mrkev 👍 

>  Can we try putting it inline before the title (ie, "[logo] Draft.js")?

Already did that, but the text was not aligning center vertically. Also, that was not looking as we are expecting it to look. Thus, my suggestion for this is to create an SVG that includes `[logo] text` aligned side-by-side and we can update it as docs logo. Would like to know your thoughts on this?

Regarding reducing image size by 50%, then it will look like this:
Screenshot (60)

mrkev:
I like this size much better. Lmk when the commit is updated and I'll import it (:

iamdarshshah:
Thanks @mrkev 👍 , commit is Updated. You can now review it :)

PR #2527, opened 2020-07-24
Update readme to suggest building with yarn

mrkev:
Every now and then there will be issues by people who try to build with NPM. The build works with yarn.


mrkev:
While at it I'm also removing the link to this Slack team, since it's no longer being maintained

PR #2530, opened 2020-07-27
Update link to Immutable Map in docs

agarwalrounak:
**Summary**
Map in docs, wherever mentioned, links to http://facebook.github.io/immutable-js/docs/#/Map and also in some cases to http://facebook.github.io/immutable-js/docs/#/Record/Record which then redirect to https://immutable-js.github.io/immutable-js/

However, the correct link for Map is https://immutable-js.github.io/immutable-js/docs/#/Map

Draft.js uses immutable 3.7.4 which released on 18 Jun 2015 according to this commit https://github.com/immutable-js/immutable-js/commit/09f04e910bd0b891d1373eb2cb2648a0546fab3d
I have linked `Map` to the docs for that version(captured on 23 Jun 2015) through the Internet Archive.

### DIFF
--- a/docs/APIReference-ContentBlock.md
+++ b/docs/APIReference-ContentBlock.md
@@ -118,7 +118,7 @@ _Properties_
 
 > Note
 >
-> Use [Immutable Map API](http://facebook.github.io/immutable-js/docs/#/Map)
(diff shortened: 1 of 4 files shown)


mrkev:
This is awesome, thanks @agarwalrounak!

PR #2554, opened 2020-08-05
Media Editor added in website hompage

nazmulshuvo03:
**Summary**
I have worked on the Draft.js website. There are only one live editor in the website which is able to edit text only. I have added another one for image, audio and video file.

**Test Plan**

Test Suites: 58 passed, 58 total
Tests: 630 passed, 630 total
Snapshots: 765 passed, 765 total
Screenshot from 2020-08-07 00-21-38

Screenshot from 2020-08-07 00-21-56


mrkev:
Hey @nazmulshuvo03, thanks for this PR! Could you post screenshots to see what the change will look like? For changes that affect UI screenshots should be part of the test plan 👍

nazmulshuvo03:
@mrkev added screenshots

mrkev:
Thanks for the screenshots! (: I have some comments on the design so people have a better idea of what they're looking at. Namely, let's:

1. put the second editor under the first, this will let us...
2. put a little description of what the second editor is. I'm thinking something like the following

```
"Here's a simple example of a rich text editor built in Draft.js."

[Rich text editor]

"Here's another simple example of a text editor with embedded media"

[Media editor] 
```

I also have some comments on the code:

1. Do we need to duplicate, `draft.css`, `example.css` and `rich-editor.css`? Why not just use the files already there for the other editor?
2. The page right now has two elements with `id="rich-example"`, which is bad form. Ideally IDs are unique (we can use class-names to apply the same style to several elements).

Anyway, passing this back to you. 👍

PR #2556, opened 2020-08-07
improve text syntax

lebogangolifant:
**Summary**

improve text syntax

### DIFF
--- a/docs/APIReference-APIMigration.md
+++ b/docs/APIReference-APIMigration.md
@@ -1,5 +1,5 @@
 ---
-id: v0-10-api-migration
+id: v0-10-api-migration 
 title: v0.10 API Migration
 ---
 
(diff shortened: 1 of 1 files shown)


mrkev:
Hey @lebogangolifant— when I import the PR the summary is what shows up as the commit message. Could you edit it and make it meaningful to the change? Doesn't have to be long, just not the template.

lebogangolifant:
@mrkev thanks for reviewing my PR. I have made changes according to your suggestion  I hope this will help.

docs/APIReference-APIMigration.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: v0-10-api-migration
+id: v0.10 API Migration
```
mrkev, line 2: Hmm, if we change the ID here won't we need to change all the places we reference it? Does this change the URL too? Maybe it's a good idea to leave it as is.

lebogangolifant:
@mrkev  I referred to Docusaurus API specification docs on how to write down markdown header features, and you are correct. So I changed back the id to its original state. Thanks for highlighting this, learned something new.
API migration docs

PR #2571, opened 2020-08-23
fix: parentNode should be cleared in `removeRangeFromContentState.updateBlockMapLinks` function

ryuever:
**Summary**

In experimental mode. There is a blockMap

1. `A` has a child `C`
2. `C` has a child `D`
3. `D` is empty

if backspace block `D`, `blockMap` will be updated as follows:

1. set `A` as `D`'s parent
2. update `A`'s children list
3. remove `C`

But, `C` is still in `A`'s children list due to the use of old `newParent` which is derived from `blockMap`


ryuever:
@mrkev May I ask for a look on this issue. 

I am working on project https://github.com/ryuever/react-tapable-editor, and this issue really be a trouble..

PR #2608, opened 2020-09-03
created examples that using hooks

NtsCiccio:
## Summary
Hello everyone,
I created some examples that use hooks, 
I did it because I looked at the examples when I needed them and I couldn't find them
some examples that used functional components and hooks.

### DIFF
--- a/examples/draft-0-10-0/color/colorHook.html
+++ b/examples/draft-0-10-0/color/colorHook.html
@@ -1,0 +1,208 @@
+<!--
+Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+
+This file provided by Facebook is for non-commercial testing and evaluation
+purposes only. Facebook reserves all rights not expressly granted.
+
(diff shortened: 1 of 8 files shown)


mrkev:
Hey @NtsCiccio, this is awesome thanks! Two things: 

- Could we rename the examples? Instead of being `linkHook.html`, `entityHook.html`, etc. Can we call them `link_usingHooks.html`, `entity_usingHooks.html`, etc?

- Could you include a test plan with your PR? Something describing how you made sure they worked. I'm assuming you ran the examples and they worked, so a just saying that or posing a screenshot of the example running should be enough.

NtsCiccio:
Hey @mrkev , i've made the changes, let me know if i missed something.

iggyZiggy:
@NtsCiccio thank you so very much for this, can't believe i had to dig this deep to find some useful examples

NtsCiccio:
> @NtsCiccio thank you so very much for this, can't believe i had to dig this deep to find some useful examples

Hey iggy, glad to know it was helpful

NtsCiccio:
HI @mrkev , are there any news?

Issue #1696, opened 2018-03-15
ariaRequired prop should render to aria-required attribute


shimedh:
**BUG REPORT (AND POTENTIALLY A FEATURE REQUEST)**

There should be an ariaRequired prop for Editor control and it should render the provided id string to the aria-required attribute.

As of Draft 0.10.5 I don't see any attribute for ariaRequired.
https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorProps.js

EJLearner:
Also, aria-invalid is an important indicator that an input is in error. Will support for that be added?

waynegerard:
Commented on the PR linked, but - is this still necessary? My understanding is that the required attribute as part of HTML5 supersedes the aria attribute, and support seems pretty widespread for that.

PR #2675, opened 2020-10-22
Add aria-required attr to DraftEditor

sjohnson-sage:
Closes: https://github.com/facebook/draft-js/issues/1696

#### Summary

`aria-required` is used to indicate that user input is required on an element before a form can be submitted. https://www.w3.org/TR/wai-aria-1.1/#aria-required 

It isn't currently possible to pass it in as a prop here so I'm just adding it in alongside the other `aria...` props. 

#### Test Plan

Tested using `npm run test && npm run lint && npm run flow`

### DIFF
--- a/src/component/base/DraftEditor.react.js
+++ b/src/component/base/DraftEditor.react.js
@@ -409,6 +409,7 @@ class DraftEditor extends React.Component<DraftEditorProps, State> {
             aria-labelledby={this.props.ariaLabelledBy}
             aria-multiline={this.props.ariaMultiline}
             aria-owns={readOnly ? null : this.props.ariaOwneeID}
+            aria-required={this.props.ariaRequired}
             autoCapitalize={this.props.autoCapitalize}
             autoComplete={this.props.autoComplete}
             autoCorrect={this.props.autoCorrect}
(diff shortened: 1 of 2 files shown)


sjohnson-sage:
All of the current lint errors existed before my change

waynegerard:
Just curious - is this necessary? My understanding is that the `required` attribute as part of HTML5 supersedes the aria attribute, and [support seems pretty widespread for that.](https://caniuse.com/mdn-api_htmlinputelement_required)

sjohnson-sage:
@waynegerard it is I think, because the `required` attribute is only supported by the input elements listed here: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

This DraftEditor doesn't use any of those elements and therefore there's nothing to put the `required` attribute on

Accessibility56:
@sjohnson-sage One of our application is dependent on "npm Package". Could you please let us know by when this package will be released.

sjohnson-sage:
Hi @mrkev - do you know when the next version of draft-js will be released? Just tagging you as you seem to be an active contributer to the repo - thanks 😄

mrkev:
Hey @sjohnson-sage, I don't. I've actually left Facebook, so I can't do releases myself anymore either and I'm not sure who to ping about this. Sorry /:

Issue #2788, opened 2021-01-08
Custom block will unmount when you type Chinese, but not when you type English


Shenenkaka:
#### Do you want to request a *feature* or report a *bug*?
bug
#### What is the current behavior?
Custom block will unmount when you type Chinese, but not when you type English, Is this normal?
#### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. You can use this jsfiddle to get started: https://jsfiddle.net/gmertk/e61z7nfa/.  
reproduce in this fiddle: https://jsfiddle.net/shenenkaka/5oqgbpc4/18/

#### What is the expected behavior?
Custom block should not unmount when you type language that DraftJs support
#### Which versions of Draft.js, and which browser / OS are affected by this issue? Did this work in previous versions of Draft.js?
Draft.js version: 0.11.7  
browser: Chrome v87,macOS Big Sur  
It is also not work in previous version.

lousander:
I'm having a similar issue when typing brazilian portuguese. We have some letters that use accents e.g., é, ê, ã, etc. Typing any of these triggers an unmount on all our custom components. It can be really troublesome if we have a custom component that does a request or some async code, as it keeps redoing it every time we type any of those letters. I made some tests and this behavior started happening on v0.11.0.
It can also be reproduced in the example @Shenenkaka made.

lousander:
After further investigation, I found the commit responsible. This behavior started happening on this PR: https://github.com/facebook/draft-js/pull/2035
It solved major issues related to Android input and word composition. The issue seems to be directly tied to the alterations made to the `DraftEditorCompositionHandler.js` file. Apparently after any kind of composition, it runs a function called `restoreEditorDOM`, which is described as "a complete re-render of the DraftEditorContents based on the current
   * EditorState". Before the PR, this function was only called when certain conditions were met, now it seems to happen everytime a composition is completed.

lousander:
Hey guys! Some updates. I made a solution and tried to return the composition behavior to what is was before `v0.11.0` version without losing the fixes made for Android. I brought back the `mustReset` flag and with that I expect the re-render of the DraftEditor to happen only on certain conditions. So, a composition on the start of a new text block will still cause the unmount on atomic blocks, this unmount is still necessary for `draft-js` to work properly, as stated in some comments in the code:
`We also re-render when the composition occurs
at the beginning of a leaf, to ensure that if the browser has
created a new text node for the composition, we will discard it.`

[...]

PR #2823, opened 2021-02-09
Reduce calls to restoreEditorDOM during compositions

lousander:
#### Summary
This PR tries to reduce the calls made to the `restoreEditorDom` method during compositions.

I've been experiencing some weirds issues with my custom components since version `v0.11.0`. The issue happens when typing brazilian portuguese, because we have some letters that use accents e.g., é, ê, ã, etc, so in draft-js they start a composition. Typing any of these triggers an unmount on all our custom components. It can be really troublesome if we have a custom component that does a request or some async code, as it keeps redoing it every time we type any of those letters. The same seems to be true when typing Chinese or Japanese characters as well because they also trigger compositions, as stated in #2788.

This PR attempts to mitigate this behavior by bringing back the `mustReset` flag that existed before `v0.11.0`, in the `DraftEditorCompositionHandler.js` file. The flag limits the calls to `restoreEditorDOM` method to happen only when necessary. With the current code I've identified it to be necessary when: 
- A mutation is starting a new text node 
- A composition is being resolved and the `entityKey` for the current selection is not `null`

#### Test Plan
It's possible to test the current behavior on [this code sandbox](https://codesandbox.io/s/draftjs-composition-0117-c47gs), which has the latest versions of draft-js(`0.11.7`) and a custom component with a `console.log` on the unmount component lifecycle. Everytime a composition is resolved, the `restoreEditorDOM` method is called, re-rendering the editor calling the unmount. (Can be seen on the `console` tab).

And here is another [code sandbox with this PR applied to it](https://codesandbox.io/s/draftjs-composition-fix-5rr3j). I has the same custom block in it. The unmount is limited only to specific cases.

### DIFF
--- a/src/component/handlers/composition/DOMObserver.js
+++ b/src/component/handlers/composition/DOMObserver.js
@@ -50,14 +50,15 @@ class DOMObserver {
   constructor(container: HTMLElement) {
     this.container = container;
     this.mutations = Map();
+    this.mustReset = false;
     const containerWindow = getWindowForNode(container);
     const MutationObserver = containerWindow.MutationObserver;
     if (MutationObserver && !USE_CHAR_DATA) {
-      this.observer = new MutationObserver(mutations =>
+      this.observer = new MutationObserver((mutations) =>
         this.registerMutations(mutations),
       );
     } else {
-      this.onCharData = e => {
+      this.onCharData = (e) => {
         invariant(
           e.target instanceof Node,
           'Expected target to be an instance of Node',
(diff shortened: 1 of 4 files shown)


src/component/handlers/composition/DOMObserver.js
```diff
@@ -124,7 +128,11 @@ class DOMObserver {
         return target.textContent;
       }
     } else if (type === 'childList') {
-      if (removedNodes && removedNodes.length) {
+      if (addedNodes && addedNodes.length) {
+        // This mutation is creating a new node, meaning it's happening
+        // at the beginning of a leaf, so we must force a re-render.
+        this.mutationAtLeafStart = true;
```
fcosta-godaddy, line 134: nit, I'd add the `has` prefix to clarify that this is a boolean.
```suggestion
        this.hasMutationAtLeafStart = true;
```

lousander: Hey, @fcosta-godaddy. Thank you for the feedback. Done!

PR #2164, opened 2019-08-21
Add ClassNames support to the customStyleMap

newsiberian:
**Summary**

This PR makes it possible to pass classNames (strings) instead of styles (objects) to the `customStyleMap`

Related with #957, #1368, #1639
and partly related with #342, #521

If you will accept this PR, please, let me know which part of documentation should be updated and how

**Test Plan**

This changes can be tested via `Color` example which was slightly modified

Issue #957, opened 2017-01-19
Can `customStyleMap` specify a CSS class so the style can be managed in a CSS file?


amireldor:
Styles defined in `customStyleMap` are specified in JavaScript objects. I was wondering if there's a way to specify a CSS class name instead of the hardcoded JS-based styles and have my styles managed in a separate CSS file (or LESS/Sass/Stylus/etc...). Is that possible?

tobiasandersen:
Currently, I don't think it is. There are no classes added to [those elements](https://github.com/facebook/draft-js/blob/5120924e4ed1a04f050ee58fa3a02016e83c373c/src/component/contents/DraftEditorLeaf.react.js#L160-L165). It wouldn't be very hard to create something similar for a className prop though, but I have no idea if that's something they (as in the Draft team) want or not.

amireldor:
Hmm, so can be possible for me to fork and add a `className` through `styleObj` or something like that? Is that a sane thought?

tobiasandersen:
That's definitely possible!

amireldor:
Hey @tobiasandersen, so I made a small hack and forked the repository but I'm facing some problems with my fork.

When I `npm install` using the URL to my repo, the repo won't build automatically to create the distribution files. This means I have nothing to import in my project because nothing is found. I tried building manually as the README suggests but it seems that the gulpfile defining the build process is missing, that is odd!

I haven't had to chance to dive into it yet but if you got any idea then I'd be happy to hear.

I'm not sure this issue is in scope for this stuff, so maybe I'm better off stackoverflowing or something similar, please let me know.

tobiasandersen:
No that doesn't work. You have to either commit the build files, or git clone and build manually inside node_modules. 

Join the slack channel, there's usually someone answering pretty quickly!

dictions:
I'm also curious about the status of this issue. 

I want to be able to add an underline style equivalent to [this](https://css-tricks.com/styling-underlines-web/). This is only possible by adding some pseudo elements to a style selection. 

I understand that we use inline styles to help with pasting into other text editors, but are there plans to allow returning a `className` string with either `customStyleMap` or `customStyleFn`?

donomans:
[...]

Issue #2864, opened 2021-03-14
yarn build error


juny1zhang:
I got this error message when I try to execute “yarn ”。

```
Unhandled rejection Error in plugin "webpack-stream"
Message:
    ./lib/DraftEditorTextNode.react.js
Module not found: Error: Can't resolve './react' in '/Users/zhangjunyi/bytedance/demo/draft-js/lib'
resolve './react' in '/Users/zhangjunyi/bytedance/demo/draft-js/lib'
  using description file: /Users/zhangjunyi/bytedance/demo/draft-js/package.json (relative path: ./lib)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: /Users/zhangjunyi/bytedance/demo/draft-js/package.json (relative path: ./lib/react)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react doesn't exist
      .wasm
        Field 'browser' doesn't contain a valid alias configuration
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react.wasm doesn't exist
      .mjs
        Field 'browser' doesn't contain a valid alias configuration
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react.mjs doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react.js doesn't exist
      .json
        Field 'browser' doesn't contain a valid alias configuration
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react.json doesn't exist
      as directory
        /Users/zhangjunyi/bytedance/demo/draft-js/lib/react doesn't exist
[/Users/zhangjunyi/bytedance/demo/draft-js/lib/react]
[/Users/zhangjunyi/bytedance/demo/draft-js/lib/react.wasm]
[/Users/zhangjunyi/bytedance/demo/draft-js/lib/react.mjs]
[/Users/zhangjunyi/bytedance/demo/draft-js/lib/react.js]
[/Users/zhangjunyi/bytedance/demo/draft-js/lib/react.json]
```

scarcoco:
me too

usecodelee:
how to solve this error 😭

mariusbutuc:
[...]

PR #2930, opened 2021-05-05
Allow to style editor nodes with CSS classes

andresgutgon:
# Summary
Until now you could pass an inline style CSS object for `BOLD` or `ITALIC`. With this new prop `customCssClassesMap` you can pass CSS classes. Which allow to use your own styles or utility classes like TailwindCSS

This `red` color is set with a TailwindCSS class
image

``` javascript
const customCssClassMap = {
  BOLD: 'text-red-600'
}
<Editor
  customCssClassMap={customCssClassMap}
  {...otherProps}
/>
```

## Related Issues and PR
https://github.com/facebook/draft-js/pull/2164
https://github.com/facebook/draft-js/issues/957


andresgutgon:
I had to develop as [explained here](https://github.com/facebook/draft-js/issues/2864)

andresgutgon:
The CI is broken due to that error : )

andresgutgon:
Although looking into documentation not sure if we need this PR having this prop:
https://draftjs.org/docs/advanced-topics-block-styling/#blockstylefn

Thoughts?

charuwts:
> Although looking into documentation not sure if we need this PR having this prop: https://draftjs.org/docs/advanced-topics-block-styling/#blockstylefn
> 
> Thoughts?

As far as I can tell that is only for block level styling. It would be a serious improvement to have them inline.

I do see a couple threads saying this won't be implemented: e.g. https://github.com/facebook/draft-js/issues/957 But there are a couple pull requests already so I don't see why it's been rejected...

charuwts:
Does anyone know of a plugin that adds this ability?

andresgutgon:
No longer working on this. Feel free to copy and re-open your own PR from here : )

PR #2956, opened 2021-05-21
Upgrade draft.js’ immutable.js dependency to v4

### LINKED ISSUE #1425: Immutable-js newest version -> draft-js incompatible
### What happened

I updated to 4.0.0-rc.7.
Suddenly i noticed draft-js stopped working. Threw  an unrecoverable error.

### How to reproduce

Install draft-js 10.3, install the version of immutable i mentioned. It will happen naturally.

acusti:
#### Summary

this is the follow-on PR to #2955 that actually upgrades draft.js’ immutable.js dependency to v4 and isn’t compatible with v3. this branch was branched off of `immutable-4-prep` and includes 4 additional commits:

* Upgrade immutable dep from 3.7.4 → 4.0.0-rc.12 803de93
* Update use of `KeyedCollection.toArray()` to be compatible with a breaking change to its behavior (it now returns an array of `[k, v]` tuples, instead of just an array of the values, see immutable-js/immutable-js#1340) 381f5d3
* Suppress unimportant flow error due to inconsistent return type (the null values that are returned and that are inconsistent get filtered out on the next line via `.filter(block => !!block)`) a930d84
* Suppress flow error in ContentState’s `getBlockForKey` 1a5114f

for that last commit, `getBlockForKey` is annotated as always returning a `BlockNodeRecord`, but calling `.get(key)` on the blockMap can return undefined. this means the method is unsafe, but because it is so widely used and the change doesn’t have to do with changes in immutable.js v4, i made the judgement that addressing the issue is out-of-scope of this PR.

with all of these changes and with v4 of immutable.js installed, `yarn test` passes. there are some pre-existing linting errors from the prettier eslint plugin that are also in master. there are also some new flow errors pertaining to `CharacterMetadata`, but i’m not sure what to do about them. there are already 6 flow errors related to those types in `master` that seem to be the result of d0171d5bf35d1962d0bda55256f2785e9849872d, so it’s hard for me to tell if the new errors are also from that change, or if they are actual issues i need to address.

resolves #1425

#### Test Plan**

all automated tests are passing. we are manually testing these changes at my company, and i will report back here with the results of that testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "fbjs": "^2.0.0",
-    "immutable": "~3.7.4",
+    "immutable": "~4.0.0-rc.12",
     "object-assign": "^4.1.1"
   },
   "peerDependencies": {
--- a/src/component/contents/DraftEditorContents-core.react.js
+++ b/src/component/contents/DraftEditorContents-core.react.js
@@ -21,6 +21,7 @@ import type {BidiDirection} from 'UnicodeBidiDirection';
 const DefaultDraftBlockRenderMap = require('DefaultDraftBlockRenderMap');
 const DraftEditorBlock = require('DraftEditorBlock.react');
 const DraftOffsetKey = require('DraftOffsetKey');
+const UnicodeBidiDirection = require('UnicodeBidiDirection');
 
 const cx = require('cx');
 const joinClasses: (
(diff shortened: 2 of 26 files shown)


wehriam:
Hi @flarnie - thanks for your work on Draft.js. Any updates on this issue? 

We are encountering a build issue with Vite.js and it would be helpful if Draft.js was moved to Immutable@4.x.

Thanks!

PR #2957, opened 2021-05-21
Handle case in getCorrectDocumentFromNode where node is a Document

acusti:
#### Summary

when dragging a native file over a draft editor instance that is being rendered within an iframe, the [`currentTarget` of the `event` object](https://github.com/facebook/draft-js/blob/master/src/component/handlers/drag/DraftEditorDragHandler.js#L39) that gets passed to `getSelectionForEvent` from the [`onDrop` handler](https://github.com/facebook/draft-js/blob/master/src/component/handlers/drag/DraftEditorDragHandler.js#L86) in DraftEditorDragHandler.js is itself a `Document` node. as a result, it fails the [`node.ownerDocument` check](https://github.com/facebook/draft-js/blob/master/src/component/utils/getCorrectDocumentFromNode.js#L13) and returns the global `document` object instead, which results in `getSelectionForEvent` throwing when it tries to [find the ancestor offsetKey](https://github.com/facebook/draft-js/blob/master/src/component/handlers/drag/DraftEditorDragHandler.js#L59). here’s what that stacktrace looks like in my codebase:

image

#### Test Plan**

`yarn test` is now all passing with this change. in addition, i verified that the error described above no longer occurs when i use the build that includes this change.

### DIFF
--- a/src/component/utils/getCorrectDocumentFromNode.js
+++ b/src/component/utils/getCorrectDocumentFromNode.js
@@ -10,6 +10,10 @@
  */
 
 function getCorrectDocumentFromNode(node: ?Node): Document {
+  if (node && node.nodeType === 9) {
+    // $FlowIgnore if nodeType is 9, node is a Document node
+    return (node: any);
+  }
   if (!node || !node.ownerDocument) {
     return document;
   }


ccpu:
i am having the same problem when dragging a text in the editor, this fix will solve the problem:

Animation
Capture

PR #3073, opened 2021-10-08
Fix composition mode

offcall:
#### Summary
After this fix https://github.com/facebook/draft-js/pull/2035
The composition mode began to break entities, this is because composedChars contains the text of the entire block (it causes unexpected behavior, since the entity text differs from its code), and not just the entered characters

My PR fixes this: gets the typed characters and inserts them in the right place
image
image
image

#### Tests
!need to replace old tests because they now depend on getSelection


ProvidenceXz:
Will this get merged?

PR #3133, opened 2022-04-12
Update mappings to handle lowercase react properly

zpao:
This should fix the build.


zpao:
Ok proved it works, will reset back to be standalone so it can import cleanly.

PR #3135, opened 2022-04-13
Upgrade fbjs

zpao:
This gets past a couple potential security issues with transitive
dependencies. This is effectively a no-op at this point.

The node-fetch change is not important, Draft doesn't make use of it.

The ua-parser-js change is relevant, however the vulnerable version has
been yanked from npm and fbjs already had an open enough range in its
dependency that people would have gotten upgraded without us needing to
explicitly upgrade fbjs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "test-ci": "cross-env NODE_ENV=test npm run lint && npm run flow && npm run test"
   },
   "dependencies": {
-    "fbjs": "^2.0.0",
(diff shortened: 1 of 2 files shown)


robrichard:
@zpao can you publish a new version to npm with this change?

stianjensen:
> can you publish a new version to npm with this change?

It doesn't seem like a new version was ever released. Will there still be a final security release of this lib before it goes into maintenance mode EOY?

